PR #4, opened 2026-08-01
Add native macOS text editing to the composer input

neo773:
## Problem

The composer input bound only bare arrows plus `home`/`end`. A laptop keyboard has no Home or End keys, so there was no way to reach the start or end of a line — and there was no undo at all.

## What this adds

| Keys | Behavior |
|---|---|
| `Cmd+Left` / `Cmd+Right` | line start / end |
| `Cmd+Up` / `Cmd+Down` | document start / end |
| `Shift` + any motion | extend the selection instead of moving |
| `Shift+Up` / `Shift+Down` | select by line |
| `Opt+Delete` / `Cmd+Delete` | delete by word / to the line start |
| `Opt+Fn-Delete` / `Cmd+Fn-Delete` | forward equivalents |
| `Cmd+Z` / `Shift+Cmd+Z` | undo / redo |

`Opt+Left/Right` for word motion was already bound and is unchanged.

## Notes on undo

- A run of single-character edits coalesces into one step, breaking on a pause (700ms), a word boundary, a paste, or a caret jump — so undo rewinds in the bursts the user actually typed rather than one character at a time.
- An IME composition snapshots before its first keystroke, so one undo drops the whole composition instead of landing on half-composed text.
- `set_text` (draft load, clear-on-submit) resets the history, so undo cannot resurrect a message that was already sent.
- History is capped at 200 steps.

## Implementation

Motion and deletion reuse the existing `previous_word_boundary` / `next_word_boundary` / `line_range_at` helpers, so there is no new text-segmentation logic. `move_vertical` was split into a pure `vertical_target` so the move and select variants share it.

Modifier-qualified keys are bound in the `PaletteSearch` context as well. That context deliberately leaves navigation keys unbound so they bubble to the palette frame — the palette navigates with **bare** arrows and enter, so the modifier-qualified bindings added here cannot collide with it.

Undo/Redo also appear in the Edit menu. They have no `OsAction` counterpart, so they dispatch as plain actions to the focused input.

## Testing

- `cargo test -p comet-ui` — 289 passed, 0 failed.
- Verified by hand in the running app on macOS: line/word motion, selection, and `Cmd+Z`.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788134676&installation_model_id=425430&pr_number=4&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F4&signature=8d0d67027cbe0548fc9479b7d3f1b141c9f261832934f3eda56cf2628647224d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/app_menus.rs
+++ b/crates/ui/src/app_menus.rs
@@ -118,6 +118,11 @@ pub fn app_menus() -> Vec<Menu> {
         // input — zed wires its editor actions identically
         // (crates/zed/src/zed/app_menus.rs, Edit/Selection menus).
         Menu::new("Edit").items([
+            // Undo/Redo have no `OsAction` counterpart — they dispatch as plain
+            // actions to the focused input, same as the composer keymap.
+            MenuItem::action("Undo", composer::Undo),
+            MenuItem::action("Redo", composer::Redo),
+            MenuItem::separator(),
             MenuItem::os_action("Cut", composer::Cut, OsAction::Cut),
             MenuItem::os_action("Copy", composer::Copy, OsAction::Copy),
             MenuItem::os_action("Paste", composer::Paste, OsAction::Paste),
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -530,18 +530,58 @@ actions!(
         Down,
         SelectLeft,
         SelectRight,
+        SelectUp,
+        SelectDown,
         SelectAll,
         Home,
         End,
+        SelectHome,
+        SelectEnd,
+        DocStart,
+        DocEnd,
+        SelectDocStart,
+        SelectDocEnd,
         WordLeft,
         WordRight,
+        SelectWordLeft,
+        SelectWordRight,
+        DeleteWordLeft,
+        DeleteWordRight,
+        DeleteToLineStart,
+        DeleteToLineEnd,
         Copy,
         Cut,
         Paste,
         Newline,
         Submit,
+        Undo,
+        Redo,
     ]
 );
+
+/// How long a run of single-character edits keeps merging into one undo step.
+/// A pause longer than this starts a fresh step, so undo rewinds in the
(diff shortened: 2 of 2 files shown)


sabari245:
Hey, thanks for working on this. Undo/redo and the extra selection controls would be really useful.

I was experimenting with similar shortcuts locally and noticed that word deletion in this PR seems to use `alt-backspace` and `alt-delete` on every platform. Option makes sense on macOS, but on Windows and Linux I would normally expect Ctrl+Backspace/Delete. Would you consider choosing that modifier based on `target_os`? The same convention could be used for selecting whole words with Shift+Left/Right.

From what I can tell, `PaletteSearch` also supports deleting the previous word but not the next one. Would it be possible to support both directions consistently in the composer and palette search fields?

Thanks again for taking this on.

wingleeio (approved):
Reviewed and verified locally: test-merged against current main (310 tests pass), no new clippy warnings, fmt-clean after 784d3cd. Undo snapshot placement, IME composition handling, and palette keymap non-collision all check out. Minor follow-ups (forward-delete coalescing, undo tests) noted in review discussion — not blockers.

PR #6, opened 2026-08-01
Add native scrolling to the composer input

sabari245:
## Problem

The composer grows with its content until it reaches the existing 260px textarea limit. After that it clips the overflow and follows the caret internally, but there is no way to scroll the prompt by hand. Reviewing an earlier paragraph means moving the caret to it, and dragging a selection stops as soon as the pointer leaves the input.

## What this adds

| Interaction | Behavior |
|---|---|
| Mouse wheel / trackpad | Scroll the overflowing prompt without moving the caret |
| Type or move the caret | Bring the caret back into view and resume following it |
| Drag above / below the input | Scroll in that direction while extending the selection |
| Move the drag farther outside | Increase autoscroll speed, capped at one line per frame |
| Scroll at the first / last line | Let the event propagate instead of trapping it |

Short prompts still use the same compact-to-expanded auto-grow behavior. Internal scrolling only becomes active once the input has more content than its existing maximum height.

## Notes on scrolling

- A manual scroll temporarily hands control of the viewport to the user, so the next paint does not snap straight back to the caret.
- Typing, clicking, or moving the caret restores the original caret-following behavior.
- Rewrapping during a resize continues to keep the caret visible unless the user deliberately scrolled away.
- Drag autoscroll runs only while an active selection is outside the top or bottom edge. It stops when the pointer returns, the mouse is released, or the content boundary is reached, so an ordinary click or in-bounds drag starts no idle frame loop.

## Implementation

The composer is a custom GPUI element that paints shaped `WrappedLine` values, selections, and the caret itself. The new behavior therefore builds on its existing `scroll_top` offset rather than introducing a second scroll container.

Wheel events accept both GPUI line deltas and precise pixel deltas. Offset calculation and caret reveal are kept as pure helpers so their direction and boundary behavior can be tested independently.

GPUI's normal element-level mouse move listener only fires while the pointer is inside the hitbox. Drag selection uses a window-level listener during paint, following the pinned Zed editor's approach. Once the pointer crosses an edge, a short-lived 60fps loop advances the viewport and remaps the selection endpoint against the newly visible text.

## Testing

- `cargo test -p comet-ui` - 313 passed, 0 failed
- Added focused tests for wheel direction, offset clamping, caret reveal, and distance-scaled drag autoscroll
- `rustfmt --check crates/ui/src/composer.rs`
- `git diff --check`
- Verified manually with a long prompt: wheel scrolling, caret-follow resumption, and drag selection in both directions

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788168373&installation_model_id=425430&pr_number=6&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F6&signature=5fec0b83130a44ef4a2162903538c27d153f8a8fc30e064c4ec13d44f5702d01"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -13,12 +13,13 @@ use std::path::PathBuf;
 use std::time::{Duration, Instant};
 
 use gpui::{
-    App, Bounds, ClipboardEntry, ClipboardItem, Context, CursorStyle, ElementInputHandler, Entity,
-    EntityInputHandler, EventEmitter, FocusHandle, Focusable, GlobalElementId, KeyBinding,
-    KeyDownEvent, LayoutId, MouseButton, MouseDownEvent, MouseMoveEvent, MouseUpEvent, ObjectFit,
-    PaintQuad, PathPromptOptions, Pixels, Point, SharedString, Style, StyledImage as _,
-    Subscription, Task, TextRun, TextStyle, UTF16Selection, UnderlineStyle, Window, WrappedLine,
-    actions, div, fill, img, point, prelude::*, px, relative, size,
+    App, Bounds, ClipboardEntry, ClipboardItem, Context, CursorStyle, DispatchPhase,
+    ElementInputHandler, Entity, EntityInputHandler, EventEmitter, FocusHandle, Focusable,
+    GlobalElementId, KeyBinding, KeyDownEvent, LayoutId, MouseButton, MouseDownEvent,
+    MouseMoveEvent, MouseUpEvent, ObjectFit, PaintQuad, PathPromptOptions, Pixels, Point,
+    ScrollWheelEvent, SharedString, Style, StyledImage as _, Subscription, Task, TextRun,
+    TextStyle, UTF16Selection, UnderlineStyle, Window, WrappedLine, actions, div, fill, img, point,
+    prelude::*, px, relative, size,
 };
 use unicode_segmentation::UnicodeSegmentation;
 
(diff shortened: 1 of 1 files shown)


wingleeio:
Maintainer push: merged `main` into this branch (resolving the `set_text` conflict with #4 by keeping both sides — the `follow_cursor` reset and the undo-history clear), plus a one-line follow-up so `restore()` re-enables caret following after undo/redo, matching the other selection paths. `cargo test -p comet-ui` (313 passed) and `cargo fmt --check` are green on the merged branch; behavior also verified in the running app (wheel scroll, caret-follow resumption, drag autoscroll).

wingleeio (approved):
Reviewed the full diff (composer.rs only), verified 313/313 comet-ui tests plus the workspace suite on the merged branch (the one shell_env flake reproduces on main and is unrelated), no new clippy lints, and exercised wheel scroll, caret-follow resumption, and drag autoscroll in the running app. Safe to merge.

PR #20, opened 2026-08-05
Terminal: resolve the ANSI palette per appearance

neo773:
The dynamic light/dark switch themed the terminal's chrome but not its grid. The panel read `Theme`, while the cells painted from two constants: a hardcoded `#090909` background and a single ANSI table tuned by eye against it.

In light mode that renders a near-black rectangle inside a white app, and the "bright" half of the palette — the half most CLI output reaches for — comes out pale-on-pale.

### What changed

`crates/ui/src/terminal/view.rs`. The background and the 16 named ANSI slots now resolve through `Appearance`:

- **Background** — `#090909` dark, `#fafafa` light. The light step is larger than dark's 3/255 for the same reason `Theme::light`'s surfaces are: a delta that reads as separation on near-black disappears on white.
- **ANSI 0–15** — a second table, same hue families walked down to the 600/700 steps, the way `Theme::light` already does for its accents. "Bright" keeps meaning *more prominent*, which on a light field is **darker**. A literal copy of the dark table would have made the bright half the invisible half.
- **The 6×6×6 cube (16–231)** — deliberately left appearance-independent. A program asking for index 196 is asking for `#ff0000` by arithmetic; remapping it would invent colors nobody picked. iTerm and Apple Terminal light themes do the same.
- **The grayscale ramp (232–255)** — mirrored. Tools use it for de-emphasis rather than for a specific grey, and its bright end (where dim hint text lands) is the end that vanishes on white.

One real finding while writing the tests: light yellow at yellow-600 measures 2.81:1 on the light background and misses AA. It moved to amber-700 — the step `Theme::light` already picked for its own `warning` token, for the same reason.

### Tests

Five new, in the same module:

- a per-slot contrast floor against each palette's own background — the check that would have caught the original omission
- bright-is-more-prominent, asserted in both directions
- the grayscale ramp mirrors; the cube does not
- the palette tables actually differ

Both contrast tests carry the black / bright-black pair as a documented exception rather than skipping it silently: those are structural greys, and the shipped dark tuning puts bright-black at 2.58:1.

`comet-ui` is 362/362 green, clippy and fmt clean. Verified by eye in the running app in both appearances and across a live switch.

SCR-20260806-bztm
SCR-20260806-bzrz

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788548347&installation_model_id=425430&pr_number=20&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F20&signature=674d91aa67d9852d8f395aa3b67f7c029c055197b39018a6c62545e364542ac6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/terminal/view.rs
+++ b/crates/ui/src/terminal/view.rs
@@ -1,7 +1,8 @@
 //! Terminal paint + input encoding.
 //!
-//! - the ANSI palette on the `#090909` terminal background (feature-inventory
-//!   §1.10) and the 256-color cube/grayscale resolution;
+//! - the ANSI palette on the terminal background (feature-inventory §1.10) —
+//!   `#090909` dark, `#fafafa` light — and the 256-color cube/grayscale
+//!   resolution, both resolved per [`Appearance`];
 //! - keystroke → PTY byte encoding (printables, control keys, arrows/nav
 //!   escape sequences, Ctrl- combos, Alt prefixing);
 //! - the 12 ms input coalescer and the 80 ms resize debounce constants (the
@@ -17,7 +18,7 @@ use gpui::{
     SharedString, Style, TextRun, Window, fill, font, outline, point, px, relative, size,
 };
 
-use crate::theme::{Theme, rgb_to_hsl};
+use crate::theme::{Appearance, Theme, current_appearance, rgb_to_hsl};
 
 use super::emulator::{CellColor, CellSnapshot};
 use super::panel::TerminalPanel;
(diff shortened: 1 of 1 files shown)


wingleeio (approved):
Reviewed the diff and verified locally: single-file UI change (color tables + tests), no I/O/unsafe/dependency changes. Merged onto current main: 362/362 tests pass, fmt clean, zero clippy warnings in the changed file. Cube indices stay appearance-independent as described; grayscale mirror is exact. Safe to merge.

PR #21, opened 2026-08-05
Terminal: mouse text selection and copy

neo773:
> **Stacked on #20.** The first of the three commits (`15e424d`, the light/dark palette) belongs to that PR and will drop out of this diff once it merges. The two commits to review here are `59f7c54` and `d645210`.

The terminal grid painted output but never let you take it back out — there was no way to highlight anything, so nothing could be copied.

### What works now

- **Drag** to select
- **Double-click** for a word, **triple-click** for a line
- **Shift+click** extends an existing selection
- **Cmd+C** (or Ctrl+Shift+C) copies

### How

`Term` already owns a selection and already knows how to keep its anchors on their text as output scrolls the grid underneath. So the emulator layer is a thin translation: viewport rows in, grid points out, `selection_to_string` for the copy. Cells carry a `selected` flag; the view merges selected runs into wash quads, painted over the cell background but under the glyphs so highlighted text stays readable.

### Two details from reading zed's terminal

Zed embeds the same `alacritty_terminal`, and their mouse layer handles two things that are not obvious until they bite:

**Overshoot has to force the selection *side*, not just clamp the cell.** Drag past the bottom edge and drift left of where you started, and a side derived from x stops the selection mid-row instead of taking the last line whole. Clamping alone leaves that bug. Alacritty and zed both force `Right` past the right/bottom edge and `Left` above the top; so does this.

**A press must not select on its own.** Without a travel threshold, the click that focuses the panel leaves a one-cell selection behind whenever the hand moves a pixel. Harmless today, but it is exactly what clobbers the clipboard the moment anything copies on selection change. 2px, matching zed's `SELECTION_DRAG_THRESHOLD` and gpui's own `div` threshold.

### Selection color

Deliberately achromatic — white at 22% on dark, black at 16% on light — rather than reusing `Theme::selection`, which is the composer's blue. A saturated veil sits on top of sixteen ANSI hues and drags every one toward itself: red output reads purple under selection, green reads teal. Neutral moves lightness only, so selected text keeps the colors the program asked for.

### Not implemented: mouse reporting

Programs that ask for the mouse (vim, htop) still do not receive events. That was already true before this change — but it means selection now owns the mouse unconditionally instead of going dead inside those programs, which is the more useful behaviour while reporting is missing. Zed gates all of this behind `mouse_mode(shift)`; when reporting lands here, that is the gate it has to go behind, and the check belongs in `TerminalPanel::on_mouse_down`.

### Tests

12 new. Pointer hit-testing including overshoot sides and degenerate metrics; word and line granularity; a selection surviving the output that scrolls it; a click without a drag reporting no selection; the wash staying achromatic in both appearances.

`comet-ui` is 375/375 green, clippy and fmt clean. Verified by hand in the running app.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788552823&installation_model_id=425430&pr_number=21&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F21&signature=4d7f69d4d540c62078874ffb2851eb3a4888279f735a145d9c89aca2da740d64"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/terminal/emulator.rs
+++ b/crates/ui/src/terminal/emulator.rs
@@ -7,6 +7,10 @@
 //! That split makes the whole escape-sequence surface unit-testable with
 //! scripted byte strings.
 //!
+//! Selection lives here too ([`Emulator::start_selection`] and friends) rather
+//! than in the panel, because `Term` is what knows how to keep anchors on their
+//! text as output scrolls the grid underneath them.
+//!
 //! API notes for the pinned `alacritty_terminal 0.26` / `vte 0.15`:
 //! - `Processor::advance` consumes a byte slice; `Term` implements the
 //!   `vte::ansi::Handler` trait directly, so no event-loop machinery is needed.
@@ -21,11 +25,19 @@ use std::rc::Rc;
 use alacritty_terminal::event::{Event, EventListener};
 use alacritty_terminal::grid::{Dimensions, Scroll};
 use alacritty_terminal::index::{Column, Line, Point};
+use alacritty_terminal::selection::{Selection, SelectionRange};
 use alacritty_terminal::term::cell::Flags;
 use alacritty_terminal::term::{Config, Term, TermMode};
 use alacritty_terminal::vte::ansi::{
     Color as AnsiColor, CursorShape, NamedColor, Processor, Rgb as AnsiRgb,
 };
+
+/// Grid coordinates and selection granularity, re-exported so the panel and
+/// view speak the emulator's vocabulary without depending on
+/// `alacritty_terminal` directly — the same seam [`CellColor`] draws for
+/// colors.
+pub use alacritty_terminal::index::{Point as GridPoint, Side};
+pub use alacritty_terminal::selection::SelectionType;
 
 /// Scrollback history kept client-side (lines). The engine's replay window is
 /// bounded separately (1 MiB); this only caps what stays scrollable in the UI.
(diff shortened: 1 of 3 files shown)


wingleeio (approved):
Reviewed both commits line-by-line and verified locally after rebasing onto main: no network/filesystem/process/unsafe code, no dependency changes; the only clipboard interaction is a write on an explicit copy keystroke (no copy-on-select). Scanned the full diff for hidden/invisible Unicode (bidi overrides, zero-width) — none; only em-dashes/arrows in doc comments. 376/376 tests pass, zero clippy warnings in the changed files, fmt clean. Rebased the branch onto main (dropping the stacked #20 commit) via maintainer edit — content is byte-identical to the reviewed diff. Safe to merge.

PR #22, opened 2026-08-05
Terminal: new-tab, select-all and close-tab shortcuts

neo773:
> **Stacked on #21** (which is itself stacked on #20). The first three commits belong to those PRs and drop out of this diff as they merge. The commit to review here is `8e63b6f`.

| Chord | Action | Scope |
|---|---|---|
| <kbd>Ctrl+Shift+`</kbd> | New terminal tab | Global |
| <kbd>Cmd/Ctrl+A</kbd> | Select all, scrollback included | Terminal |
| <kbd>Cmd/Ctrl+W</kbd> | Close active tab | Terminal |

New-tab opens the panel when it is closed and adds a tab when it is already open — `set_open` creates the first tab itself, so doing both would open two. Close-tab goes through the same path as middle-click and the tab's ×, so closing the last one still collapses the drawer and still tells the engine to reap the PTY.

### Three things that fail silently if you get them wrong

**Bindings must be registered from `apply_keymap`, not `terminal::panel::init`.** That function clears the keymap, and `Shell::new` runs it during startup — so anything bound at app init is gone before the first frame. No error; the shortcut simply never fires. There is a test keeping these out of `init`.

**Shifted punctuation is written as the character it produces.** `ctrl-shift-\`` parses, builds, and then never matches: the platform reports `~` with no shift modifier, so a binding demanding both matches nothing. Letters are the opposite — `ctrl-shift-a` is right, because Shift stays a modifier. zed's keymaps show the same split (`ctrl-~`, `ctrl-?` for punctuation; `ctrl-shift-c`, `ctrl-shift-n` for letters). A test pins the chord to key `~`, ctrl on, shift off.

**macOS "Close Window" moves off ⌘W to ⌘⇧W.** AppKit fires menu key equivalents in `performKeyEquivalent:`, before the key reaches any view, so a menu item holding ⌘W makes every view-level ⌘W binding unreachable. zed and Safari move it for the same reason.

### Consequences worth flagging

- **⌘W no longer closes the window.** ⌘⇧W does, and that is now the action's only binding — gpui derives the menu equivalent from the keymap, and a second ⌘W binding risks handing the menu back the chord it just gave up. Outside the terminal, ⌘W does nothing.
- **On Windows and Linux, select-all and close-tab resolve to Ctrl+A and Ctrl+W**, shadowing readline's beginning-of-line and delete-word inside the terminal. Same trade VS Code makes; shifted forms are the alternative if it grates.
- New-tab keeps literal Ctrl on macOS rather than Cmd, matching VS Code and zed.

Scoping select-all and close-tab to the `Terminal` context is also what gives them precedence: gpui ranks matches by context depth before registration order.

### Testing

`comet-ui` is 380/380 green, clippy and fmt clean. New coverage: select-all reaching into scrollback and not panicking on an empty grid; the binding table parsing, scoping, and resolving `mod-` per platform.

Verified in the running app that ⌘⇧W closes the window. The remaining chords are covered by unit tests but have not been through a manual pass end to end.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/22"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788555112&installation_model_id=425430&pr_number=22&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F22&signature=8fcb2d69db867d0c95ff4671ea9084672a37219b08a4fe2a68e549377f0decc2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/app_menus.rs
+++ b/crates/ui/src/app_menus.rs
@@ -91,7 +91,13 @@ fn macos_key_bindings() -> Vec<KeyBinding> {
         KeyBinding::new("cmd-h", Hide, None),
         KeyBinding::new("alt-cmd-h", HideOthers, None),
         KeyBinding::new("cmd-m", Minimize, None),
-        KeyBinding::new("cmd-w", CloseWindow, None),
+        // AppKit fires menu key equivalents in `performKeyEquivalent:`, before
+        // the key reaches any view, so a menu item holding ⌘W would make every
+        // view-level ⌘W binding unreachable. ⌘⇧W leaves the chord free for the
+        // terminal's close-tab, as zed and Safari do. Keep this the action's
+        // only binding: gpui derives the menu equivalent from the keymap, and a
+        // second ⌘W binding could hand the menu back the chord it just gave up.
+        KeyBinding::new("cmd-shift-w", CloseWindow, None),
     ]
 }
 
@@ -264,7 +270,9 @@ mod tests {
         };
         let combo = |source: &str| vec![Keystroke::parse(source).unwrap()];
         assert_eq!(find(Quit.name()), Some(combo("cmd-q")));
-        assert_eq!(find(CloseWindow.name()), Some(combo("cmd-w")));
+        // Not ⌘W: that chord belongs to the terminal's close-tab, and a menu
+        // equivalent would intercept it before any view saw it.
+        assert_eq!(find(CloseWindow.name()), Some(combo("cmd-shift-w")));
         assert_eq!(find(Minimize.name()), Some(combo("cmd-m")));
     }
 }
--- a/crates/ui/src/shell.rs
+++ b/crates/ui/src/shell.rs
@@ -44,7 +44,9 @@ use crate::state::{
     AppState, ConnectionStatus, EngineBootConfig, GatePhase, Indicator, OrgRow, format_time_ago,
     org_name_valid, parse_orgs, sort_memberships,
 };
-use crate::terminal::panel::{TerminalPanel, ToggleTerminal, clamp_terminal_height};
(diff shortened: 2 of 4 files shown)


neo773:
@wingleeio 
Could I get a review on this when you get some time?

wingleeio:
@neo773 I'll review this and try to get it in tomorrow

wingleeio (commented):
Shortcut feedback:

- Approve `Ctrl+Shift+\`` for opening a new terminal.
- Approve `Cmd+A` / `Cmd+W` in the terminal on macOS, and `Cmd+Shift+W` for closing the window.
- Do not approve `Ctrl+A` / `Ctrl+W` in the terminal on Linux and Windows; prefer shifted or configurable alternatives so those shell controls remain available.
- Do not approve `Cmd+W` doing nothing outside the terminal; prefer it to close the window there if routing permits.

wingleeio:
Code LGTM.

PR #30, opened 2026-08-08
Make Comet local-first with optional multi-device sync

### LINKED ISSUE #3: Allow users to run it locally with optional WorkOS integration
The creation of the account failed with an ''account not found''. There's no way to use the app without log in, Im stripping it out in a fork just to try it out.

jsgrrchg:
I tested this extensively on macOS, but I haven't been able to test it on Linux or Windows since I don't currently have environments set up for those platforms.

Please review this carefully, as these changes affect important parts of the app. I've also extended the E2E tests to improve macOS coverage.

Feel free to request any changes. I'll keep an eye out for feedback and address anything that comes up.

Closes #3 

-- The following is AI-- :

### Summary

- Make a clean Comet installation fully usable without an account or network connection.
- Turn multi-device sync into an explicit opt-in while keeping local and synced workspace data isolated.
- Harden authentication and runtime transitions so changing accounts never swaps storage or transports underneath a running engine.

#### Local-first engine profiles

- Introduce immutable `Local`, `Synced`, and `Development` workspace scopes, resolve the active profile once at startup, and expose it through the new `EngineInfo` RPC.
- Store local snapshots, run journals, and uploads under `profiles/local/` with a stable installation-local identity.
- Preserve the existing `orgs/{org_id}/{user_id}` layout for authenticated profiles, so current synced installations reopen their existing data in place.
- Keep Edge rooms, relays, and uploads disabled for local profiles. Development mode also stays offline unless an explicit bearer is configured.
- Keep device-scoped resources such as repositories, worktrees, agent credentials, settings, and device identity available across profiles without exposing transcripts or attachments from another profile.

#### Authentication and lifecycle

- Make `comet headless` and the Linux installer start immediately in local-only mode; signing in is no longer required to run the daemon.
- Make `comet login` and `comet logout` select the profile for the next engine start, while refusing to mutate credentials behind a running engine.
- Report the live runtime scope in `comet status`, with signed-out local mode treated as healthy.
- Add graceful engine shutdown over RPC and SIGTERM so desktop/daemon transitions drain state and release IPC before a replacement runtime starts.
- Fence canceled OAuth exchanges so an in-flight callback cannot restore credentials after sign-out, and back off transient offline refresh failures without treating them as revocation.
- Publish device and local-profile identities atomically and recover an empty legacy device identity safely.

#### Desktop experience

- Replace the signed-out account gate with a ready local workspace and an optional **Enable sync** action.
- Keep the current local workspace open throughout sign-in and require a restart before entering the synced profile; local sessions are never uploaded implicitly.
- Coordinate pending sync/restart state across viewports, block every synced viewport after sign-out, and wait for a remote daemon to stop before quitting for a profile change.
- Treat older daemons without `EngineInfo` as synced so they retain the previous authentication gate instead of being mistaken for local runtimes.

#### Tests and documentation

- Add integration coverage for signed-out startup, offline transport behavior, profile persistence and isolation, attachment jails, legacy synced-layout reuse, auth changes during a local runtime, and graceful daemon shutdown.
- Update the two-device smoke driver to materialize transcript reset/delta frames and make Worker process-group cleanup work on macOS and Linux.
- Document local-first installation, optional sync, profile storage boundaries, privacy guarantees, and the intentionally deferred cross-profile import/switching work.

### Upgrade behavior

- A clean installation starts in the local profile without contacting the Edge service.
- An installation with a valid saved WorkOS session continues to start in its existing synced profile and storage layout.
- Signing in does not upload, import, link, or delete local sessions or attachments.
- Signing out returns to the same local profile on the next start; switching profiles always requires an engine restart.

### Testing

- `cargo test --workspace --features comet-sync/mock-server -- --skip repos_round_trip_add_branches_worktrees --skip falls_back_when_interactive_attempt_hangs`
- `cargo test -p comet-rpc --example e2e_driver`
- `cargo clippy --workspace --all-targets --features comet-sync/mock-server`
- `scripts/e2e-smoke.sh`
- `git diff --check upstream/main...HEAD`

The changes introduced by this branch pass Clippy. The workspace Clippy command exits successfully with only pre-existing warnings outside the modified hunks: `collapsible_if` in `crates/update`; four `cloned_ref_to_slice_refs` warnings in `crates/doc`; `needless_return`, two `obfuscated_if_else` warnings, and one unused test variable in `crates/engine`; and, in `crates/ui`, seven `unexpected_cfgs` warnings from the macOS Objective-C macros plus `manual_is_multiple_of`, `collapsible_if`, `question_mark`, `unnecessary_sort_by`, four `doc_lazy_continuation` warnings, three `type_complexity` warnings, `too_many_arguments`, and `len_without_is_empty`. Running Clippy with `-D warnings` therefore stops at the first unchanged `collapsible_if` warning in `crates/update/src/lib.rs`. The workspace test command skips two unchanged baseline tests on macOS: one compares `/var/...` with its canonical `/private/var/...` path, and one is an existing interactive-shell timeout flake. `cargo fmt --all -- --check` currently reports only pre-existing formatting drift in unchanged `crates/sync/src/room.rs`, which is identical to `upstream/main`.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/30"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788805314&installation_model_id=425430&pr_number=30&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F30&signature=bcfd1bdaa628abea960847bdf8c766d652e82afd16e906a95353e921cfb758a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jsgrrchg:
I'll resolve conflicts now

wingleeio (changes requested):
## Review findings and Linux validation

Reviewed head `9a775fc890f4f592fbaca4c3355ade546074c94d`.

### Findings

1. **Local-only still contacts Edge.** [`assemble_runtime`](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/crates/engine/src/lib.rs#L546-L557) starts the updater regardless of `WorkspaceScope` or the computed `online` flag. After the initial delay, [the updater calls `fetch_latest`](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/crates/update/src/lib.rs#L539-L549), which requests the release manifest and fallback version file. This conflicts with the clean-install “without contacting Edge” behavior described by the PR. Please gate update checks on online scope or explicit update consent.

2. **Synced-profile attachment isolation is not account-scoped.** [All Synced and Development profiles still use `{data_dir}/uploads`](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/crates/engine/src/profile.rs#L55-L66), and [the attachment jail trusts the entire configured upload root](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/crates/engine/src/uploads.rs#L232-L240). Switching accounts therefore leaves known attachment paths, staging IDs, and commit filenames accessible or collidable across profiles. Please use account-scoped upload roots and define a migration/read-only fallback for the legacy cache.

3. **Engine assembly failures can leave the UI falsely Ready.** The embedded engine is assembled in a detached task, but [`attach_engine` marks the connection Ready immediately](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/crates/ui/src/state.rs#L826-L873). If store or journal assembly later enters `DeferredEngineState::Failed`, the standing watches log the subscription failure and return. A corrupt or inaccessible local store can therefore show an empty ready application with no error/retry path. Please propagate deferred failure into `ConnectionStatus::Failed`, or await local assembly before publishing Ready.

4. **The systemd crash-loop limit was removed while `Restart=on-failure` remains.** Both [the generated unit](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/apps/comet/src/daemon.rs#L227-L240) and [installer unit](https://github.com/zeronsh/comet/blob/9a775fc890f4f592fbaca4c3355ade546074c94d/edge/src/install.sh#L73-L84) now restart every five seconds without the previous explicit rate limit. Permanent errors such as a corrupt store, invalid environment, or bind failure will churn indefinitely. Please restore an explicit start limit.

### Device identity locking

The short device-identity lock is useful because identity resolution and runtime assembly can race before the lifetime engine lock. A kernel-owned file lock also recovers automatically after a crash.

The raw `unsafe { libc::flock(...) }` call is not needed on the workspace's rolling stable Rust toolchain. Safe cross-platform `std::fs::File::{lock, try_lock, unlock}` has been stable since Rust 1.89. The blocking identity lock can use `file.lock()?`; the lifetime instance lock can use `try_lock()` while preserving its retry and error handling.

### Validation

- Workspace test suite with `comet-sync/mock-server`: passed, with the two documented baseline tests skipped
- `cargo test -p comet-rpc --example e2e_driver`: passed
- Workspace Clippy across all targets: passed with existing warnings
- `git diff --check`: passed
- Headed Linux runtime: clean local boot, local workspace operation, restart persistence, optional-sync UI, login UI, and Devices settings all rendered successfully
- `cargo fmt --all -- --check`: the PR-touched failure is import ordering in `crates/ui/src/shell.rs`; the other reported files are outside this PR

### UI captures

Clean local start:

Clean local start

Local account menu with the optional sync action:

Optional sync menu

Enable Sync browser handoff:

Enable Sync dialog

Login gate:

Login view

Local Devices settings:

Local Devices settings

wingleeio:
thanks a lot for this PR, i did a cursory read and had codex go through this on a linux machine

nit: i think enable sync should be above settings

@SinaKhalili can you take a look as well please

jsgrrchg:
Thanks for the review, I’ve addressed all four findings and pushed the changes to this PR:

1. Local runtimes no longer start the Edge release updater. Update checks are now gated on the runtime being online.
2. Synced and development profiles now use account-scoped upload roots. The legacy device-wide cache is assigned to one account as a read-only compatibility fallback.
3. Deferred engine assembly failures now propagate to `ConnectionStatus::Failed`, shutting down the failed handle and exposing the existing retry path instead of leaving the UI falsely `Ready`.
4. The explicit systemd crash-loop protection has been restored with `StartLimitIntervalSec=60` and `StartLimitBurst=5` in both generated service definitions.

I also moved **Enable sync** above **Settings** in the account menu, as suggested.

The branch has been merged with the latest `main`, and the conflict resolutions preserve both the local-first behavior and the newer UI/runtime changes. 

Validation completed:

- `cargo check -p comet-ui -p comet-rpc --all-targets`
- All 396 `comet-ui` tests
- All local-first and profile-isolation tests
- `cargo test -p comet-rpc --example e2e_driver`
- Daemon/systemd unit tests
- `git diff --check`

Thanks again for the thorough review. Let me know what you think.

sahilsk11:
Just pulled down branch and pulled from source. Can confirm it seems to work fine - i'm able to send a message and it works.

Non-blocking nit is that "unknown-device" shows up in a few places on the UI:

Screenshot 2026-08-11 at 1 21 03 PM
Screenshot 2026-08-11 at 1 20 39 PM
Screenshot 2026-08-11 at 1 32 41 PM

jsgrrchg:
Captura de pantalla 2026-08-11 a las 14 25 02

image

I simplified the UI and changed the name by default for local only devices. User can still change the name.

wingleeio:
will re-review in a moment, the device default device name should probably generate smth instead of just defaulting to unknown, not related to this PR. i will push a separate update for this

wingleeio:
@jsgrrchg in the no project state screenshot, it says "Or start without a project"

let's remove that line for now, there's no composer in this view so the statement is not true, and i've disabled project-less chats for the moment.

jsgrrchg:
Fixed the device-name fallback across macOS, Windows, and Linux. New registrations now resolve the native machine name instead of showing `unknown-device`, and existing `unknown-device` entries are repaired for the current device without overwriting custom names.

I also added tests covering each platform fallback, blank values, unique fallback names, legacy repairs, and custom-name preservation.

This solves the ghost ''unknown-device'' from before.

jsgrrchg:
Haha, I think I may have moved a little too quickly here 😅

I added the device-name fallback fix in this PR just before seeing your comment that you were planning to handle it separately. Would you prefer that I remove that commit from this PR to avoid overlapping with your update?

In the latest commit, I also removed the “Or start without a project” action as requested.

jsgrrchg:
fix(engine): resolve native device names

I'm refering to this

jsgrrchg:
I found another issue while testing the full local-first flow. After starting fresh and then enabling online sync, previously registered devices reappear, including some legacy `unknown-device` entries.

The underlying issue is that the registry allows multiple device rows for the same Comet installation. Device identity is canonical only while the local `device-id` file survives. If that file is removed during a reset or reinstall, Comet generates a new UUID and registers the same machine as a new device. The previous remote rows remain indefinitely, and there is currently no automatic lifecycle mechanism to retire them.

This behavior is pre-existing in upstream. The local-first flow only makes it more visible: the isolated local profile initially shows a single current device, while signing in loads the complete remote device registry.

Multiple projects per device are expected, but multiple active device rows for the same installation are not. I think the intended invariant should be one active registry row per `(account, installation identity)`. The installation identity should survive ordinary profile or data resets, and stale offline rows without associated projects or sessions should eventually be retired automatically.

I think is best to solve this in another PR

wingleeio (changes requested):
## Re-review at `a895c487`

The four blockers from my previous review are addressed at this head: local update polling is gated, synced uploads are account-scoped with a one-owner legacy fallback, deferred assembly failures now surface in the UI, and the systemd start limits are restored.

I still found two blocking lifecycle issues:

1. **A synced runtime that starts during a transient Edge outage stays offline for the lifetime of the process.** [`assemble_runtime` snapshots `online` from one `access_token()` attempt](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/engine/src/lib.rs#L539-L554), then only constructs the updater, link cache, and host relay inside that one-time branch ([relay construction](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/engine/src/lib.rs#L563-L590)). The [background refresh loop](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/engine/src/auth.rs#L372-L419) can recover a bearer later, but there are no online transport supervisors left to consume it. A laptop with a saved synced profile that launches offline therefore will not rejoin registry/chat rooms, device relay, attachment mirroring, or update polling when the network returns; it requires an engine restart. Keep `Local` transport-free, but construct/supervise `Synced` transports around the dynamic token provider so the unavailable-to-available transition recovers in place.

2. **Signing out leaves already-authenticated relay sessions alive until a separate stop/quit action succeeds.** The [sign-out RPC result immediately moves the UI to `SignedOutRestartRequired`](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/ui/src/shell.rs#L1519-L1541), while [`Auth::sign_out` only clears token state](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/engine/src/auth.rs#L468-L475). `HostRelay` reads the bearer before opening a session, then [`host_session` runs without observing auth state](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/rpc/src/device_room.rs#L257-L292) and can continue dispatching remote RPC until the socket closes ([session loop](https://github.com/zeronsh/comet/blob/a895c487219282e5392163f2093443decc91f518/crates/rpc/src/device_room.rs#L360-L421)). The UI can therefore say **Signed out** while the daemon still serves an existing relay connection; shutdown depends on the user completing a second action. Make successful sign-out atomically chain into runtime shutdown and wait for IPC/data-lock release, or cancel every live online transport when auth becomes `SignedOut`.

### Validation

- 541 focused tests passed; 3 live/credentialed tests were ignored.
- `cargo build -p comet` and `git diff --check` passed.
- Headed Linux validation used a clean data directory and isolated IPC port.

### Linux captures from this head

Clean local-only startup:

Clean local-only startup

Optional sync entry point:

Optional sync menu

Local Devices view:

Local Devices view

jsgrrchg:
Addressed the sign-out lifecycle feedback.

Signing out now immediately closes the authenticated device relay and all active virtual RPC connections, while token refreshes continue without interrupting healthy peer links.

The app then shuts down the synced runtime, clears account-scoped UI state and subscriptions, and automatically boots the local runtime in the same session. This works for both embedded engines and attached background daemons.

Added coverage for immediate relay shutdown on sign-out, token-rotation continuity, and the full synced-to-local runtime transition.

wingleeio (changes requested):
## Re-review at `e91598d`

The offline-boot recovery fix is effective: synced profiles now retain their transport supervisors, token changes wake retries, and the attachment mirror has a persistent outbox. The headed sign-out path now chains into runtime shutdown, and the host relay closes its authenticated socket when credentials disappear.

One blocking lifecycle gap remains:

1. **A headless daemon does not shut down its already-connected sync rooms when authentication is revoked or `SignOut` is called over RPC.** A definitive refresh rejection [calls `Auth::sign_out`](https://github.com/zeronsh/comet/blob/e91598d691096c35324abe0b11a8e05f598a347e/crates/engine/src/auth.rs#L711-L722), but [`Engine::run` only waits for an OS shutdown signal or `StopEngine`](https://github.com/zeronsh/comet/blob/e91598d691096c35324abe0b11a8e05f598a347e/crates/engine/src/lib.rs#L631-L674); it never observes the auth transition. The new token signal closes `HostRelay`, but an already-joined registry client remains in its [event loop without an auth cancellation branch](https://github.com/zeronsh/comet/blob/e91598d691096c35324abe0b11a8e05f598a347e/crates/engine/src/workspace_host.rs#L335-L366), and an already-joined session room is [stored on the handle before its join supervisor exits](https://github.com/zeronsh/comet/blob/e91598d691096c35324abe0b11a8e05f598a347e/crates/engine/src/doc_host.rs#L891-L899). The `Shell` transition fixes headed use, but there is no `Shell` in `comet headless`, so existing registry/chat sockets can continue carrying account data until they independently disconnect or the daemon is stopped. Enforce this boundary in the engine: when a synced runtime reaches `SignedOut`, cancel every live Edge transport and transition or stop the headless runtime. Please add a daemon-level test that revocation or the `SignOut` RPC tears down already-connected room sockets.

### Validation

- 576 focused tests passed; 4 live/credentialed tests were ignored.
- Focused Clippy across the engine, RPC, UI, updater, and CLI completed successfully with warnings.
- `git diff --check` passed.
- Headed Linux validation used a clean data directory and isolated IPC port.

### Linux captures from this head

Local-only startup and optional sync entry point:

Local-only optional sync menu

Local Devices view:

Local Devices view

jsgrrchg:
Addressed the remaining headless daemon lifecycle issue.

A synced engine now observes authentication transitions directly. When authentication is revoked or `SignOut` is called over RPC, it disconnects all active Edge transports—including registry, chat/session rooms, device relay, and cached peer links—and then shuts down the headless runtime cleanly.

Token refreshes and rotations continue without interrupting healthy connections; teardown only occurs when the runtime reaches `SignedOut`.

I also added daemon-level coverage verifying that `SignOut` closes already-connected room sockets, returns the RPC response, releases IPC, and terminates the daemon.

wingleeio (changes requested):
## Re-review at `9d97dc3`

The previous headless-runtime blocker is fixed: the daemon now observes `SignedOut`, revokes relay/peer/registry room ownership, exits, and has integration coverage for reply delivery, socket teardown, and IPC release.

One blocking lifecycle issue remains:

1. **The synced updater survives the headed switch to Local and keeps polling Edge.** [`Updater::spawn`](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/update/src/lib.rs#L510-L531) moves a strong clone into a detached task, while [`check_loop`](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/update/src/lib.rs#L549-L568) runs forever with no cancellation path. The new same-process sign-out flow shuts down the synced `EngineRuntime` and bootstraps Local, but [`EngineRuntime::disconnect_edge`](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L439-L453) only tears down transports, and the updater created during [runtime assembly](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L603-L620) owns itself through that detached clone. Consequently, after [the UI replaces the synced runtime with Local](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/ui/src/shell.rs#L1640-L1683), the old updater still calls the release endpoint and can auto-apply/restart managed installs. This breaks the Local no-Edge boundary after sign-out. Please make the updater loop lifecycle-owned by `EngineRuntime` (for example with an owned cancellation token or join handle), cancel/await it during disconnect/shutdown, stop the token-change wake task as well, and add a synced-shutdown-to-Local regression test against a counting release endpoint.

### Validation

- 578 focused tests passed; 4 live/credentialed tests were ignored.
- `cargo clippy -p comet-engine -p comet-rpc -p comet-ui -p comet-update -p comet --all-targets` completed successfully with warnings.
- `git diff --check` passed for the latest commit and the complete PR diff.
- The only reported GitHub check is skipped.
- The PR is currently conflicting with `main` and needs to be rebased or resolved before merge.

### Current-head Linux captures

Fresh Local account menu:

Local-only account menu

Local Devices screen:

Local Devices screen

wingleeio (changes requested):
## Consolidated full-pass review at `9d97dc3`

I re-reviewed the complete 32-file PR (+5,278/-500), not only the latest fix. The daemon/room teardown change now looks effective. The earlier updater finding is valid, but it is one symptom of a larger lifecycle gap. Two root blockers remain:

### 1. Make shutdown a real ownership boundary for the old synced runtime

Assembly stores a strong `DocHost` in `SessionsEngine` and a strong `SessionsEngine` in `DocHost` ([assembly](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L200-L202), [sessions side](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/sessions.rs#L97-L103), [doc-host side](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/doc_host.rs#L165-L170)). Both are Arc-backed handles stored in non-clearable `OnceLock`s ([setters](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/doc_host.rs#L391-L408)), so the replaced synced graph is strongly cyclic and cannot be retired by dropping `EngineRuntime`.

At the same time, [`EngineRuntime::shutdown`](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L439-L453) only removes the host relay, disconnects selected transports, and calls the core shutdown methods. It does not own, cancel, or await all workers that can use Edge:

- The updater and token-change wake task are detached indefinitely ([spawn site](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L603-L620), [infinite loop](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/update/src/lib.rs#L510-L568)).
- The [attachment mirror worker](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/uploads.rs#L331-L364) is also detached. Its drain obtains one bearer before iterating the entire durable outbox, so sign-out after bearer acquisition can let it continue sending queued attachments after Local has booted ([drain](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/uploads.rs#L375-L452)).
- Diff sidecar and chat seed/checkpoint/tail work follow the same detached-task pattern. Even finite in-flight requests need to be inside the shutdown boundary.

Please break the strong cycle (for example, use a weak or explicitly clearable back-reference) and give `EngineRuntime` one owned cancellation/join boundary shared by every Edge-capable supervisor and worker. Revoke/cancel first, await all joins, then release the graph and bootstrap Local.

The regression should assemble a synced runtime against a counting/blockable fake Edge, start release and attachment work, sign out and shut down, then bootstrap Local on the same Tokio runtime. Assert that a `Weak` sentinel for the old graph no longer upgrades, every owned task has ended, and no endpoint receives a request after shutdown returns.

### 2. Serialize replacement-engine election across viewports

The shell intentionally has [every viewport independently reattach](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/ui/src/shell.rs#L913-L923). They can all observe the old port and instance lock as free in [`stop_synced_runtime`](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/ui/src/shell.rs#L445-L500), then enter `EngineHandle::bootstrap` together. Bootstrap probes TCP, decides to embed, and [binds IPC before runtime assembly](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/ui/src/state.rs#L221-L326); the actual [instance lock is acquired later](https://github.com/zeronsh/comet/blob/9d97dc381b9d781a505ac76c76104a712161c12f/crates/engine/src/lib.rs#L167-L173).

I reproduced this with two simultaneous `EngineHandle::bootstrap` calls: one viewport failed with "another comet engine is already running". The bind/lock ordering also permits the process that owns the IPC listener to lose the data-dir lock, leaving the winning engine without an attachable listener.

Please elect/serialize one replacement owner, or make bind/lock losers retry and connect to the winner. Add a two-viewport regression that starts both bootstraps simultaneously and proves both reach Ready with exactly one in-process owner and one remote client.

### Current-head UI captures

Local account menu with Enable sync

Local devices view

### Validation

- Changed-package suites: 578 passed, 4 ignored.
- Non-UI workspace suite with the sync mock-server feature: 361 passed, 11 ignored.
- Focused Clippy completed with warnings only; `git diff --check` passed.
- A temporary concurrent-bootstrap diagnostic reproduced blocker 2 and was removed; the review worktree is clean.

Separately, GitHub currently reports this branch conflicts with `main`.

wingleeio:
@jsgrrchg to close out the two remaining blockers from my consolidated review, I've pushed two commits to this branch (maintainer edits were enabled — hope that's alright; both needed fairly deep changes and I didn't want to keep you iterating on our internals).

**`ad1ea9e` — fix(engine): own every Edge worker across runtime shutdown**

- `EngineCore::shutdown` now cancels and awaits every Edge-capable worker: the release check loop (`Updater::shutdown`), its token-change wake task, the attachment-mirror drain (a cancel mid-drain drops the in-flight `PUT` rather than finishing it under a stale bearer), and the diff/spaces supervisors.
- Every doc-host task rides a `TaskTracker` raced against one `CancellationToken`; `DocHost::shutdown_workers` cancels, awaits them all, flushes open docs, then releases the chat handles.
- The sessions ⇄ doc-host strong Arc cycle is now severable (clearable back-edges, cut at shutdown), so a replaced synced runtime's graph actually drops instead of polling Edge forever behind the Local boot.
- Regression: `online_runtime_shutdown_stops_edge_workers_and_retires_the_graph` — an online runtime against a counting edge (with a seeded mirror intent retrying across the boundary) must produce zero requests after `shutdown()` returns, and a weak sentinel on the doc-host graph must fail to upgrade once the runtime drops. I verified the test fails against the previous shutdown behavior.
- One integration catch along the way: releasing doc handles before the flush lost transcripts across restart (caught by the existing `born_chat2_race` test) — the flush now precedes handle release.

**`a1a5810` — fix(ui): elect one replacement engine across viewports**

- `EngineHandle::bootstrap` serializes probe+embed behind a process-wide gate, and the embed path acquires the data-dir `InstanceLock` *before* binding IPC (via a new `Engine::assemble_runtime_with_lock`), so simultaneous viewport bootstraps after the sign-out → Local switch elect exactly one embedded owner while the rest attach over IPC — and the listener owner can no longer lose the data-dir lock. A lock held by a mid-start daemon is retried against its listener before surfacing the error.
- Regression: `concurrent_bootstraps_elect_one_embedded_engine` — two simultaneous bootstraps both reach Ready, one `InProcess` + one `Remote`.

**Linux validation at `a1a5810`** (since this branch had only been hand-tested on macOS):

- `cargo test --workspace --features comet-sync/mock-server`: **647 passed, 15 ignored, 0 failed** across all 34 test binaries — the two macOS-only skips from the PR description aren't needed on Linux.
- `cargo clippy -p comet-engine -p comet-ui -p comet-update --all-targets`: warnings only, all pre-existing (confirmed by re-running with the new commits stashed).
- `git diff --check`: clean.
- Local `cargo fmt` drift spans many files this PR never touched (toolchain-version noise), so I deliberately reformatted nothing.

I'll give the full branch one last pass before sign-off.

wingleeio (approved):
All blockers from the previous rounds are closed with regression coverage, and the branch is Linux-validated (647 tests green, clippy/diff-check clean). Thanks for the persistence on this one — the profile capture + shutdown lifecycle ended up in a really solid place.

PR #42, opened 2026-08-11
Fix initial message positioning in empty desktop chats

jsgrrchg:
## Summary

Improve the first-turn layout in an empty desktop chat without changing message persistence or the behavior of existing conversations.

When the first prompt is sent in an empty chat:

- position the prompt near the top of the visible transcript;
- let the assistant response grow below it while it fits;
- hand off to the existing follow-tail spring once the response overflows.

All subsequent sends retain the previous scroll-to-bottom behavior.

## Root cause

The transcript always re-engaged its bottom pin on send. That made the first prompt in an otherwise empty viewport rest at the bottom, leaving no space for the response to grow naturally beneath it.

## Implementation

- Pass the optimistic message ID and empty-chat state through the composer event.
- Add a temporary viewport-height trailing runway for the first turn.
- Remove the runway and re-engage the existing stick-to-bottom spring at real overflow.
- Cancel the temporary runway when the user navigates the transcript manually.
- Preserve the original follow-tail behavior for the second and later sends.

This is a presentation-only change. It does not modify the message schema, persisted ordering, document synchronization, or stored conversations.

## Validation

- `cargo test -p comet-ui` — 387 tests passed
- `git diff --check upstream/main...HEAD`

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/42"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789077683&installation_model_id=425430&pr_number=42&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F42&signature=01a3e3a1f48848aad93b3bfd294efeeb3486604df7487185854f37ebbfa73c93"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -378,9 +378,18 @@ pub fn send_button_mode(run_live: bool, has_text: bool) -> SendButtonMode {
     }
 }
 
-/// The prompt-at-top treatment belongs only to a chat's first send.
-fn send_starts_empty_chat(transcript_len: usize, pending_echoes_len: usize) -> bool {
-    transcript_len == 0 && pending_echoes_len == 0
+/// The prompt-at-top treatment belongs only to a chat's first send. The
+/// loaded transcript alone can't prove that: selecting a chat clears it
+/// synchronously and the watch task refills it later, so a fast send into a
+/// just-selected chat would read a history-laden transcript as empty. The
+/// chat row's own activity — bumped on every dispatch, `None` on a fresh row
+/// — closes that window.
+fn send_starts_empty_chat(
+    transcript_len: usize,
+    pending_echoes_len: usize,
+    chat_never_messaged: bool,
+) -> bool {
+    transcript_len == 0 && pending_echoes_len == 0 && chat_never_messaged
 }
 
 /// Find the unresolved input request the panel should serve, if any: an
@@ -4349,7 +4358,20 @@ impl Composer {
         // original follow-tail behavior.
         let started_empty = {
             let state = self.state.read(cx);
-            send_starts_empty_chat(state.transcript.len(), state.pending_echoes().len())
+            let chat_never_messaged = match state.selected_chat.as_deref() {
+                // New-chat canvas: the chat doesn't exist yet, so no history.
+                None => true,
+                Some(id) => state
+                    .chats
+                    .iter()
+                    .find(|c| c.id == id)
+                    .is_none_or(|c| c.last_message_at.is_none()),
+            };
+            send_starts_empty_chat(
+                state.transcript.len(),
+                state.pending_echoes().len(),
+                chat_never_messaged,
+            )
         };
         // Chat id: existing selection, or client-minted for the new-chat canvas
         // (the chat then appears from the doc host once the doc materializes).
@@ -6192,10 +6214,13 @@ mod tests {
 
     #[test]
     fn only_the_first_send_starts_an_empty_chat() {
-        assert!(send_starts_empty_chat(0, 0));
-        assert!(!send_starts_empty_chat(1, 0));
-        assert!(!send_starts_empty_chat(0, 1));
-        assert!(!send_starts_empty_chat(2, 1));
+        assert!(send_starts_empty_chat(0, 0, true));
+        assert!(!send_starts_empty_chat(1, 0, true));
+        assert!(!send_starts_empty_chat(0, 1, true));
+        assert!(!send_starts_empty_chat(2, 1, true));
+        // A transcript that merely hasn't LOADED yet reads len 0 — the chat
+        // row's activity is what says "this chat really has no history".
+        assert!(!send_starts_empty_chat(0, 0, false));
     }
 
     #[test]
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -378,6 +378,11 @@ pub fn send_button_mode(run_live: bool, has_text: bool) -> SendButtonMode {
     }
 }
 
+/// The prompt-at-top treatment belongs only to a chat's first send.
+fn send_starts_empty_chat(transcript_len: usize, pending_echoes_len: usize) -> bool {
+    transcript_len == 0 && pending_echoes_len == 0
+}
+
 /// Find the unresolved input request the panel should serve, if any: an
 /// unresolved input part on the LAST assistant entry — regardless of the
 /// entry's run status. The question stays answerable until the user actually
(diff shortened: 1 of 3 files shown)


wingleeio (commented):
Reviewed the diff and ran the branch through a headless rig with real composer sends against a paced mock harness — the feature works as described, including the paths that are easy to get wrong. Also merged current main locally: clean merge, 390 tests pass on the preview (387 on the branch as-is).

## Verified live

**First send in an empty chat** — the prompt anchors near the viewport top immediately (no bottom-flash before positioning: frame-stepping the video shows the echo appearing already anchored), and a short reply grows beneath it and *stays* anchored when the run ends without overflow:

https://github.com/user-attachments/assets/61412e51-e95b-44de-9608-d44c2082cd47

prompt top-anchored with the finished reply below

**Overflow handoff** — at a short viewport the same reply overflows mid-stream; the runway hands off to the follow-tail spring continuously (no one-viewport jump), and the transcript ends bottom-pinned with no phantom runway space:

https://github.com/user-attachments/assets/3f98e364-3916-42ec-a9ad-371fdaa64c74

**Second send while the runway is still active** — exercises the `own_turn.take()` + `release_pending` branch: the runway is removed, the new turn follows the tail classically, and the resting layout has no leftover synthetic height.

second send pinned at bottom

**Manual-scroll cancel** — one wheel tick during the anchored state removes the runway and the layout reverts to the ordinary bottom-hugging arrangement. Works, with one observation below.

| anchored | after one wheel tick |
|---|---|
| before | after |

## Finding: `started_empty` can misclassify a not-yet-loaded chat

`send_starts_empty_chat` reads `state.transcript.len()`, but `select_chat` clears `state.transcript` synchronously and the watch task repopulates it asynchronously (`apply_transcript`). A send that lands in that window — a pre-typed draft plus a quick Enter right after switching chats, or a remote-hosted chat on a slow link — sees `len() == 0` for a chat that has history, and the first-turn treatment (top anchor + viewport-high runway) kicks in mid-conversation once the history loads in above it.

It self-corrects on any manual scroll and does no data harm, so it's a polish-level issue, but the fix is cheap: gate on a signal that knows the transcript has actually synced — e.g. also require the selected chat row itself to look unused (no `last_message_at` / never-run status), or an explicit "transcript delivered" bit next to `transcript_task`. Happy to leave it as a follow-up.

## Observations (not blockers)

- The wheel-tick cancel is an instant one-frame reflow from top-anchored to bottom-hugging (up to ~a viewport of travel). It's the user asking for control, so a snap is defensible — but a short glide would read less like a glitch, and the "Scroll to bottom" pill briefly shows even though the content is already resting at the bottom (visible in the right screenshot above).
- `own_turn_anchor_ix` and the `bottom_pad` anchor-existence guard are both linear scans per step/layout — fine at transcript sizes, just noting they run per frame while the anchor is live.

## Code notes

The implementation is idiomatic for this file: the step machine follows the existing spring-driver pattern (single-flight `on_next_frame`, kick flags), the pure helpers (`send_starts_empty_chat`, `own_turn_has_overflow`) are unit-tested in the same style as `should_restick`, and the cancel surface is complete — wheel/touch, jump pill, rail clicks (`set_scroll_task`), and chat switches (with the nice touch of *preserving* the anchor when the switch lands on the anchored chat, which is exactly what the new-chat mint flow needs). The `remeasure_items` invalidation on runway migration between last rows shows real understanding of the list virtualizer.

Nice work — this is a subtle piece of scroll choreography and it holds up under prodding.

wingleeio (commented):
Pushed 47a0c29 addressing both findings from my first review, re-verified everything in the rig.

## Fix 1: `started_empty` trusts the chat row, not the unloaded transcript

`send_starts_empty_chat` now also requires the selected chat row to have no `lastMessageAt` — the row is bumped on every dispatch and `None` on a fresh chat (and the yet-unminted new-chat canvas), so it stays correct while the transcript vec is cleared-but-not-yet-refilled after a chat switch.

Reproduced the race directly: froze the engine daemon (`SIGSTOP`) so the transcript watch couldn't deliver, selected a chat with history (transcript pane empty — `len() == 0`), and sent. The echo lands in the **classic bottom layout** — no top anchor, no runway — and after thawing the daemon, history loads in above a normal conversation. Pre-fix, this exact sequence anchored the prompt at the top with a viewport of blank below it and then dumped history on top.

| sent while the transcript hadn't loaded (daemon frozen) | after the daemon thaws — history loads in above, layout normal |
|---|---|
| race frozen | race loaded |

## Fix 2: wheel/touch is clamped while the anchor holds, not obeyed

While the first-turn anchor is live, everything already fits on screen — the prompt at the top, the whole reply below, and under that only synthetic runway. A scroll gesture has nothing to reveal in either direction, so the scroll handler now clamps back to the anchor instead of cancelling the runway (which collapsed the synthetic height and teleported the layout by up to a viewport, flashing the jump pill on the way).

Verified with five wheel ticks (three down, two up) against the anchored state: the before/after screenshots are **pixel-identical** (`ImageChops.difference(...).getbbox() == None`), and no jump pill appears. The video also shows the second send afterwards still doing the classic follow-tail:

https://github.com/user-attachments/assets/28c95fd2-1e79-4b1a-b4a1-dfa23a6a1c83

anchored state the wheel can no longer disturb

The runway now retires through exactly one automatic path — the overflow handoff — plus the explicit navigations (jump pill, rail clicks, chat switches). Re-verified the handoff end-to-end with a long first turn: anchored during the stream, continuous handoff at overflow, bottom-pinned at rest:

https://github.com/user-attachments/assets/b32f8934-4bb4-4700-aab8-c17743be7eac

## Validation

- `cargo test -p comet-ui` — 387 passed on the branch (the `started_empty` unit test now pins the unloaded-transcript case)
- Local merge preview with current main: clean merge, 390 passed
- Rig: first-send anchor (no flicker), wheel clamp (pixel-identical), race guard (reproduced via frozen daemon), second send, overflow handoff — all on the merged build

wingleeio (approved):
First-turn anchor, overflow handoff, second-send, and chat-switch paths all rig-verified on a merged build (390 tests green); both review findings fixed in 47a0c29 with the race reproduced and the clamp pixel-verified. The scroll choreography holds up under real prodding — nice work.

PR #45, opened 2026-08-11
Changes header: truncate long refs, tighten collapsed-takeover inset

wingleeio:
Two takeover-header reports on v0.1.45:

- **Long ref overflow** — a long base-ref name plowed over the fold/expand/close buttons: the base trigger was `flex_none`, so it couldn't shrink and the row overflowed (paint order stacked the text under the buttons). Both sides of the ref selector now truncate (`min_w_0` + truncating labels on the trigger and the branch label), the controls wrapper clips as a backstop, and the stats-strip label truncates too.
- **Collapsed-sidebar takeover hole** — the strip's left inset double-counted `title_bar_content_start`'s 10px text margin on top of its own 8px pad, leaving a gap after the nav cluster's `+`. The text margin is dropped; the scope dropdown now sits 8px off the `+`.

| | |
|---|---|
| Long base ref — both sides ellipsize, buttons clear | Collapsed-sidebar takeover — tight after the `+` |
| truncated | collapsed |

Driven live in the rig (long branch picked through the search menu). Tests green.

### DIFF
--- a/crates/ui/src/changes.rs
+++ b/crates/ui/src/changes.rs
@@ -1505,19 +1505,32 @@ impl Changes {
         popover::popover_card(theme)
             .w(px(180.0))
             .on_mouse_down_out(cx.listener(|this, _, _, cx| this.close_scope_menu(cx)))
-            .children(DiffScope::ALL.into_iter().enumerate().map(|(ix, scope)| {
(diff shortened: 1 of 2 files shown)


wingleeio:
Also folded in: the scope menu's rows had no flex gap — adjacent washes abutted into one slab. Now the same 2px column as every other menu.

scope menu gap

PR #49, opened 2026-08-11
feat(ui): add configurable interface typography

jsgrrchg:
## Summary

Add device-local interface typography settings under Settings → Appearance.

Users can now:

- Choose Geist, Geist Mono, System UI, or compatible fonts installed on the current operating system.
- Select an interface font size from 12 px to 20 px.
- Apply typography changes immediately without restarting Zeron.
- Keep their selection across cold starts and hot restarts.

The default remains Geist at 16 px, preserving the existing appearance for both new and existing users.

## Font packaging

- Bundle only Geist and Geist Mono, including the static weights required by the UI and technical text roles.
- Inter and Atkinson Hyperlegible Next are not bundled or offered as bundled choices.
- Discover installed system fonts at runtime and expose compatible families in the selector, filtering out faces that cannot provide the glyph metrics used by GPUI.
- Persist installed-font selections safely and fall back to Geist if a previously selected family is unavailable.
- Package the bundled font licenses with macOS and Linux distributions.

## Implementation

- Add compact, keyboard-accessible dropdowns for font family and size.
- Apply the selected typography to interface text and conversational prose.
- Keep code blocks, inline code, diffs, tool surfaces, and terminal content on their existing fixed Geist/Geist Mono roles.
- Preserve the selected family when switching between system, light, and dark appearance modes.
- Centralize `ui-settings.json` mutations behind a single serialized writer, preventing immediate typography updates from racing with debounced shell persistence.
- Invalidate Markdown text caches after typography changes.
- Re-measure the virtualized transcript when the font family or size changes so off-screen row heights and scroll positioning remain accurate.

## Testing

- `cargo test -p zeron-ui` — 439 tests passed after merging the latest `upstream/main`.
- `git diff --check` passed.
- The branch is updated with the latest `upstream/main` and the worktree is clean.


SinaKhalili:
This looks good to merge @wingleeio

wingleeio (commented):
## Review: font binary audit + code audit

Because this PR replaces/adds 16 binary `.ttf` files that a diff can't show, I did a byte-level provenance analysis of every font against Vercel's official releases before looking at the code. Summary up front: **the code is clean and tests pass, but the bundled fonts are not pristine upstream files — they are official Geist v1.7.2 binaries that have been edited**, and one of the edits (a new `liga` GSUB feature) is functionally broken in monospace contexts. LGTM once the font provenance question below is resolved.

### 1. Font provenance (hash + table analysis)

None of the 16 `.ttf` files hash-match any asset in the official `vercel/geist-font` releases (checked v1.7.2, 1.8.0, 1.7.0, 1.6.0). Table-by-table comparison against v1.7.2 (the closest match — `name`, `post`, `cmap`, `GDEF`, `GPOS`, `hmtx`, `fpgm`/`prep`/`cvt` are all **byte-identical**) shows exactly three kinds of edits:

| Edit | Where | Assessment |
|---|---|---|
| `head` table: checksum + modified-timestamp only | all 16 faces | benign build re-stamp (main's current fonts have the same property) |
| `glyf`: ±1 font-unit nudges on ~1–8 letter glyphs per face (`@ 2 § Ю g ǥ o ѳ ө ∂ ↝ ф Ω`…), plus matching +1s in those glyphs' hint bytecode | 8 faces | benign re-fit noise, sub-pixel at UI sizes |
| `GSUB`: a **new `liga` feature** (+768 bytes, identical 69-rule set in all 8 mono faces) wiring up the coding-ligature glyphs that ship dormant in official Geist Mono; `glyf`: the `░▒▓` shade glyphs re-serialized (contour reordering — verified same contour *set*) | all 8 GeistMono faces | deliberate, undisclosed, and **defective** (below) |

No new tables, no `cmap` remapping, no name/post changes, no homoglyph-style substitutions (every liga rule maps punctuation sequences to the canonical ligature glyph of the same name, and those glyphs' outlines are identical to official). Hinting programs identical. Supply-chain side: `fontdb` and `ttf-parser` are reputable registry-pinned crates; the only unfamiliar new lock entry (`zcheapstr`) is published by the zbus maintainer and checksum-matches crates.io.

### 2. Render-equivalence proof

I rasterized **every glyph of all 16 PR faces against its official v1.7.2 counterpart** (standalone TrueType rasterizer, no app code) at 64 px and 128 px. Result: all ~17,000 glyph pairs render pixel-identical except the ±1-unit-nudged letters and their composites, which differ by 1–4 edge pixels at 128 px (invisible at UI sizes). The shade glyphs `░▒▓` render pixel-identical despite their ~450-byte serialization diff.

The one *behavioral* change is the new `liga` feature:

ligature comparison

### 3. The `liga` edit is broken in monospace contexts

Every one of the 69 ligature glyphs has a single 600-unit (1-cell) advance while replacing 2–4 characters, and draws its ink up to **2.9 cells backward** past its own cell (`--->` has ink extent x ∈ [−1738, 538]). These glyphs were designed for the FiraCode-style `calt`+spacer technique, where the component advances are preserved; wiring them through a plain type-4 `liga` substitution collapses columns *and* paints over preceding characters wherever a shaper honors `liga` (the default in CoreText and HarfBuzz-family shapers):

overlap defect

Top: official font. Bottom: PR font with `liga` active — the arrow strikes through the `x`, `!==` smears over the `a`. If Zeron's text stack never enables `liga`, the edit is dead weight; if it does (or ever starts to), code blocks, diffs, and the terminal will misrender. Either way this edit should not ship silently inside "Geist Mono". Recommendation: **rebundle byte-for-byte official release assets** and, if coding ligatures are wanted, pursue them upstream or as a properly-built `calt` follow-up.

### 4. Code audit

Clean. Full diff reviewed with a focus on anything a font/typography PR shouldn't do: no network access, no process spawning, no file access outside `ui-settings.json` (atomic tmp+rename writer), font family strings never used in path construction, `ttf-parser` parse results handled fallibly (no panics on malformed system fonts), and the mechanical `px(N)` → `typography::*` sweep in ~20 files verified hunk-by-hunk to be exactly that. Packaging scripts add only the two license copies; the OFL text is verbatim.

Minor, non-blocking:
- `register_fonts` runs `fontdb::Database::load_system_fonts()` + a per-face `m`-glyph probe synchronously on the main thread before first frame — on font-heavy systems this adds real startup latency; the enumeration is only needed by the settings picker and could move to the background executor.
- Settings writes now happen synchronously on the UI thread (tiny file, but a slow home dir will hiccup a frame).
- The markdown cache's LRU cap check runs before insert, so the cap can be exceeded by one entry per (family, size) switch — bounded, harmless.

### 5. Tests

`cargo test -p zeron-ui` on the PR branch merged with current main: **569 passed, 0 failed**.

### Verdict

LGTM pending the font provenance answer requested in the comment below — specifically byte-for-byte reproducibility of the bundled `.ttf`s (ideally: pristine official release assets, with the `liga` edit dropped or moved to a disclosed, upstream-blessed follow-up).

wingleeio:
@jsgrrchg thanks for the thorough PR — the code side reviewed clean and the feature works as described. One thing we need to close out before merging: **font provenance**.

The 16 bundled `.ttf`s don't byte-match any official `vercel/geist-font` release asset. They're clearly derived from v1.7.2 (most tables are byte-identical), but they've been modified: a new `liga` GSUB feature in all Geist Mono faces, re-serialized `░▒▓` shade glyphs, and ±1-unit outline nudges on a handful of letters (details and rendered comparisons in my review). Since fonts are opaque binaries that get parsed by every user's system, we hold them to the same bar as vendored code. Could you please:

1. Say where these exact files came from (which release/commit/build of `geist-font`, npm package, Google Fonts, or a local tool run)?
2. If you edited or rebuilt them: which tool + exact steps, so we can reproduce the bytes from an official upstream artifact (`sha256`-for-`sha256`)?
3. If the `liga` addition was intentional: our analysis shows those ligature glyphs are built for a `calt`+spacer approach and misrender via plain `liga` (single-cell advance, ink overlapping preceding cells — see the second image in the review). We'd prefer to bundle pristine official assets and treat ligatures as a separate, disclosed follow-up.

If pristine v1.7.2/1.8.0 release assets work for the feature (they should — the static weights you need are all in the official zips), swapping them in is probably the fastest path to merge. ¡Gracias!

jsgrrchg:
Im so sorry, i reviewed my checkpoints, I used the TTF files committed to the official Geist repository via `raw.githubusercontent.com` under the `v1.7.2` tag. They match the repository files byte for byte but differ from the rebuilt binaries published in the release ZIP and npm package.Im fixing this now.

jsgrrchg:
@wingleeio  Thanks for flagging this. The previous TTFs were downloaded from the files committed to the upstream repository under the v1.7.2 tag via raw.githubusercontent.com, which differ from the rebuilt release artifacts. I’ve now replaced all 16 Geist and Geist Mono faces with the binaries from the official v1.7.2 release ZIP and verified their checksums. The problematic liga feature is no longer present in any Geist Mono face, while Geist Sans retains the standard ligatures shipped in the official release. All typography tests pass.

wingleeio (approved):
## Re-review after c3d29cd — verified, approving

Re-ran the full audit against the new head (`c3d29cd`). Everything checks out:

**Fonts — pristine.** All 16 `.ttf`s are now **byte-identical (16/16) to the official `vercel/geist-font` v1.7.2 release ZIP assets**. The `liga` GSUB and every other delta from my previous review are gone (each mono face shrank by exactly the 768-byte liga table).

**Provenance story — independently confirmed.** I also verified @jsgrrchg's explanation of the original fonts: all 16 previous `.ttf`s byte-match the TTFs committed in the upstream repo *tree* at the `v1.7.2` tag (`raw.githubusercontent.com/vercel/geist-font/v1.7.2/fonts/...`), 16/16. So the earlier deltas — including the misbehaving `liga` wiring — were Vercel's repo-tree build state, not a local modification. Good faith fully established; the repo-tree/release-ZIP discrepancy (and the broken plain-`liga` wiring in the repo-tree mono builds) is an upstream `geist-font` issue.

**Code delta since the audited head — mechanical only.** Besides the font swap (which touches zero code), the only changes are the upstream `v0.2.28` merge; I diffed the PR-side hunks line-by-line against the previously audited diff and every difference is the same `px(N)` → `typography::ui_rems(N)` substitution re-applied to main's restructured picker code, plus rustfmt reflow. Cargo.lock changed only by the internal 0.2.27 → 0.2.28 workspace bumps — no new packages since the audited (clean) state.

**Tests:** `cargo test -p zeron-ui` on `c3d29cd`: **571 passed, 0 failed**.

LGTM — the provenance condition from my previous review is resolved. Thanks for the quick, verifiable fix, José.


---

### Screenshots (feature verification on `c3d29cd`, Linux)

New **Interface font** section under Settings → Appearance (defaults: Geist / 16 px):

typography section

Family picker: bundled Geist / Geist Mono, System UI, plus compatible installed system fonts discovered at runtime:

family picker

Applied live, no restart — conversation prose and UI chrome follow the chosen family (DejaVu Sans here) while inline code and code blocks stay on Geist Mono, as promised. The selection also survived a cold restart (`ui-settings.json`):

chat in DejaVu Sans

Size picker (12–20 px):

size picker

At 20 px — transcript re-measures cleanly, code block keeps its fixed mono size:

chat at 20px

PR #50, opened 2026-08-12
Folder picker: slash-to-descend, tab completion, ctrl-n/p

SinaKhalili:
The add-space palette (⌘K → New project…) browsed folders with arrows only. Three additions make it read like a shell prompt.

## Slash descends

Typing `/` after a query that names a folder in the current listing — exact name, or a unique prefix — enters that folder exactly as if you'd picked it, and clears the query. A slash that matches nothing just stays in the query, and the list shows "No folders match".

Scoped to single path segments, since the palette browses one level at a time. Chaining works naturally: `fun/` descends, then `comet/` descends again.

## Tab completion with a ghost preview

`ComposerInput` gained first-class ghost-text support: a `set_ghost` API plus prepaint/paint logic drawing the completion suffix in faint ink at the end-of-text caret. It only paints for a collapsed caret at the end of non-empty, non-IME text, and never joins the real text layout — hit-testing, selection, and the caret are unaffected.

The palette recomputes the completion each render: the highlighted row if the query prefixes its name, else the first prefix match. Matching is case-insensitive but the ghost keeps the folder's real casing, so typing `doc` previews `uments` for `Documents`. That's `completion_prefix_len` in `pickers.rs` — a pure helper returning a byte length into the *name*, unit-tested including multibyte input.

**Tab** accepts the ghost, filling in the full folder name; `/` or Enter then descends. New `tab · Complete` chip in the footer legend.

## ctrl-n / ctrl-p

`classify_key` now maps ctrl-n → Down and ctrl-p → Up. Every picker routes keys through it (folder browser, spaces filter, branch picker, harness/model, traits), so they all get readline motion, not just this palette. Neither chord is bound in the composer or palette keymaps, so both bubble to the frame unconsumed.

## Testing

`cargo test -p comet-ui` — 386 passed. New unit tests cover `completion_prefix_len` (casing, multibyte, non-prefix) and the ctrl-n/p classification including the negative cases (bare `n`, cmd-`p`).

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/50"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789101724&installation_model_id=425430&pr_number=50&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F50&signature=2062dc071f939ed31afa4c5a450ac32ebe883c0d4e8a2182f8dc284609316526"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -1267,6 +1267,11 @@ pub struct ComposerInput {
     last_width: f32,
     /// Raw Markdown → chip display projection from the last layout pass.
     projection: TextProjection,
+    /// Inline completion preview: painted in faint ink after the text while
+    /// the caret sits at the end (palette tab-completion). Owned by the
+    /// wrapper — it recomputes and re-sets this on every render pass, so the
+    /// input never has to know what the completion means.
+    ghost: Option<SharedString>,
     /// File mentions are a composer feature, not a behavior of generic inputs
     /// (picker searches and rename fields also use this type).
     mentions_enabled: bool,
@@ -1337,6 +1342,7 @@ impl ComposerInput {
             max_line_width: 0.0,
             last_width: 0.0,
             projection: TextProjection::default(),
+            ghost: None,
             mentions_enabled: false,
             layout_epoch: 0,
             display_is_placeholder: true,
@@ -1483,6 +1489,16 @@ impl ComposerInput {
 
     pub fn is_empty(&self) -> bool {
         self.content.is_empty()
+    }
+
+    /// Set (or clear) the inline completion preview. Only paints while the
+    /// caret sits at the end of a non-empty draft — see the prepaint gate.
+    pub fn set_ghost(&mut self, ghost: Option<SharedString>, cx: &mut Context<Self>) {
+        if self.ghost == ghost {
+            return;
+        }
+        self.ghost = ghost;
+        cx.notify();
     }
 
     pub fn has_newline(&self) -> bool {
@@ -2741,6 +2757,10 @@ struct ComposerTextPrepaint {
     mention_quads: Vec<PaintQuad>,
     mention_hits: Vec<MentionHit>,
     selection_quads: Vec<PaintQuad>,
+    /// Completion preview: window-space origin of the end-of-text caret plus
+    /// the suffix to paint there (shaped at paint time — it never joins the
+    /// content's own layout, so hit-testing and the caret ignore it).
+    ghost: Option<(Point<Pixels>, SharedString)>,
 }
 
 impl IntoElement for ComposerTextElement {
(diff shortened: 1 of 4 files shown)


wingleeio (commented):
Reviewed the diff and took the branch for a spin in a headless rig (sway + seeded demo engine) — all three features work as described. Full demo take:

https://github.com/user-attachments/assets/8f9ef83f-5b25-4d7b-91c7-d00b892c6fd1

## Verified live

- **Ghost preview** paints the completion suffix in faint ink at the caret, keeps the folder's real casing, and never disturbs selection/hit-testing. It correctly disappears once the query equals the match, and the `set_ghost` no-op guard keeps the render-pass recompute from looping.

  ghost preview

- **Tab** fills in the full folder name; `/` then descends and clears the query, and chaining (`gith` → ⇥ → `/` → `co` → ⇥ → `/`) walks two levels exactly like a shell prompt.

  tab accepted

- **A slash that matches nothing stays in the query** with "No folders match" — honest, as advertised.
- **ctrl-n / ctrl-p** move the highlight in the folder browser. I also checked the keymaps: neither chord is bound in `Composer` or `PaletteSearch`, so the frame-wide claim in the `classify_key` comment holds, and every `classify_key` caller (branch palette, project filter, generic pickers) picks the motion up for free.
- `cargo test -p comet-ui` — 386 passed, matching the PR.

## Finding: exact-match tiebreak ignores casing

`add_space_slash_descend` resolves an exact name case-insensitively (`completion_prefix_len(&e.name, query) == Some(e.name.len())`) and takes the **first** hit in listing order. With case-colliding siblings, exact casing can't disambiguate: my home has both `github` and `GitHub`, and typing `GitHub/` — the exact casing of the second folder — descends into lowercase `github`:

| typed `GitHub`, highlight already on `github` | after `/` — wrong folder |
|---|---|
| before | after |

Suggest trying a case-*sensitive* exact match first, then falling back to the current case-insensitive resolution:

```rust
dirs.iter()
    .find(|e| e.name == query)
    .or_else(|| dirs.iter().find(|e| completion_prefix_len(&e.name, query) == Some(e.name.len())))
    ...
```

Niche (needs case-colliding sibling folders), so fine as a follow-up too.

## Nit

- Slash-descend re-evaluates on **every** `Edited` event, not just a typed `/`. A query like `src/` that legitimately stuck around (typed while the listing was still loading, or a mid-query edit leaving a trailing slash) can descend on a later unrelated edit — e.g. `src/x` → backspace → `src/` fires the descend, so a *deletion* navigates. Scoping the check to edits that end with a newly-inserted `/` would rule that out. (The flip side is pleasant though: pasting `Documents/` descends immediately.)

Nice work — `completion_prefix_len` returning a byte length into the *name* is exactly right for the casing-preserving suffix, and the multibyte tests cover the boundary that would have bitten.

wingleeio:
Pushed 6904151 addressing the case-collision finding: the segment resolution now lives in `pickers::segment_target` — case-sensitive exact match first, then the existing case-insensitive exact, then unique prefix — with a unit test covering the `github`/`GitHub` collision.

Re-verified in the same rig: `GitHub/` now descends into `GitHub` (below), `github/` still reaches its lowercase sibling, and ghost + tab completion behave as before. `cargo test -p comet-ui`: 387 passed.

GitHub/ descends into GitHub after the fix

wingleeio (approved):
All three features verified live in the rig (demo + stills in my review), the case-collision fix is in with a unit test, and the suite is green at 387. Nice addition — the palette really does read like a shell prompt now.

PR #51, opened 2026-08-12
Desktop notifications: banners on the session ping edges

SinaKhalili:
## Summary

Sessions already chimed when a run finished or an agent asked a question, but those pings were inaudible with Comet in the background behind another app. This posts a **system banner** on the same transitions, with a new Settings page to control it.

**`crates/ui/src/notify.rs`** (new) follows `sound.rs`'s philosophy — platform facilities, no new dependencies, failures logged and swallowed:

- **macOS**: `NSUserNotification` through the already-linked ObjC runtime, so banners carry Comet's name and icon.
- **Dev runs**: unbundled processes (`cargo run`) have no notification center, so they adopt the installed bundle's identity via an `NSBundle bundleIdentifier` override — the banner still shows the comet icon rather than being attributed to Script Editor. Falls back to `osascript` if that fails.
- **Linux**: `notify-send`. **Windows**: no-op until there's a registered AppUserModelID (an installer concern); the chime still covers it.
- `COMET_DISABLE_NOTIFICATIONS` env kill-switch, mirroring `COMET_DISABLE_SOUND`.

**Trigger**: banners ride the *same* staleness-gated and send-pending-gated transitions as the chime in `shell::on_state_changed` — one detector, two outputs — so a banner can never fire where a chime wouldn't. Title is the session title; body is "Run finished" or "Waiting on your input".

**Background-only** (default on): the banner is skipped while any Comet window is focused, since the chime already covers that case. The focus check is app-level on purpose — a ping for a background *chat* in a focused app stays a chime plus the sidebar dot.

This required taking explicit control of presentation: AppKit's notification center silently suppresses banners while the app is frontmost, a policy our setting couldn't override, so turning background-only *off* would have silently done nothing in the packaged app. A center delegate now answers `shouldPresentNotification:` YES unconditionally, moving the whole show/skip decision into our code and making every platform path behave identically.

**Settings → Notifications** (new page): three toggle rows — Sounds, Desktop notifications, and Only when in the background (a sub-option that dims and goes inert when banners are off, the same treatment as uninstallable agents on the Agents page). `soundEnabled` had no UI toggle before this; it was file-only.

## Testing

- `cargo check -p comet-ui` — clean, no new warnings
- `cargo test -p comet-ui` — 386 passed, including new coverage for the settings round-trip and back-compat defaults (files written before this change load with banners and background-only on)
- Live smoke run of the unbundled path: banner delivered with the delegate installed (`delivered = 1, delegate set = true`)

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/51"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789101806&installation_model_id=425430&pr_number=51&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F51&signature=59ab69d7249fc9433d6fd967b2706e30c9836592100360d82d5f4209b3891e3e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (commented):
Reviewed the diff and ran the branch through the headless rig — Linux path end-to-end with a real notification daemon (mako) on the bus, mock runs driving the actual transitions. Both banner edges fire with the right copy, the focus gate works, and the settings page persists correctly.

## Verified live

**Run finished** — Comet tiled in the background, terminal focused; the banner carries the session title and is attributed to app-name Comet:

Run finished banner while Comet is backgrounded

**Waiting on your input** — `COMET_MOCK_QUESTION=1` run flips the session to AwaitingInput; banner posts while the QuestionPanel and the sidebar "Input" badge sit behind it:

Waiting on your input banner

**Focus gating** — with background-only on and a Comet window genuinely holding keyboard focus, a finishing run produced *no* banner (checked `makoctl list` empty); the same run with a terminal focused produced one. The one-detector claim holds: the banner block sits inside the exact transition/staleness/send-pending gate the chime uses, so no divergence is possible.

**Settings → Notifications** — new page renders with the three rows; flipping Desktop notifications off dims the sub-option and makes its toggle inert (clicked it: no state change, no event), and `ui-settings.json` round-trips `notificationsEnabled`/`notificationsBackgroundOnly` through the save debounce. `soundEnabled` finally has a UI home too.

| default | banners off → sub-option dimmed + inert |
|---|---|
| settings default | settings dimmed |

`cargo test -p comet-ui` — 386 passed on the branch. The branch is a few commits behind main but touches none of the same files; expect a clean merge.

## Finding: leading-dash titles get eaten as notify-send options

`notify-send` is invoked as `["--app-name=Comet", title, body]`, so a title that starts with `-` parses as an option. Reproduced against mako:

```
$ notify-send --app-name=Comet "--urgency=low" "Run finished"
# → summary: "Run finished" (the BODY got promoted to title), urgency silently low
```

Chat titles are model-generated, so a leading dash is reachable — worst case a `--urgency=critical`-shaped title makes banners sticky, mildest case the title/body shift. One-token fix: terminate option parsing before the positionals —

```rust
.args(["--app-name=Comet", "--", &title, &body])
```

(verified: with `--`, the same string renders literally as the title). The osascript path has the analogous-but-lesser cousin: `applescript_escape` handles `\` and `"`, but a literal newline in a title would break the AppleScript statement — the failure is swallowed, so it just drops the banner; replacing control chars with spaces alongside the existing `\0` strip would cover it.

## Idiom notes (all fine, for the record)

- `notify.rs` mirrors `sound.rs` faithfully: same module-doc structure, `DISABLE_ENV` kill-switch shape, background-thread CLI spawn, `tracing::debug` + swallow on failure.
- The settings page speaks the existing widgets vocabulary (`section_card`/`card_row`/`row_tile`/`meta_line`/`toggle_switch`), and the dimmed-inert sub-row matches the Agents-page treatment.
- Back-compat rides the struct-level `#[serde(default)]` + `Default` impl, and the added test pins it.
- The `COMET_OPEN_ROUTE` route addition and the defensive `Class::get` (instead of `class!`) for a deprecated API are both the right calls.

Nice work overall — the delegate trick to reclaim the show/skip decision from AppKit is well-reasoned and well-documented.

wingleeio:
Pushed e14ef1c for the dash-title finding: `--` now terminates option parsing on the notify-send invocation, and the AppleScript escape flattens newlines (a raw one ends the statement mid-literal, silently dropping the banner) with the test extended to pin it.

Re-verified end-to-end in the rig — renamed the session to `--urgency=critical injection probe` and let a run finish while backgrounded: the banner renders the title literally at **normal** urgency (pre-fix, this exact title was consumed as a flag). `cargo test -p comet-ui`: 386 passed.

dash-leading title rendered literally after the fix

wingleeio (approved):
Both ping edges, the focus gate, and the settings round-trip verified live in the rig (screenshots in my review); the dash-title fix is in and re-verified end-to-end. The one-detector design means banners can never drift from the chime — nice and tight. Approving.

PR #52, opened 2026-08-12
Tool chips expand to the full tool call

SinaKhalili:
## What

Clicking a tool chip in the transcript now always expands it to show the **full tool call**. Previously a chip only expanded when the call carried an output or diff payload — and even then the expansion showed only the result, never the invocation itself, which stayed a single truncated header line. A long command or an MCP input was unreachable.

## How

All in `crates/ui/src/transcript.rs`:

- **New `call_block()`** builds a full-invocation block per tool kind: the whole multi-line command for Run, full path/pattern/URL, todo items as a `[x]`/`[ ]` checklist, MCP/unknown inputs as pretty-printed JSON under the `server · tool` line. It reuses the `ToolDetail::Output` payload, so rendering and the analytic height math are the same implementation as output blocks. Long single lines soft-wrap at 80 char-counted columns (block heights must stay analytic for the fold tweens) instead of ellipsizing, capped at the output line cap with the counted "… N more lines" tail.
- **`ToolItem` gained an `invocation` field**, precomputed at row-build time (same rationale as the existing `detail` field — rows are cached by fingerprint, nothing is built per scroll frame). The row fingerprint hashes the invocation bytes so an in-place call update (e.g. a streaming MCP input) re-splices its row.
- **`render_tool_group`** treats a chip as expandable when it has an invocation *or* an output detail — in practice every chip, so every chip now carries the chevron affordance. The open card shows the invocation first (what was asked), then the output/diff (what came back), each under its own hairline separator. The "Show full output/diff" sidecar affordance, open/close tweens, and group fold behavior are unchanged; group and card heights account for both blocks.

A truly blank call (e.g. an empty command) still renders as the old plain, non-clickable chip.

## Testing

- `cargo test -p comet-ui` — 387 passed, including a new `call_block_carries_the_full_invocation` test covering multi-line commands, 80-col soft-wrap, MCP JSON pretty-printing, todo checklists, and the blank-call fallback.
- Verified live against the mock harness (`COMET_MOCK_CODE=1` multiline exec): previously-plain chips expand to the full command.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/52"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789105855&installation_model_id=425430&pr_number=52&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F52&signature=12c2a60d883792240671080ee77ada8fd469b4decaacb0fb292b902a8ab256c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/transcript.rs
+++ b/crates/ui/src/transcript.rs
@@ -226,6 +226,11 @@ pub struct ToolItem {
     /// Precomputed here because rows are cached by fingerprint — diffing and
     /// tokenizing per paint would run on every scroll frame.
     pub detail: Option<Arc<ToolDetail>>,
+    /// Expandable full-invocation block: the complete tool call (whole
+    /// command / pattern / URL / input JSON) that the chip header collapses
+    /// to one truncated line. Rendered above `detail` in the open card.
+    /// Precomputed for the same reason as `detail`.
+    pub invocation: Option<Arc<ToolDetail>>,
     /// Sidecar key of the full output (chat2-sync A3) — the doc carries only
     /// a one-line summary; expanding offers a lazy "Show full output" fetch.
     pub output_ref: Option<SharedString>,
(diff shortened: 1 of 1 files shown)


wingleeio (approved):
Reviewed the diff and took the branch for a spin in a headless rig (sway + seeded mock engine, `COMET_MOCK_CODE=1`) — clicking a previously-plain chip now expands to the full invocation, including the multiline exec the mock harness uses as the round-9 breaker.

full demo: both chips expanded

## Verified live

- **Every non-blank chip carries the chevron.** The two `Run` chips that used to be inert (no output/diff payload) now expand. Same for the `COMET_MOCK_CODE` multiline exec that previously stayed a single flattened header line.

  chips closed, chevrons present

- **A long single-line command expands to the full invocation** instead of staying truncated in the header.

  git log chip expanded

- **A multiline command expands to verbatim lines**, not the flattened `set -e fixture_in_original=0 grep …` chip line. That's the user-report shape.

  multiline exec expanded to three lines

- Open/close tween, group fold, and the sidecar "Show full output" affordance are unchanged. Blank calls still skip the block (covered by the new unit test).
- `cargo test -p comet-ui` — 402 passed, including `call_block_carries_the_full_invocation` (multi-line, 80-col wrap, MCP JSON, todo checklist, blank fallback).

Reusing `ToolDetail::Output` for the invocation is the right call — height math stays analytic, so the fold tween doesn't drift. Hashing the invocation bytes in the row fingerprint is also necessary: the one-line chip hash only covers length, and a streaming MCP input would otherwise stick on a stale row.

Nice, small, exactly the click-doesn't-expand report.

PR #58, opened 2026-08-12
Fix changes pane resizing and remove arbitrary width limits

jsgrrchg:
## Summary

- move the right-pane resize handle outside its clipped container and expand its hit target from 5px to 12px
- remove the fixed 760px and 52% viewport ceilings so the right pane can use the available workspace
- preserve a 300px minimum conversation width, including while resizing the window
- make the composer react immediately to measured width changes instead of waiting for an unrelated repaint
- expand the composer live while narrowing, while deferring collapse until resize settles to prevent layout chatter
- replace the traits summary with a compact three-dot trigger at narrow widths and restore the full effort/options summary when space returns
- preserve large persisted right-pane widths while clamping them safely to the current window

## Motivation

The Changes pane was difficult to resize because its narrow handle lived inside an `overflow: hidden` container. Even after a drag started, arbitrary fixed and viewport-relative ceilings prevented the pane from taking useful space on larger displays.

Removing those ceilings exposed two responsive-layout issues in the conversation column. The composer consumed stale input measurements, so its compact/expanded layout could lag behind the divider. Its traits control could also crowd the prompt at narrow widths or remain stuck in its compact presentation after the chat became wide again.

This update makes resizing feel direct in both directions. The right pane remains flexible, the chat retains a usable 300px floor, and the composer adapts its layout and controls from stable width measurements.

## Validation

- `cargo check -p zeron-ui`
- `cargo test -p zeron-ui --lib` (426 tests passed)
- focused tests for composer resize behavior, responsive traits presentation, and the 300px chat-width floor
- `git diff --check`

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/58"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789157817&installation_model_id=425430&pr_number=58&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F58&signature=efc63c23595649908e3a9c86fc86595d581f8f05caad2c9c318e9a3f132c47d7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
Re-audited the latest head (`6f15324`). `cargo check -p zeron-ui`, all 424 UI library tests, and `git diff --check` pass.

One cleanup remains before approval: `cargo fmt --all -- --check` reports three PR-introduced formatting differences in `crates/ui/src/settings.rs` and `crates/ui/src/shell.rs`. Running `cargo fmt --all` should resolve them.

jsgrrchg:
@wingleeio i added a few commits to harden the interactions, and i also did the cleanup

jsgrrchg:
Fixed a bug i introduced

The 300px minimum chat width was unintentionally being applied to both manual resizing and expanded takeover mode. As a result, expanding the right pane could no longer fully hide the chat column.

The behaviors are now separated:

- Manual resizing still preserves a usable 300px minimum chat width.
- Expanded takeover mode can consume the entire conversation column.
- Collapsing the pane restores its previously saved width.

I also added a regression test covering full-width takeover behavior.

jsgrrchg:
I found and fixed the high idle CPU usage introduced by the composer reflow changes.

The root cause was a redraw loop: `ComposerInput` emitted a `LayoutChanged` event from inside Taffy’s measurement callback whenever its measured width changed. Since Taffy can evaluate multiple provisional widths during a single layout pass, the event triggered another render and layout, continuously repeating the cycle even while the app was idle.

The fix removes layout invalidation from the measurement callback. The composer now tracks the stable outer width reported by the shell and schedules a follow-up render only when that width changes meaningfully, ignoring subpixel noise with a 0.5 px threshold.

This preserves the intended responsive behavior during panel resizing while eliminating the idle redraw loop.

Validation:

- Idle CPU dropped from approximately 45–57% to around 1.6–2.1%.
- All 448 `zeron-ui` tests pass.
- `cargo check -p zeron-ui` passes.
- `cargo build -p zeron` passes.
- `git diff --check` passes.

wingleeio (approved):

PR #59, opened 2026-08-12
Sessions: quiesce watchdog + fold self-continued output (stuck-Working fix)

wingleeio:
## The bug (diagnosed live, 2026-08-12)

A session read **Working for 19 minutes after the agent finished**. Root cause chain, established from the agent's session data, comet's doc journal, and the engine log:

- The engine only settles a turn on the harness's `Done`; for ACP that means the `session/prompt` response. That response was lost upstream (agent finished with `stop_reason: end_turn` at 21:03:32; child idle, zero open sockets, empty queue — no turn-end ever arrived).
- Nothing recovers from a lost turn-end: the liveness heartbeat keeps the session row fresh (so the UI's 45s staleness gate never fires), and there's no per-turn timeout by design. The same symptom was reported on a Codex session — the trap is in the shared engine contract, not one adapter.
- The same incident exposed **transcript content loss**: Claude Code re-invokes itself when background tasks finish (turns no prompt started), and the parked gate dropped that streamed output. "Build finished successfully…" / "Confirmed — Xcode 26.2…" exist in the agent's session data and are absent from the chat doc, exactly in the 20:45:43→20:55:25 parked window.

## The fix (engine-level, harness-agnostic)

**Turn-quiesce watchdog** — when the stream is silent past the window (default 120s; `COMET_TURN_QUIESCE_MS` overrides, `0` disables) and the fold shows nothing in flight (no unresolved tool call — live-plan chip exempt — and no open question), the turn parks exactly like a `Done` would: segment finalized `Complete`, status Idle, child + mailbox warm. It never ends the run or errors anything — this is not the stall timeout the design comment rejects. A false trip costs a brief status dip, because…

**Parked self-continuation** — fresh text or a genuinely new tool call on a parked session now resumes it (new segment, Working) instead of being dropped, and the resumed turn settles again via `Done` or the watchdog. A 1s resume gate keeps a finished turn's tail flush inert (arrives within ms of the park; real self-continuation is a whole new agent round trip — 5 minutes in the incident), and stale tool echoes stay gated via `seen_tools`, so the original eternally-running-session fix holds.

The two changes are mutually load-bearing: resume is only safe because the watchdog guarantees a re-settle; the watchdog is only safe because resume makes a false park lossless.

## Verification — the incident workflow, mocked end-to-end

`crates/engine/tests/turn_quiesce.rs` drives a feed-by-hand harness through the exact observed sequence:

1. `parked_self_continuation_folds_and_requiesces` — turn completes → parks; agent self-continues with no prompt → output folds into the doc as its own Complete entry (this exact text was lost in the incident), Working while it streams, watchdog re-parks.
2. `missing_turn_end_settles_instead_of_working_forever` — steer ("what about now") becomes the next turn, the answer streams, its `Done` is lost → watchdog settles instead of Working-forever; the answer survives in the doc.
3. `open_tool_call_never_quiesces` — an unresolved tool call holds the watchdog off through many windows of silence (the legitimately-quiet case the no-stall-timeout comment protects).
4. `stale_tool_echo_stays_parked` — post-turn noise still never re-arms a parked session.

`cargo test -p comet-engine`: all suites green, including the pre-existing `parked_session_ignores_trailing_frames_and_stays_idle` contract test (the resume gate is what keeps it green). `repos_round_trip_add_branches_worktrees` fails identically on `main` (pre-existing, environmental).

Not addressed here (upstream): the adapter losing the `session/prompt` response for a prompt delivered into a merged turn, and settling prompts while background tasks are pending. This PR makes comet correct and lossless regardless of which layer drops the ball.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/59"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789169137&installation_model_id=425430&pr_number=59&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F59&signature=ab189d1167f39f605bcce0c8d8ec6dae7cc4389f755306df1e43a55c579333a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/engine/src/sessions.rs
+++ b/crates/engine/src/sessions.rs
@@ -1119,6 +1119,28 @@ async fn drive_run(
     const SESSION_IDLE: std::time::Duration = std::time::Duration::from_secs(30 * 60);
     let mut idle_since: Option<tokio::time::Instant> = None;
     let steerable = harness.supports_steering();
+    // TURN-QUIESCE WATCHDOG (2026-08-12 stuck-Working incident): a harness
+    // that loses a turn's Done — the adapter never settles `session/prompt`
+    // even though the agent finished — strands Working forever: the live
+    // heartbeat above keeps the row fresh, and there is no per-turn timeout
+    // by design. This is NOT that stall timeout: it never ends the run or
+    // errors anything. When the stream has been silent past the window AND
+    // the fold shows completed output with nothing in flight (no unresolved
+    // tool, no open question), the turn parks exactly like a Done would —
+    // segment finalized Complete, status Idle, child and mailbox warm. A
+    // false trip (the agent was quietly waiting on something invisible)
+    // costs a status dip: the parked-resume path below re-arms Working the
+    // moment output flows again, and nothing is lost. `COMET_TURN_QUIESCE_MS`
+    // overrides the window; 0 disables.
+    let quiesce_after: Option<std::time::Duration> = match std::env::var("COMET_TURN_QUIESCE_MS")
+        .ok()
+        .and_then(|v| v.parse::<u64>().ok())
+    {
+        Some(0) => None,
+        Some(ms) => Some(std::time::Duration::from_millis(ms)),
+        None => Some(std::time::Duration::from_secs(120)),
+    };
+    let mut last_stream_activity = tokio::time::Instant::now();
 
     let final_status = loop {
         let event: AgentEvent = tokio::select! {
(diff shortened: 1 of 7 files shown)


wingleeio:
Second commit (430dbec) upgrades this from mitigation to root-cause fix, after live verification:

**Reproduced the drop against the real adapter** (`scripts/acp-starve-repro.py`, no comet code): a `session/prompt` sent while the Claude CLI runs a self-continued turn (background-task re-invocation) never gets its response from `claude-agent-acp@0.66.0`. Steering answers `promptRequired`/`noRunningTurn` *while the CLI is visibly mid-turn* — the adapter doesn't track turns it didn't start, so the merged turn's result is discarded as unowned and the pending prompt starves. The prompt's text still reaches the model via the CLI's own queue.

**Cross-referenced Zed** (same adapter package): Zed's send discipline (UI-side queueing; cancel → await old response → prompt on interrupt) narrows the window but doesn't close it — any ACP client that prompts into a self-continued turn starves.

**New in the harness:** `noRunningTurn`-with-a-prompt-outstanding is protocol *proof* the prompt can never settle. After a 2s grace (covers the legitimate turn-end race), the dead turn is closed with a `Done` and the queued steer is promoted to a fresh prompt — verified to settle against the real adapter, and covered by `scenario:starve` in the fake-adapter suite. No adapter patch, no guess-based timers: the wedge now self-heals end to end (watchdog settles status; the next message repairs the turn state).

PR #63, opened 2026-08-13
Changes header: keep the expand/close buttons still going into takeover

wingleeio:
Fixes the reported bug where the right-pane header buttons (fold-all, expand, close) shift slightly when entering expanded/takeover mode.

## Root cause

The trailing titlebar strip (pane-header controls + expand/close buttons) is right-pinned by a `flex_1` spacer inside a row with 8px child gaps, and its width is capped to `viewport − row_left − pr`. The gaps live *outside* the strip's width, so the cap was 8px too generous. Normally the pane is narrower than the cap and the spacer absorbs the gap — buttons stay pinned at `viewport − pr`. In takeover the strip spans everything right of the sidebar, the cap binds exactly, the row overflows by one gap, and the whole button cluster slides 8px right.

The fix budgets the row's gaps into the cap (one gap in takeover, two while the title row is present). The takeover `row_left` tuning already assumed the gap sits before the strip, so the scope dropdown's left alignment is unchanged.

## Verification (headless rig, 2x)

Icon stroke columns in the physical 2560px screenshot, pane open → expanded:

| | fold | expand | close |
|---|---|---|---|
| before fix | 2367 → **2383** | 2424 → **2440** | 2485 → **2502** |
| after fix | 2367 → 2367 | 2424 → 2424 | 2485 → 2486 |

Exactly the 8-logical-px gap, gone after the fix. Red line = fixed reference column:

before/after

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/63"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789234473&installation_model_id=425430&pr_number=63&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F63&signature=37c08306ad22ee6cc8bf803447a8936bcd610d4b1db872d2bad58c6266b251ee"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
Two more fixes on this branch (both user-reported):

**Working trailer sending bridge** (29b76c3) — the loader under the streaming reply based its timer on the in-flight send, so it counted the send round-trip and restarted from zero when the turn actually began. It now shows **Sending…** with the same spinner and no timer until the session row's turn start outdates the send; the flavour word + timer then take over, counting from the turn. Verified by SIGSTOPping the daemon mid-send: "Sending…" held with no timer at +5s, then "Plotting… 2s" right after resume. Unit test on the bridge predicate included.

**Even ref truncation in the Branch-changes header** (2d4b51e) — both labels shrank with the default flex factor, which splits the deficit proportionally to content width, so a long branch stayed near-whole while a short base ("main") collapsed to a bare ellipsis. Each side's `flex_shrink` is now weighted by its own length squared (mono font, chars ∝ px): the deficit lands ~cubically on the longer name — the short side stays whole (a linear weight still cost it a char under a ~200px deficit; middle row below) and equal lengths split evenly.

sending + refs

wingleeio:
Two more changes on the branch:

**Resume id survives helper startup crashes** (f186be7) — the failed-resume fallback fired on "errored Done before SessionStarted" and tombstoned the chat's harness session id (permanently — it survives restarts). Since the ACP conversion a stale id falls back *inside* the harness (`session/load` → `session/new`), so that signature only ever means the agent child failed to start — and the tombstone was severing good conversations. The id now stays; the once-only retry keeps the resume; and a child that dies before the handshake surfaces its exit status + stderr tail in the journaled error instead of a bare "transport closed". Tests remodeled to cover both the transient-blip retry (resume kept) and the down-hard case (id intact, no revival loop).

**Every send reserves the reply's space** (ecb88a6) — the prompt-at-top runway existed only for the first send; a steer or later send collapsed it and dropped the messages down. Now every send glides the prompt to the top (~230ms ease-out) and reserves the space below, notes-app style: the runway is a true min-height for the turn (`usable − turn_height`, recomputed post-layout), so streaming consumes it with zero net motion, short finished turns hold stable, and the bottom-spring handoff at overflow is height-neutral. Revisiting a session still lands at the bottom.

Demo — send after a finished turn, steer mid-stream, leave + revisit:

https://github.com/user-attachments/assets/41bedba3-f012-4ea8-8b0b-7c63a2812ae6

PR #68, opened 2026-08-13
Message queue: hold what you type while the agent is busy

danperks:
## What

Typing while an agent is mid-turn used to be a gamble. Agents that accept mid-turn input got the message; the rest had it posted into a run mailbox that a cancelled or errored turn quietly discarded — leaving a bubble in the transcript that looked sent and never was.

Messages now land on a `queue` MovableList on the session doc. They show as pending above the composer on **every** device, can be reordered, retyped or dropped from either, and only the host takes from them.

## How it drains

`DocHost::drain_queue` runs on the host on every doc change, plus on a session-status watcher (a turn *ending* isn't a doc change the queue can see):

- **Idle** → send the head as the next turn, and loop.
- **Working, agent takes mid-turn input** → steer it in, then stop; the rest waits for the turn's end.
- **Working, anything else** → hold.

Attachments never steer — the steer path carries a prompt and nothing else, so a message with files waits for a turn that can inline them rather than losing them.

"Send now" interrupts the running turn to jump one message to the front. That's also what Enter on an empty composer means when messages are waiting. It's guarded by a `sending_now` flag: the interrupt makes the chat read Idle, and an idle chat with a queue is exactly what the flush drains, so without it the override would take the rest of the queue with it.

A row is only removed from the queue once the send is known to be going out, and it's pushed back on failure — a row that survives a failed send is recoverable, a vanished one isn't.

## Why a MovableList

Reordering is a real move op. Two devices dragging at once converge on one order with every row intact, where a delete+insert list would duplicate or lose rows. There's a test for that.

## Surface

- **doc** — `crates/doc/src/queue.rs`: push / read / take / move / set-text on the `queue` container.
- **rpc** — `WatchQueue`, `QueueMessage`, `UpdateQueuedMessage`, `MoveQueuedMessage`, `RemoveQueuedMessage`, `SendQueuedMessageNow`. All host-addressed, same as the command ledger next door.
- **engine** — queue ops + drain/flush on `DocHost`; `steers_mid_turn` / `turn_in_flight` on `SessionsEngine`. The existing `Steer` command arm now routes through a shared `deliver_prompt`, which queues rather than mailboxes for turn-boundary agents.
- **desktop** — `crates/ui/src/queue.rs`: the panel above the composer, with drag-to-reorder, arrows, edit-in-composer, send-now and discard. Editing a row to nothing is how you drop it.
- **iOS** — the same panel and the same doc writes, so the phone is a first-class participant rather than a viewer.

## Test plan

- [x] `cargo test -p comet-doc --lib queue` — 5 tests incl. concurrent-reorder convergence
- [x] `cargo test -p comet-engine --test message_queue` — 8 tests: idle drain, hold-and-flush-in-order, steer, attachments-never-steer, send-now interrupt, edit-to-empty, RPC reorder + stream
- [x] `cargo test -p comet-ui --lib queue`
- [x] `cargo test --workspace` — only pre-existing failures (`repos_round_trip_add_branches_worktrees` `/private/var` symlink, `falls_back_when_interactive_attempt_hangs`); both reproduce identically on a clean `origin/main` worktree
- [x] `xcodebuild test` on iPhone 17 sim — TEST SUCCEEDED, incl. 6 new `MessageQueueTests`

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/68"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789250337&installation_model_id=425430&pr_number=68&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F68&signature=2e687ac4259e8e046a9300d40279016fbe595b0e971f8f83237bf4e2d7a1f7ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (commented):
Took the branch for a spin in a headless rig (sway + seeded mock engine). The panel itself looks right — docked on the composer, one-line rows, hover affordances — and the doc/RPC surface is the right shape. Two engine races need fixing before this is the honest “held, not sent” story.

queue panel above composer while the agent is working

## Verified live

- Messages typed (or `QueueMessage`’d) while a run is live land in **QUEUED N queued** above the box, flattened to one line, numbered in send order.
- Hovering a row reveals the down-arrow (and the rest of the action cluster). Edit-to-empty as delete is a nice gesture.

  hovering a queued row shows the reorder arrow

- Attachment-bearing rows hold even on the mock (StepBoundary) — that matches `drain_queue`’s “attachments never steer” rule.
- `cargo test -p comet-doc --lib queue` — 7 passed, including concurrent-reorder convergence.
- `cargo test -p comet-ui --lib queue` — 2 passed.
- `cargo test -p comet-engine --test message_queue` — **flaky**. `queue_holds_during_a_turn_and_flushes_in_order_at_its_end` failed 3 of 6 isolated reruns on this box (see below).

## Finding: turn-end flush can empty the whole queue

`drain_queue` is invoked from the doc-change task *and* the session-status watcher, with no per-chat lock. `sending_now` only covers the send-now interrupt gap. Two overlapping drains both see Idle (or both see Working + steerable), both `take_queued` a different head, and both dispatch.

That’s exactly what `queue_holds_during_a_turn_and_flushes_in_order_at_its_end` asserts against:

```237:241:crates/engine/tests/message_queue.rs
    assert_eq!(
        queue_texts(&core),
        vec!["third"],
        "only the head goes: the next turn is now running"
    );
```

On failure the left side is `[]` — `second` *and* `third` went out. Live, three rapid `QueueMessage`s while mock was Working all showed up as transcript bubbles (the “looks sent” shape this PR is meant to kill) instead of sitting in the panel:

three queued prompts rendered as sent user bubbles

A drain-in-flight flag (same idea as `sending_now`, but held across the whole `drain_queue` body) would serialize this. The idle `loop` is fine once only one drain runs at a time — after `dispatch` the next iteration re-reads status.

## Finding: `AwaitingInput` is not “busy”

```1922:1924:crates/engine/src/doc_host.rs
            let busy = sessions
                .session_status(&handle.chat_id)
                .is_some_and(|s| matches!(s.status, SessionStatus::Working));
```

Composer queues whenever `run_live` (Working **or** AwaitingInput). `turn_in_flight` / `held_until_turn_end` already treat a question as in-flight. `drain_queue` and the send-now interrupt check do not — a follow-up typed while the agent is parked on a question is `NextTurn`’d. `busy` should be `sessions.turn_in_flight(&handle.chat_id)` (and the interrupt arm should match).

## Nit

- `queue_rpc` assigns `self.queue_task = Some(cx.spawn(...))`. A second click drops the first Task and can cancel the in-flight MOVE/UPDATE. Rapid arrow taps will desync the optimistic list from the doc. Don’t cancel; queue or join.
- Failed `dispatch_queued` `push_queued`s the row back on the **tail**, so a send-now failure buries the urgent message.
- Desktop rows don’t show an attachment glyph; iOS does. Empty-composer Enter pops the head on desktop (`queue_pop_head`) but iOS `send()` still no-ops on an empty box — `sendQueueHeadNow` exists and isn’t wired.

The MovableList + host-only take is the right split, and the panel chrome matches the composer. I’d land the drain serialization (and the AwaitingInput check) before calling the original report fixed.

danperks:
Thanks — both engine findings were real. Fixed in 2888956.

## Drain serialization

You were right that `sending_now` was the wrong shape: it only ever covered the idle window a send-now's interrupt opens, not the drain body. Replaced it with a per-chat `drain_lock` held across the whole of `drain_queue`, and across the take+dispatch in `send_queued_now`. That subsumes what the flag was doing, so the flag is gone rather than sitting alongside a lock that already covers it.

The idle `loop` is unchanged, as you said — with one drain running at a time the next iteration's status re-read is enough.

I couldn't reproduce the flake on this box (6/6 green before the fix), so I pinned it deterministically instead of trusting timing — `concurrent_drains_release_one_message` drives three `drain_queue` calls concurrently against a queue of three with an idle agent. Confirmed it fails without the lock:

```
assertion `left == right` failed: one drain released the head; the others found a busy agent
  left: ["third"]
 right: ["second", "third"]
```

That's your exact failure mode: two messages escaped, one release expected.

## AwaitingInput

Agreed, and it was inconsistent in two places, not one. Both `drain_queue`'s `busy` and the `QueueSend::Interrupt` arm in `dispatch_queued` now use `sessions.turn_in_flight(...)`, matching `held_until_turn_end` and what the composer queues on.

Pinned by `a_message_holds_while_the_agent_waits_on_a_question`, which needed the mock to park on a real question — an `InputRequested` synthesized into the stream gets dropped as an unknown id, so it goes through `controls.request_input`. Verified it fails against the old `Working`-only check (`left: []`, the follow-up had already gone out as a fresh turn).

## Nits

- **`queue_task` cancellation** — fixed. These are independent one-shot mutations, so they `.detach()` now instead of sharing a slot; the field is gone.
- **Failed dispatch buries the row** — fixed. Added `SessionDoc::insert_queued(index, item)` and both restore paths use index 0, so a failed send never reorders the queue behind the user.
- **Desktop attachment glyph** — added, matching iOS.
- **iOS empty Enter** — not wired, and I removed `sendQueueHeadNow` instead. On iOS that button is also stop, gated on `showStop && !hasContent`, and the queue is only visible while `runLive` — which is exactly when `showStop` is true. Making a waiting queue count as content would replace stop with send for the whole of every run. The per-row send-now button already covers the gesture, and there's no Enter-to-submit on the multi-line iOS composer for the desktop shortcut to mirror. Happy to reconsider if you'd rather have it behind something else.

## Tests

`comet-doc` queue 9 passed (2 new), `comet-engine` message_queue 10 passed (2 new, 5 consecutive clean runs), `comet-ui` queue 2 passed, iOS 28 passed including `MessageQueueTests`.

Full workspace: 760 passed, 2 failed — `repos_round_trip_add_branches_worktrees` and `shell_env::unix::tests::falls_back_when_interactive_attempt_hangs`, both of which also fail on a clean `origin/main` worktree here and are untouched by this branch.

jsgrrchg:
I tried this branch, but couldn't use it because Codex messages go straight to steering, same with Claude. Could it make sense to add a setting that controls how messages submitted during an active turn are handled? 

With step-boundary providers such as Codex or Claude, a newly queued message is immediately removed from the queue and delivered as steering input. In practice, the queue panel often never becomes visible, so users cannot reorder, edit, discard, or intentionally hold messages—the main interactions this PR introduces.

One possible design:

- **Automatic steering**: preserve the current behavior.
- **Always queue**: keep every message submitted during an active turn in the visible queue, regardless of whether the provider supports mid-turn steering.

This would make the queue useful and predictable across all providers while preserving immediate steering for users who prefer the existing behavior. Right now, provider capability implicitly determines the UX, and with steerable providers the queue can appear not to work at all.

jsgrrchg:
Also, stoping inference with pending queue, sends inmediatly the next message, it should be freezed and resume until the next turn.

danperks:
@jsgrrchg I added a Behaviour settings menu with a toggle in - I had it locally but I have a bunch of other toggles in that menu that aren't on main, so it didn't make it in.

PR #70, opened 2026-08-13
Runway: input restick re-arms the hold, never the bottom pin

wingleeio:
Fixes the report that the runway wasn't working as a runway: streaming text sat pinned near the viewport top instead of filling the reserved space, and the sent message never went to the top.

## Root cause (rig-reproduced with a trace)

Wheel/touch input releases the hold as designed — but the old escape/restick rules then re-engaged the **bottom pin**, whose bottom is the bottom of the reservation pad. The spring parks there, the offset re-glues, and layout hard-tracks the stale pad end on every streaming commit. In that state:

- the stream tail renders at a fixed spot near the viewport top (the pad below it never fills),
- the prompt is scrolled away above,
- and the pad can't resize — its sizing needs the anchor row's bounds, which is off-screen.

Trace of the broken state: `held=false pinned=true glued=true dist=0`.

**Why it always broke on the Mac and never in the rig**: macOS trackpad momentum. Leftover momentum events arriving right after clicking send release-and-restick within a single gesture, so the prompt never landed at the top at all. The rig's virtual pointer emits no momentum.

## Fix (notes-app parity: returning to the bottom returns you to the reservation)

While a runway lives, the bottom pin stays out of input handling entirely:

- **escape** releases the hold and resets the spring; the reservation stays behind as plain scrollable space;
- **restick** re-arms the *hold* (fresh glide re-lands the prompt at its inset) instead of setting `pinned` — and the hold's absolute per-layout re-assert keeps glue out;
- momentum-down right after a send hits the `!held` gate and leaves the live hold alone;
- the jump pill shows while released and still retires the runway on click.

## Verified

Send with momentum scroll spam → prompt lands at top, reply streams downward into the runway; wheel-up then wheel-down mid-stream → hold re-lands (trace: `held=true pinned=false`) instead of freezing tail-at-top; completion holds stable. 394 UI tests pass.

https://github.com/user-attachments/assets/5a6aadf1-b297-43ec-9aef-0b8af33818ca

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/70"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789252887&installation_model_id=425430&pr_number=70&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F70&signature=14eeae0ab4ee8b9d528001eefe1a350144cc0b528482b82896f02d8bf0278ddd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/transcript.rs
+++ b/crates/ui/src/transcript.rs
@@ -1493,15 +1493,52 @@ impl Transcript {
         let this = cx.weak_entity();
         cx.defer(move |cx| {
             this.update(cx, |this: &mut Transcript, cx| {
-                // Wheel/touch input RELEASES an own-turn hold (clamping it
-                // locked scrolling for as long as the reservation held —
-                // user report). The reservation stays behind as plain
-                // scrollable space; normal escape/restick applies below.
-                if let Some(anchor) = this.own_turn.as_mut()
-                    && anchor.held
-                {
-                    anchor.held = false;
-                    this.own_turn_last_tick = None;
+                // Wheel/touch while a runway lives: input owns the viewport,
+                // and the BOTTOM PIN must stay out of it entirely. Escaping
+                // releases the hold (the reservation stays behind as plain
+                // scrollable space); returning toward the bottom re-arms the
+                // HOLD, never `pinned` — a restick pin glued the view to the
+                // bottom of the reservation pad, where streaming reads as
+                // text stuck at the viewport top with the runway never
+                // filling (user report; the pad can't resize there either,
+                // its anchor being off-screen). macOS trackpad momentum can
+                // even release-and-restick within one gesture right after a
+                // send, so under the old rules the prompt never landed at
+                // the top at all.
+                if this.own_turn.is_some() {
+                    let distance = this.distance_from_bottom();
+                    let previous = this.last_scroll_distance;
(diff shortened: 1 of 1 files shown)


wingleeio:
Follow-up commit locks in the lifecycle per spec: **only navigating away and back clears the runway.** Scroll up/down releases and re-arms the hold; the jump pill now glides back to the runway hold instead of destroying it; rail navigation releases the hold but keeps the pad. Rig-verified: pill click from deep scroll re-lands the prompt at top with the reply still streaming into the reservation.

PR #72, opened 2026-08-13
Runway: overshoot the provisional pad — short chats report no bounds

wingleeio:
Fixes the v0.1.56 report: the prompt doesn't go to the top at send (most visibly a new session's first send), though scrolling "repairs" the runway immediately.

## Root cause (rig-traced)

gpui's bottom-aligned list reports **no item bounds while its content is shorter than the viewport** — `bounds_for_item(0)` stayed `None` for a row plainly visible on screen. The provisional pad deliberately *undershot* by a 160px turn allowance, which in a one-row chat leaves total content shorter than the viewport:

- no scroll room exists, so the bounds-free `scroll_to` hold **clamps** — the prompt sits ~150px below its inset;
- the refinement that would grow the pad to exact is bounds-gated — it can never run;
- wedged, until any scroll input re-arms the entry against a by-then-taller transcript (hence "repairs itself when I scroll").

## Fix

The undershoot was a spring-era artifact (overshoot used to carry the anchor off-screen where the sizing went blind). Under the absolute hold that risk is gone — `scroll_to` pins the prompt regardless of bounds — so the provisional pad now **overshoots at full `usable`**: scroll room is guaranteed at every chat size, the hold lands bounds-free, and the surplus sits invisibly below the fold until the refinement (or the growing reply) trues it.

## Verified

New-session first send lands the prompt at the top and holds through the reply; existing-chat sends, momentum-during-send, scroll-away/restick re-arm, and streaming-into-the-runway all re-verified unchanged. 394 UI tests pass.

https://github.com/user-attachments/assets/a162361f-4651-490a-bc60-5079d0e5bbff

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/72"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789254771&installation_model_id=425430&pr_number=72&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F72&signature=0553f3c3314e4627cf9cac231ef8cce0e98e17dba898e6e0d65a15a20b85aa10"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/transcript.rs
+++ b/crates/ui/src/transcript.rs
@@ -114,6 +114,16 @@ pub const GLIDE_MAX_VIEWPORTS: f32 = 2.5;
 /// The titlebar overlays the full-height list, so its height is part of the
 /// inset; the extra 10px matches the first row's breathing room.
 const OWN_SEND_TOP_INSET_PX: f32 = Theme::TITLEBAR_HEIGHT + 10.0;
+/// Scroll slack kept UNDER the reservation: the held layout stays this many
+/// px taller than the viewport, so the list never drops into its
+/// shorter-than-viewport regime — where a bottom-aligned list reports no
+/// item bounds (sizing goes blind), scroll_to clamps, and the prompt's
+/// position becomes a function of content height instead of the hold
+/// (rig-traced: the prompt crept up while chips streamed in and dropped
+/// back down when the turn completed and the trailer vanished). Well inside
+/// the 70px restick band, so returning to "the bottom" still re-arms the
+/// hold.
+const OWN_SEND_SCROLL_SLACK_PX: f32 = 24.0;
 /// Per-60fps-frame fraction of the remaining entry glide retained (~90%
 /// covered in ~230ms, ease-out).
 const OWN_SEND_GLIDE_RETAIN: f32 = 0.85;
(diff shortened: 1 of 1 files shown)


wingleeio:
Second commit fixes the drift wing spotted in the demo: the exact reservation put the held layout *shorter than the viewport*, dropping the list back into its no-bounds regime where content height (not the hold) governs position — hence the creep up during streaming and the drop at completion. The reservation now over-fills by a 24px scroll slack so the layout stays scrollable and the absolute hold always wins; and row 0 no longer double-counts the titlebar chrome, so first and later sends land identically. Rig-measured: bubble top pixel-identical (y=78) at send, mid-stream, and done — for both the new-session first send and a follow-up send against a real claude stream.

wingleeio:
Fresh demo on the current branch (both commits) — new-session first send, stream to completion, then a second send. The prompt bubble's top measures y=78 in every sampled frame across both turns: no low landing, no creep during streaming, no drop at completion.

https://github.com/user-attachments/assets/76b6bf01-d3b8-46bb-8415-652be2ec6c61

PR #79, opened 2026-08-14
Seamless local→synced switch with one-time import wizard

wingleeio:
Enabling sync no longer requires quitting and reopening Zeron, and a device's local-first work can come along into the synced workspace. Follow-up to #30.

https://github.com/user-attachments/assets/f05f59fd-94cf-4ced-ac08-91e2aa4215ab

## What changed

**In-place switch.** Sign-in on a local workspace now ends in a switch wizard instead of the "quit and reopen" notice. The switch is the exact mirror of the sign-out transition that #30 already shipped: stop the local runtime, wait for it to release the IPC port and data-dir lock, re-bootstrap — the window never closes. On this branch the swap plus a 2-session import completes in well under a second (the burst screenshots couldn't catch the intermediate states at 2 fps; the progress step is visible with larger profiles).

**One-time import.** The wizard's offer step ("Bring my work" / "Start fresh" / "Later") drives a new engine-side importer that carries the local profile into the synced one through the same write paths live mutations use — nothing bypasses the existing invariants:

- chat docs copy as `save_snapshot_with_cursor(id, bytes, cursor=0, epoch=2)` — the born-chat2 shape, so the first room join pushes the doc's full update log from VV zero (the existing first-contact path) and `DocHost::open` never routes an import down the s2 discard-and-adopt branch
- registry chat/space rows land as live upserts via `WorkspaceHost::import_chat_row` / `import_space_row`, with `room_gen` forced to 2
- run journals + resume budgets copy per chat; the command ledger merges (`INSERT OR IGNORE`) so imported pending commands can never re-execute under the new profile
- attachments are not copied or rewritten: transcripts embed absolute paths under the local uploads root, so that root becomes a read-only jail root of the synced profile at runtime, re-armed on later boots by a `local-import.json` marker — the same pattern as the legacy-uploads adoption
- idempotence is structural: rows already present in the target are skipped, which also gives delta imports for free (work local for a weekend after signing out; signing back in offers exactly the new sessions)

**RPC:** `LocalImportStatus` (unary) + `ImportLocalWorkspace` (progress stream), served only by synced runtimes.

**Fallback.** The old restart dialog is still there as the failure path: any error stopping the runtime, booting the replacement, or finishing the import surfaces "Sync needs a restart" — the local profile is untouched throughout, so quit-and-reopen always remains a safe exit.

## The flow

| Local-first work | Wizard after sign-in | Done |
|---|---|---|
| local | offer | done |

| Synced workspace, transcripts intact | Sign out… | …and back to local, in place |
|---|---|---|
| synced | signout | local again |

The round trip is fully in-place now: local → sign in → (import or fresh) → synced → sign out → local, no process restart anywhere on the happy path, and the local profile survives untouched as the sign-out landing zone.

## Testing

- `cargo test --workspace` green (~650 tests; includes 3 new integration tests in `crates/engine/tests/local_import.rs`: full import correctness incl. doc bytes/cursor/epoch + journals + marker, no-local-profile no-op, and delta import after a later signed-out stretch)
- new UI unit tests for the switch state machine (`switch_lifecycle_survives_the_runtime_replacement_window`, wizard offer transitions) plus a wire-shape regression test for the import progress events
- live headed run on the Linux rig (video above): local boot → 2 mock sessions → sign-in (WorkOS-mode auth against a stub edge, real loopback-callback flow) → org creation → wizard → import → synced workspace with transcripts → sign-out → local, verified at each step with `EngineInfo` probes and on-disk profile/marker inspection

## Notes for review

- `sync_flow_after_auth` keeps the switch states driver-owned so the scope `None` window mid-replacement can't reset the wizard; unit-tested.
- serde gotcha worth knowing elsewhere: `#[serde(rename_all)]` on an enum renames *variants*, not struct-variant *fields* — the import stream initially shipped `imported_chats` while the UI read `importedChats`. Fixed with `rename_all_fields` + a regression test that asserts the wire shape.
- Pre-existing nit, unchanged here: the lazy sign-in loopback listener survives an in-place runtime swap, so a *pinned* `ZERON_CALLBACK_PORT` can't rebind on a later sign-in in the same process. Production uses an ephemeral port (`callback_port: None`), so the authorize URL always points at the live listener — rig-only annoyance, noting it for a future cleanup.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/79"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789302800&installation_model_id=425430&pr_number=79&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F79&signature=925ed2f2a341817218f941e7c8f7f0f865676d34fc57d08d5a7b86507c3ed0b8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (commented):
## Blocking review

I did not find backdoors, hidden command execution, credential exfiltration, dependency changes, or an obvious upload-jail escape in this patch. The journal filename sanitizer and canonicalized attachment-root checks are appropriate. I did find the following correctness and data-durability blockers:

1. **Partial import failures are rendered as success.** `LocalImporter::run` deliberately collects per-space, per-chat, and command-ledger failures in `ImportEvent::Summary.errors`, records the import marker, and returns `Ok(())`. But `Shell::apply_import_event` never reads `errors`; every summary transitions to `ImportDone`, whose dialog says “You’re all set.” This also bypasses the documented restart/retry fallback because the stream ended normally. A failed or partial migration can therefore be presented as complete. Please make any non-empty error summary enter an explicit failure/retry state, and add a regression test that injects a per-item failure.

2. **The attachment-access marker is neither durable nor failure-aware.** `record_import` updates the shared `local-import.json` with truncate-in-place `std::fs::write`, silently treats malformed JSON as an empty marker, and only logs write/serialization failures. A crash/short write can erase every account grant; concurrent imports can also lose one account through the read-modify-write race. Imported transcript attachments then work only for the current runtime and stop resolving after restart, while the UI still reports success. Please serialize marker updates, publish them atomically (temporary file + rename, with appropriate durability), return persistence failures to the import summary, and cover corrupt/concurrent/write-failure cases.

3. **Spaces-only local work is silently excluded from the import choice.** The offer uses only `self.state.chats.len()` to compute `has_local_work`. A local profile with one or more spaces/projects and zero chats receives “Switch now,” which calls `start_synced_switch(false)` and never imports those spaces. This patch adds `LocalImportStatus` with both `availableChats` and `availableSpaces`, but the UI never calls it. Please drive the import decision from both resource types (and use the target-deduplicated status once the synced runtime is available), with a spaces-only regression test.

4. **The submitted head fails the formatting gate.** `cargo fmt --all -- --check` reports changes across the new engine/import/UI code. Please run `cargo fmt --all` and commit the result.

Validation on head `451b6ea76bd30d02c28c83953756826aadc982ee`:

- `cargo test -p zeron-engine`: 152 passed, 3 expected environment/token-dependent tests ignored.
- `cargo test -p zeron-ui --lib`: 423 passed.
- Focused `cargo test -p zeron-engine --test local_import`: 3 passed.
- `git diff --check`: passed.
- Prospective merge tree against current `origin/main`: conflict-free.
- `cargo fmt --all -- --check`: failed as noted above.

This is a blocking review, not an approval.

wingleeio:
All four blockers addressed as of `c67146f` (fixes in `4468a70`, then a merge of current `main`):

**1. Partial imports now fail loudly.** A summary carrying any error drives the wizard into an explicit **"Import didn't finish"** step with *Retry import* / *Not now* — never "You're all set". A stream that dies before its summary lands in the same state, so there is no path where a partial migration presents as complete. Retry is in-place and safe because the import is structurally idempotent; the injected-failure regression test (`per_item_failures_surface_in_the_summary_and_leave_the_row_retryable`) also asserts the failed chat's registry row stays absent, so the retry picks it up rather than skipping a half-imported chat. `import_summary_outcome` is pure and unit-tested, including the errors-field-missing (older engine) case.

**2. Marker is durable, serialized, and failure-aware.** `record_import` now publishes via temp file + `fsync` + atomic rename (a crash or short write can never truncate away other accounts' grants), serializes the read-modify-write behind a process-wide lock (cross-process exclusion is the data-dir `InstanceLock`'s job), and returns persistence failures into the summary — which the UI now surfaces as a failure, since a grant that only lives in memory breaks imported attachments on the next restart. A corrupt marker is moved aside to `local-import.json.corrupt` instead of being silently clobbered (its grants were already unreadable at boot, so keeping evidence beats pretending). New tests: corrupt-marker aside + rebuild, marker-path obstruction surfacing in the summary, and per-account entries surviving a second account's import.

**3. Spaces count as local work.** The offer now derives from chats **and** spaces via `local_work_phrase` (unit-tested): a projects-only profile gets the full import choice with copy naming what it brings ("the 1 project", "the 2 sessions and 1 project"), and a new engine test covers the spaces-only import end to end. On the parenthetical: the offer necessarily reads the local runtime's own registry — the synced store isn't open yet (single-owner data dir), so `LocalImportStatus` can't answer pre-switch. The run itself is target-deduplicated and the done step reports imported vs already-present, which is that status, one screen later.

**4. fmt.** `rustfmt --check` is clean on every file this branch touches. I scoped the run to the branch's files rather than `--all` — on my rig a repo-wide pass drags in unrelated `main` files (rustfmt 1.9.0-stable here); happy to run `--all` instead if your toolchain agrees it's clean.

Also merged current `main` (one conflict: `uploads.rs`, where the R2 mirror removal met this branch's `RwLock` read-roots — resolution keeps the mirrorless shape with the runtime-grantable roots).

Validation on `c67146f`: full workspace suite 846 passed / 0 failed; clippy clean on touched crates; `rustfmt --check` clean on touched files.

wingleeio (commented):
## Quick follow-up review

The four findings from my previous review are addressed on `c67146f`, and the focused regression suites pass. I found one remaining blocking lifecycle issue:

**Do not discard the only retry entry point when “Not now” is selected.** After a partial import, the flow correctly becomes `ImportFailed`, but the `import-failed-dismiss` handler changes it to `SyncFlow::Idle` and clears the error. At that point the replacement runtime is already `WorkspaceScope::Synced`; `account_menu_action(Synced, Idle)` exposes only **Sign out**, and the import offer is only derived while running the local profile. A restart also returns directly to the synced profile. Consequently, “Not now” does not mean retry later — it makes the remaining local rows unreachable from the UI unless the user signs out and repeats the entire switch or invokes the RPC externally.

Please preserve a resumable pending-import state and expose a working account-menu entry to reopen/retry it (similar to the postponed switch flow), or remove the dismissal. Add a reducer/lifecycle test that dismisses an import failure under synced scope and proves the retry remains reachable.

Validation on `c67146f432a3457a6568d63f67fa83e3853f467e`:

- `cargo test -p zeron-engine --test local_import`: 8 passed.
- `cargo test -p zeron-ui --lib shell::tests::`: 29 passed.
- `git diff --check` against the merged `main`: passed.
- The original failure-summary, marker, spaces-only, and touched-file formatting fixes are present.
- No new malicious or backdoor-like behavior found in the follow-up patch or merge resolution.

This remains a blocking, non-approving review.

wingleeio:
Fixed in `2a0deea`.

`ImportFailed` is now postponable, mirroring the switch-offer flow: **"Later"** (renamed from "Not now") hides the dialog but keeps the failure pending — it no longer transitions to `Idle` or clears the error. While pending, `account_menu_action` on the **synced** runtime yields the reopen entry ("Finish sync setup") instead of Sign out, and reopening shows the same retry dialog with the preserved error message. Retry stays in-place and idempotent. Once the failure is resolved (retry → done → Continue), the menu returns to the normal synced state.

Reducer test as requested — `dismissed_import_failure_stays_reachable_on_a_synced_runtime`: a dismissed failure survives auth/scope updates under `Synced` scope, `account_menu_action(Synced, ImportFailed { notice_open: false })` still exposes the reopen action, and resolving restores Sign out.

One honest note on scope: the pending state is in-memory, like the postponed switch offer it mirrors. If the app fully restarts with a failure pending, the notice is gone and the remaining local rows become reachable again only through sign-out → sign-in (which re-offers, and the import dedupes). Deriving a persistent nag from `LocalImportStatus` at synced boot would misfire on users who chose "Start fresh" (they also have importable rows, by choice), so persistence would need the marker to record intent — happy to add that in a follow-up if you want restart-durable pending state.

Validation on `2a0deea`: full workspace suite 847 passed / 0 failed; `cargo test -p zeron-ui --lib shell::tests::` 30 passed; rustfmt clean on the touched file.

wingleeio (commented):
## Follow-up review

The in-session retry path is now correct: **Later** preserves `ImportFailed`, the synced account menu reopens it, and the new reducer test covers that lifecycle. One part of the previous blocker remains:

**The retry path still disappears after an application restart.** `SyncFlow` is initialized to `Idle` for every new `Shell`, and a signed-in synced runtime keeps that state. The UI never calls `LocalImportStatus`, nor does this patch persist the pending failure. Therefore the sequence **import partially fails → Later → close/restart the app** returns to a normal synced account menu with only **Sign out**; the remaining local rows again have no discoverable import/retry path.

This is not just a theoretical edge: restart is a natural response to an import or filesystem failure, and the button now explicitly promises a later recovery path. Please either persist pending-import intent and restore it on synced boot, or provide a durable/manual **Import local work** entry that consults `LocalImportStatus`. If restart durability is intentionally out of scope, the UI should not offer **Later** without another discoverable recovery route. Add a bootstrap/restart regression test rather than only a reducer test over one in-memory `Shell`.

Validation on `2a0deea27305f3898db40d090f37007d2e9bc0bd`:

- `cargo test -p zeron-ui --lib shell::tests::`: 30 passed.
- `cargo fmt --all -- --check`: passed.
- `git diff --check` for the follow-up: passed.
- No malicious or backdoor-like behavior found in this change.

This remains a blocking, non-approving review.

PR #80, opened 2026-08-14
feat(ui): make Git history interactive and customizable

jsgrrchg:
## Summary

This PR makes Git History more enjoyable and customizable. 

### Interactive graph

- Highlight the hovered graph path and its related commits.
- Dim unrelated graph lines and commit metadata to make the selected branch easier to follow.
- Add per-branch folding controls directly to branch tips.
- Animate individual branch expansion and collapse with an accordion-style transition.
- Add an instant global toggle between the complete history and branch tips only.
- Keep the graph column width stable while folding to prevent commit titles from shifting horizontally.
- Preserve the user's scroll position when commits are folded or restored.

### Customizable columns

- Add a compact column picker accessible from the History header button.
- Allow Author, Date, and SHA columns to be shown or hidden.
- Support resizing and reordering optional columns.
- Persist column visibility, widths, order, and author display preferences.
- Normalize and clamp persisted settings to recover safely from missing or invalid values.
- Align metadata cells consistently, including the interactive SHA copy target.

### Author avatars

- Show author avatars by default, with the author name available in a tooltip.
- Allow switching the Author column between avatar and name modes from its header.
- Resolve GitHub avatars through the engine for repositories with a GitHub origin.
- Download and cache images locally instead of asking GPUI to render remote URLs.
- Match avatars by commit SHA and normalized author email.
- Limit concurrent downloads, enforce timeouts and size limits, and retry transient failures on refresh.
- Avoid GitHub API requests entirely while the Author column is configured to display names.
- Fall back to the author's initial when no GitHub avatar is available.

### Commit search

- Add fuzzy search across commit subjects and SHA prefixes.
- Filter the loaded history immediately, then search the complete repository history so matching commits outside the current page are included.
- Preserve a coherent graph by compacting non-matching commits into the visible search paths.
- Keep the search result count, pagination, refresh behavior, and scroll anchor consistent with the filtered view.
- Present search as a compact icon that expands into an input on demand.
- Automatically collapse an empty, idle search after 1.5 seconds, unmounting the input to avoid GPUI's persistent focused-caret behavior.
- Keep `Esc` unclaimed by the search control so it remains available for future global agent-turn cancellation.

### History data

- Include deduplicated local and remote branch tips in the first history page.
- Keep branch-tip loading independent from regular history pagination so the compact overview is available immediately.
- Forward avatar resolution through RPC for local and remote workspaces.

## UX notes

- Global “branch tips only” switching is intentionally immediate.
- Individual branch folding retains its animated transition.
- Search expansion and collapse are animated, while typed filtering remains immediate.
- An active search stays open; blur and the clear button reset it, and an empty control returns to its compact trigger after a short idle delay.
- Reduced-motion preferences bypass folding animations.
- Hidden columns retain their configured position and width when shown again.

## Testing

- `cargo test -q -p zeron-ui history::tests --lib`
  - 24 tests passed.
- `cargo test -q -p zeron-engine --test m5_repos_diffs_terminals git_history_search_is_fuzzy_complete_and_accepts_sha_prefixes`
  - passed.
- Added coverage for:
  - branch collapsing and transition row ordering
  - scroll-anchor preservation
  - graph layout and hover detection
  - column resizing and reordering
  - persisted column settings
  - avatar decoding, GitHub remote parsing, and local image caching
  - branch-tip history pagination
  - fuzzy search by subject and SHA prefix across complete history


jsgrrchg:
updating the branch

jsgrrchg:
I almost forgot about the flicker, I solved it the latest commit.

jsgrrchg:
Added a compact, animated Git History search control:

- Search now starts as an icon button and expands on demand.
- It filters commit subjects and SHA prefixes with fuzzy matching.
- Empty, idle searches automatically collapse after 1.5 seconds, unmounting the input to prevent GPUI’s persistent caret/focus issue.
- Search remains open while it contains a query; blur, Escape, and clear reset it gracefully.

jsgrrchg:
Removed the History search `Esc` handler so the shortcut remains available for a potential future agent-turn cancellation flow. Search can still be cleared via blur or the clear button.

jsgrrchg:
Added a lightweight branch divergence indicator to Git History.

It compares the current checkout against the best locally available integration branch and shows `N ahead` and/or `N behind` in the header. The status refreshes with History and `Fetch all`, never polls or performs extra network requests, and automatically hides on narrow panes to keep the toolbar uncluttered.

Now I feel git history reached it's final form

jsgrrchg:
Split Git History into its own right-pane surface.

- `Diffs` now opens the normal changes viewer with its existing scopes.
- `History` opens as a dedicated tab with its own graph, search, and toolbar state.
- History no longer appears in the Diff scope picker, preventing an existing Diff tab from changing its identity.
- Both still reuse the same underlying `Changes` viewer and commit-opening event path, so selecting a commit continues to open its diff in a normal Diff tab.

jsgrrchg:
Improved Git History behavior in narrow panels.

- Added responsive graph sizing to preserve commit title space.
- Stabilized lane resizing to prevent shimmer and resize flicker.
- Added a compact rail mode for dense graphs, keeping every commit visible while rendering colored dots connected linearly.
- Preserved branch colors, hover behavior, and commit interactions.
- Added hysteresis so the graph does not rapidly switch modes near the width threshold.

jsgrrchg:
Smoothed the responsive graph mode transition.

When History enters compact rail mode, the graph lanes now converge into the rail instead of switching abruptly. Expanding follows the reverse motion, while the graph width and commit column move in sync.

The transition is short, respects Reduce Motion, and preserves the existing compact-mode hysteresis and hover behavior.

jsgrrchg:
Fixed compact Branch Tips mode to keep tips visually independent.

When the graph narrows, tips now remain as colored nodes rather than being connected by a synthetic rail, which better reflects that their ancestry is intentionally omitted in this overview.

jsgrrchg:
Removed the horizontal dividers between Git history rows while preserving the row and graph hover effects. The cleaner layout reduces visual noise and makes the history view look much better.

jsgrrchg:
Aligned the ahead/behind labels with the branch text for a cleaner visual baseline.

jsgrrchg:
Fixed the tiny seam artifacts on dimmed graph paths by avoiding alpha accumulation at overlapping row segments.

jsgrrchg:
Replaced the redundant “History” header label with the current branch name, keeping the context useful without repeating the tab title.

wingleeio (changes requested):
## Review

Requesting changes for two blocking issues.

### Blocking findings

1. **History search does not retain keyboard focus.** With the main composer focused, clicking the new History search trigger expands the control, but typing still goes into the chat composer. The History input remains empty, its 1.5-second idle timer collapses it, and no filtering occurs. I reproduced this consistently in the headed app, including after clicking inside the expanded search field. The focus handoff in `GitHistorySearchControl::expand` needs to survive the shell's composer focus restoration; please cover the real window-focus path as well as query state.

2. **Complete-history search can stack-overflow and terminate the engine.** `compact_history_commits` recursively follows every hidden parent in `nearest_visible_parents` without a depth bound or iterative fallback. In a linear 20,000-commit repository where the query matched the newest and oldest commits, one `SearchGitHistory` call closed the RPC connection and the engine aborted with `thread 'tokio-rt-worker' has overflowed its stack`. Large repositories can reach this shape through ordinary use. Please make the contraction iterative (ideally memoized) and add a deep-history regression test.

The graph layout, responsive lane compaction, branch-tip view, avatar/fallback rendering, divergence metadata, column presentation, hover treatment, and commit-to-diff opening all behaved correctly in the headed pass.

### Visual verification

Full History graph with responsive lanes, refs, avatars, divergence, and metadata columns:

Git History V2 full graph

Branch-tips overview with row/path focus:

Git History V2 branch tips

Search focus failure: the History field is expanded but empty while the typed query appears selected in the main composer:

History search focus failure

### Validation

- `cargo test -q -p zeron-ui history::tests --lib` — 33 passed
- `cargo test -q -p zeron-engine --test m5_repos_diffs_terminals git_history_` — 3 passed
- `CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo build -q -p zeron` — passed
- `git diff --check origin/main...origin/pr/80` — passed
- Headed Wayland smoke test against the real repository — full graph, avatars, branch tips, hover focus, and commit-diff opening verified
- Deep-history RPC stress test — reproduced the engine stack overflow above

`cargo fmt --all -- --check` is not clean with the current stable rustfmt on this head; most reported drift is inherited from `main`, with an additional rewrite in the new History picker chain in `shell.rs`.

jsgrrchg:
On it, thank you for the review

jsgrrchg:
Resolved both blockers, 

- Preserved keyboard focus in History search when results update.
- Replaced recursive commit compaction with an iterative, memoized traversal to prevent stack overflows on deep histories.

Added regression coverage, including a 20,000-commit history. Tests and scoped Clippy checks pass.

crates/ui/src/history.rs
```diff
@@ -196,16 +749,398 @@ fn layout_graph(commits: &[GitHistoryCommit], head_sha: Option<&str>) -> GraphLa
...
+fn history_search_matches(commit: &GitHistoryCommit, query: &str) -> bool {
+    let query = query.trim().to_ascii_lowercase();
+    if query.is_empty() {
+        return true;
+    }
+    let sha = commit.sha.to_ascii_lowercase();
+    if sha.starts_with(&query) {
+        return true;
+    }
+    let haystack = format!("{} {}", sha, commit.subject.to_ascii_lowercase());
```
wingleeio: [P2] Use consistent Unicode case normalization in both search matchers

The engine's `fuzzy_score` uses `to_lowercase()`, but this matcher uses `to_ascii_lowercase()` for the query and subject. For a commit titled `RÉPARER la recherche` and query `réparer`, I reproduced `git_history_matches == true` and `history_search_matches == false`. Since `recompute_view` runs this matcher again on RPC search results, the valid result is discarded while the engine's result count still includes it. Use the same Unicode normalization in both paths (ideally a shared matcher), and add coverage ensuring a non-ASCII case-insensitive match returned by the engine remains visible in the UI.

wingleeio (changes requested):
Requesting changes for a reproduced search correctness issue: the engine and UI use different case normalization, so the UI can discard valid results returned by complete-history search.

Validation on 6a02da0: 35 UI history tests, 3 engine history integration tests, and the engine's 20,000-commit compaction regression test passed. A separate reproducer using the unchanged engine and UI matching functions confirms the mismatch described inline.

crates/ui/src/history.rs
jsgrrchg: Fixed in a8ce88a3. The UI now reuses the engine’s Unicode-aware history matcher, preventing valid RPC results from being discarded by inconsistent ASCII-only normalization. Added engine and UI regression coverage for the `RÉPARER`/`réparer` case; all 36 History tests and the relevant engine tests pass.

wingleeio (approved):
Re-audited a8ce88a. The reported Unicode search mismatch is fixed by sharing the matcher between the engine and UI, with a regression test confirming returned results stay visible. No new findings in the update. Validation: all 36 UI history tests and the engine Unicode regression test passed locally; UI and macOS frame-recovery CI checks passed.

PR #83, opened 2026-08-14
changes: comment a diff line, stage it on the next prompt

neo773:
Adds line comments to the changes pane. Hover any diff line, hit `+`, write a note; it stages against the composer and rides the next prompt.

### How it travels

The comment ships as plain text, the same way `attachments.rs` ships image refs — no second data model, nothing to persist or sync. `with_comments` appends a block of `path:line (L|R): body` bullets; the agent reads the citation exactly as a human would.

```
Address the review comments below.

Comments on the diff (each cites the file and line it belongs to; ...):
- crates/ui/src/changes.rs:3292 (R): why is this cloned?
```

`L`/`R` is load-bearing: a comment on a deleted line can only cite the pre-change file, so `path:42` alone lands on the wrong line once a hunk shifts.

### The pill

Once sent, the reader should not see that block. `badges.rs` lifts it back out and draws one pill over the bubble, the same pill the composer showed while the prompt was being written. Hovering it opens a card listing each note — same accent bar and mono location header as the staged card in the diff pane, so it reads as the same object in both places.

`badges.rs` knows nothing about diff comments. A badge is `{ icon, label, details }`, and a detail is three neutral slots (`location`, optional `tag`, `body`). Adding another kind of pill is one `Extractor` in `EXTRACTORS` plus a stager on the composer side. No click handlers or action hooks yet — those are speculative until there's a second badge to shape them against.

### Notes for review

- `Changes::replace_file_body` now splices only the rows that moved. `ListState::splice` clamps the scroll anchor to the range start when the anchored row falls inside it, so replacing a whole file body jumped the pane to the top every time a comment card opened.
- The hover card is wrapped in `frost::frosted`. That is not only the blur: it also gives the card its own scene layer. Sharing the transcript's layer let the bubble's text paint over the card, since equal draw orders render quads before text.
- `popover::CARD_RADIUS` replaces the `12.0` that was hardcoded in the card, the frost mask, and `frosted_menu`.

Comment cards are not threads — one note that rides the next prompt and then stops existing.

### Testing

437 tests pass; 9 new, covering the prompt round trip (including a body that itself contains `(L): `), multi-line bodies, paths containing a colon, and the card-height arithmetic the fold tween depends on. Exercised by hand in the app.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/83"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789333465&installation_model_id=425430&pr_number=83&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F83&signature=ae1a8286c910bfe857c8ce6d706915eef5936a95a097e145712bf7fdc0fe91d6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/changes.rs
+++ b/crates/ui/src/changes.rs
@@ -1052,7 +1052,14 @@ struct HoverRow {
 }
 
 struct CommentDraft {
+    /// Composer the note will stage onto, captured when the card opened. A
+    /// draft belongs to the checkout it was written over, so it must not
+    /// follow the user onto whatever chat is selected by commit time.
+    key: String,
     path: String,
+    /// The file's pre-rename path, when it moved — carried onto the comment so
+    /// an `Old`-side citation names the file that line lives in.
+    old_path: Option<String>,
     side: CommentSide,
     line: u32,
     input: Entity<ComposerInput>,
@@ -1597,6 +1604,7 @@ impl Changes {
 
     /// Reconcile parsed content with the currently-active diff.
     fn sync(&mut self, cx: &mut Context<Self>) {
+        self.discard_stale_draft(cx);
         // The watch follows the selected chat's host device (idempotent when
         // the target is unchanged); a boot-deferred attempt retries here too.
         self.ensure_watch(cx);
@@ -1906,6 +1914,29 @@ impl Changes {
             .collect()
     }
 
+    /// The parsed diff's pre-rename path for `path`, when the file moved.
+    fn old_path_of(&self, path: &str) -> Option<String> {
+        self.parsed
+            .as_ref()?
+            .files
+            .iter()
+            .find(|file| file.path == path)?
+            .old_path
+            .clone()
+    }
+
+    /// A draft belongs to the checkout it was opened over. Chat navigation
+    /// swaps both the diff under it and the composer it would stage onto, so
+    /// the half-written note is dropped rather than following the user across.
+    fn discard_stale_draft(&mut self, cx: &mut Context<Self>) {
+        let key = self.state.read(cx).composer_key();
+        if self.draft.as_ref().is_some_and(|draft| draft.key != key) {
+            self.draft = None;
+            self.sync_comment_rows(cx);
+            cx.notify();
+        }
+    }
+
     fn draft_anchor(&self) -> Option<(String, CommentSide, u32)> {
         self.draft
             .as_ref()
@@ -2011,8 +2042,12 @@ impl Changes {
             _ => {}
         });
         let handle = input.read(cx).focus_handle(cx);
+        let key = self.state.read(cx).composer_key();
+        let old_path = self.old_path_of(&path);
         self.draft = Some(CommentDraft {
+            key,
             path,
+            old_path,
             side,
             line,
             input,
@@ -2039,9 +2074,12 @@ impl Changes {
             cx.notify();
             return;
         }
-        let comment = DiffComment::new(draft.path, draft.side, draft.line, body);
+        let comment =
+            DiffComment::new(draft.path, draft.side, draft.line, body).renamed_from(draft.old_path);
+        // `draft.key`, not the live one: the note stages onto the composer it
+        // was written against even if the selection moved under it.
+        let key = draft.key;
         self.state.update(cx, |state, cx| {
-            let key = state.composer_key();
             state.add_diff_comment(&key, comment);
             cx.notify();
         });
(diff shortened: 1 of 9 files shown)


crates/ui/src/composer.rs
```diff
@@ -4379,9 +4422,11 @@ impl Composer {
             return;
         }
         let text = self.input.read(cx).text().trim().to_string();
+        let no_content =
+            text.is_empty() && self.staged().is_empty() && self.staged_comments(cx).is_empty();
```
wingleeio: Comment-only content is recognized here, but `button_mode()` still computes `has_text` from typed text and attachments only. During a live run, staging comments without typing text therefore leaves the button in `Stop`; `on_submit` matches that arm first and interrupts the run instead of steering the review notes. Please include staged comments in the mode predicate and cover the live-run/comment-only case.

crates/ui/src/changes.rs
```diff
@@ -1996,7 +2343,66 @@ impl Changes {
                     .as_deref()
                     .map(|highlights| highlights.spans(line))
                     .unwrap_or(&[]);
-                diff_line_row(line, spans, &theme, gutter_width(file_diff))
+                let gutter_px = gutter_width(file_diff);
+                let row = diff_line_row(line, spans, &theme, gutter_px);
+                let Some((side, line_no)) = line_anchor(line) else {
+                    return row;
+                };
+                let path = file_diff.path.clone();
```
wingleeio, line 2389: For a renamed file this is always the post-rename path, even when `side == CommentSide::Old`. A comment on a deleted line is then serialized as `new_name.rs:line (L)`, although that line belongs to `old_path`; the agent cannot reliably resolve the citation. Please choose the side-appropriate path (and keep the card/grouping lookup consistent with it).

crates/ui/src/changes.rs
```diff
@@ -1694,24 +1854,211 @@ impl Changes {
...
+        };
+        let body = draft.input.read(cx).text().trim().to_string();
+        if body.is_empty() {
+            self.sync_comment_rows(cx);
+            cx.notify();
+            return;
+        }
+        let comment = DiffComment::new(draft.path, draft.side, draft.line, body);
+        self.state.update(cx, |state, cx| {
+            let key = state.composer_key();
```
wingleeio: The draft is not bound to the composer key that was active when it opened. If the user opens a draft, switches chats, and then commits, this reads the new chat key and stages a note from the previous checkout on the new conversation. Capture the key in `CommentDraft` or cancel/reconcile drafts on chat navigation.

wingleeio (changes requested):
I exercised the feature end to end and the visual treatment is coherent, but there are three correctness paths that should be fixed before merge.

Validation:
- `cargo test -p zeron-ui` - 437 passed
- `cargo build -p zeron`
- `cargo clippy -p zeron-ui --all-targets --no-deps` - completes with the repository warning baseline (`-D warnings` currently fails in unchanged code)
- Manual render/interaction pass under a headless compositor

Draft state:

Diff comment draft

Staged inline card and composer badge:

Staged diff comment and composer badge

wingleeio (approved):
Re-reviewed `56ac020` after the follow-up fixes. The three prior correctness issues are addressed:

- comment-only staged reviews now count as composer content and steer a live run;
- renamed-file comments preserve the diff grouping path while citing the side-appropriate old/new path;
- drafts capture their composer key and are discarded on chat navigation, so they cannot leak into another conversation.

Validation:
- `git diff --check b53a68e..HEAD`
- `cargo test -p zeron-ui` -- 441 passed
- `cargo build -p zeron`
- `cargo clippy -p zeron-ui --all-targets --no-deps` -- completes with the existing 15-warning baseline

No new findings. The interaction screenshots remain attached to the prior review.

PR #84, opened 2026-08-14
changes: read a diff side by side, not just top to bottom

neo773:
https://github.com/user-attachments/assets/7d87fa7b-9531-4741-90aa-70c708f4dbbd

**Stacked on #83** — review that first. Until it merges this PR's diff also carries its 3 commits; the one to review here is the last, `changes: read a diff side by side, not just top to bottom`. (GitHub can't base a cross-fork PR on a fork branch, so this targets `main` rather than `feat/diff-comments`.)

---

The Changes pane only ever stacked deletions above additions. A wide window has room for both files at once, which is how an edited line is actually read: old on the left, new on the right, aligned.

A toolbar toggle (new `split-columns` icon) switches layouts. It latches when active, and the choice persists in `ui-settings.json`.

## How it works

Split is a **re-flatten of the same parse**, not a second viewer. `split_pairs` walks a hunk into rows: deletions pair index-for-index against the additions that follow them, context sits on both sides, and the longer side's leftovers get one-sided rows. A deletion arriving after additions opens a new block, so `-a +b -c +d` reads as two edits rather than one four-line one.

Everything else carries over unchanged — the virtualized row model, folds, syntax highlights, comment cards. Row heights stay analytic (`DIFF_LINE_HEIGHT` either way), so the fold tween still needs no measurement.

## Virtualization

Two places needed care, both on the per-frame path:

- The fold tween's stand-in row re-renders every frame and builds only the slice its clip can reveal. The unified arm walks lines lazily and breaks early; a naive split arm paired the **whole hunk** first, which dropped frames on any large file. `split_pairs_upto` bounds both the output and the pending runs — a change block yields `max(dels, adds)` rows, so anything past the budget can only land past it.
- Per-row allocations were trimmed to match unified: the path is refcounted rather than cloned per listener, and a context row's two columns share one set of syntax runs instead of building them twice.

## Comment anchoring

The left column is inert. Notes are cited against the post-change file, which is what the agent edits, so a deleted line has nothing to ask for — only the right column takes a `+`. A test pins the invariant this rests on: a deletion can never land in a pair's right slot.

Old-side notes already staged from the unified layout still render their cards, so toggling layouts never looks like it dropped one.

## Testing

`cargo test -p zeron-ui` — 443 pass. `cargo clippy --all-targets` clean. New tests cover the pairing (change blocks, one-sided files, the `\ No newline` meta line), the capped pairing agreeing with the full walk at every prefix, split flattening and its analytic heights, both columns' cards landing once, and the right-column-is-never-a-deletion rule.

Driven in the running app, not just under test.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/84"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789339007&installation_model_id=425430&pr_number=84&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F84&signature=cd089b5bd6309f70ce441f8fc48ec7656cc286950b26fc663bfc518940c9cf75"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


crates/ui/src/changes.rs
```diff
@@ -442,10 +494,132 @@ pub fn truncate_file_lines(file: &mut FileDiff, max_lines: usize) {
...
+                if remaining == 0 {
+                    break;
+                }
+                if adds.len() < remaining {
+                    adds.push(ix as u32);
+                }
+            }
+            // Context sits on both sides; `\ No newline…` is carried the same
+            // way but renders across the row (see [`meta_line_row`]).
+            LineKind::Context | LineKind::Meta => {
```
wingleeio: [P2] Preserve change pairing across no-newline markers

`Meta` cannot be treated as a context boundary here. For a normal modification where both old and new files lack a trailing newline, Git emits `-old`, `\ No newline at end of file`, `+new`, `\ No newline at end of file`. This arm flushes the pending deletion before the addition arrives, so split mode renders the edit as two one-sided rows instead of one aligned old/new row (and carries two full-width marker rows). Please preserve the pending change block across the marker, or associate each marker with its preceding side, and add a regression test for a modified line where both sides lack the final newline.

wingleeio (changes requested):
Re-reviewed this PR against current `main` after the stacked #83 merge. The combined tree merges cleanly.

Validation performed:
- `git diff --check`
- `cargo test -p zeron-ui` (447 passed)
- `cargo build -p zeron`
- `cargo clippy -p zeron-ui --all-targets --no-deps` (passes with the existing 15-warning baseline)
- Manual headless UI pass: opened Changes, switched between unified and split modes, scrolled mixed add/delete hunks, and verified the paired two-column layout.

One split-pairing correctness issue remains in the no-trailing-newline case; details are inline.

Manual UI evidence:

Split diff view with aligned old and new columns

wingleeio (approved):

PR #90, opened 2026-08-15
Sync reorderable pinned sidebar sessions across desktop and iOS

jsgrrchg:
## Summary

- Add a collapsible, manually ordered Pinned section to the desktop sidebar without replacing the automatic sort of regular sessions.
- Synchronize pin membership and order across connected desktops and iOS through a dedicated per-user registry preference.
- Support desktop Pin/Unpin actions, reordering within Pinned, and animated drag transfers between pinned and regular sessions, including a single pin.
- Refine the Pinned and Archived headers and spacing while preserving their disclosure animations.

## Desktop behavior

- Pinned sessions appear above regular sessions in their saved order. Project filters preserve the ordering of pins that are not currently visible.
- Drag a regular session into Pinned to pin it at the indicated position; drag a pinned session into the regular section to unpin it.
- Dragging regular sessions only previews an insertion position. Dropping without entering Pinned returns the card with animation and never persists a manual order for regular sessions; live activity updates continue to determine their ordering.
- The actual full session card moves vertically within the sidebar rather than appearing as a tooltip or floating duplicate. Neighboring rows animate to open a destination slot, and the source vacancy closes smoothly when crossing sections. Canceling reverses the movement.
- Support edge autoscrolling, reduced motion, and cancellation when the active filter, profile, or pin snapshot changes.
- When there are no pins, Pinned is normally hidden but appears as a temporary drag target for the first pin. A collapsed Pinned header also accepts drops and expands after a successful pin.
- Pinned and Archived use muted labels with right-aligned chevrons and no horizontal rule inside their headers. Collapsed headers show item counts. The separate divider between expanded pinned and regular sessions remains and animates away when Pinned collapses.
- Remove section-wide drag highlights and reduce the extra spacing above Archived. Keyboard session navigation follows the visible sidebar projection.

## iOS behavior

- Home and Space session lists project the same synchronized pin order and show a pin indicator.
- A leading swipe action toggles Pin/Unpin. Manual reordering remains a desktop interaction; iOS reflects the saved order.
- Regular sessions retain their existing automatic ordering.

## Persistence, synchronization, and migration

- Store the complete ordered pin list in the registry's `preferences/sidebar-v1` row, scoped to the user and organization, rather than modifying chat records. Add the corresponding preferences watch and mutation RPC.
- Limit writes to 200 unique, non-empty session IDs.
- Distinguish a missing preference row from an explicitly empty list so removing all pins cannot trigger an unintended re-import from another device.
- Migrate existing desktop-local pins only after authoritative preferences and chat state are available. An existing remote preference wins; local-only profiles retain local persistence.
- Use optimistic mutations and the registry's existing offline/last-writer-wins behavior. A cached preference row permits offline edits; initial edits wait for authoritative state when no row is cached.
- Preserve pins for archived sessions so restoring a session restores its pinned placement. Prune hard-deleted sessions only after authoritative chat synchronization.

## Picture & Video

image

https://github.com/user-attachments/assets/1982db2d-45b5-4d49-afb7-31c7e22c713d

## Test coverage

- Registry preference validation, convergence, and explicit-empty state; engine preference watch and mutation wire shape.
- Pin projection, filtered reorder, deletion cleanup, profile isolation, and live activity ordering.
- Desktop disclosure layout, drag transfers in both directions, single/empty/collapsed-section cases, cancellation, source/destination gap animation, and pin snapshot invalidation.
- iOS pinned-first ordering and regular-session recency behavior.

## Validation after the upstream merge

- `cargo test --locked -p zeron-ui pinned_session_tests --lib` — 17 passed.
- `git diff --check upstream/main...HEAD` — passed.
- iOS tests and remote CI were not rerun or awaited during this branch refresh.

## Branch refresh

Merged the latest fetched `upstream/main` into this branch without conflicts. The PR retains the line divider between Pinned and regular sessions; the experimental whitespace-only replacements are not included.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/90"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


crates/ui/src/shell.rs
```diff
@@ -1268,6 +1315,26 @@ impl Shell {
             self.settings.space_filter = None;
             self.schedule_save(cx);
         }
+        // Pins are device-local presentation state. Keep archived ids so an
+        // unarchived session returns to its pinned slot; prune only after the
+        // synchronized chat list confirms a hard deletion.
+        if state.read(cx).chats_synced && !self.settings.sidebar_pinned_session_ids.is_empty() {
```
wingleeio: [P2] Preserve pins across workspace profile switches

ui-settings.json is device-global, but state.chats contains only the currently active local/synced/development profile. After a user pins local sessions and switches to a synced profile (or another account), the first chat frame for that profile sets chats_synced = true, and this block treats every absent local chat as hard-deleted and permanently removes its pin. Switching back cannot restore the preference even though those sessions still exist in the isolated local store. Please either key pins and cleanup by profile identity or avoid pruning until deletion can be distinguished from ordinary profile absence.

wingleeio (changes requested):
Audited and executed safely on head 1db0058.

Validation:
- cargo fmt --all -- --check
- git diff --check origin/main...HEAD
- cargo test -p zeron-ui — 433 passed, 0 failed
- isolated visual smoke: pinned section and Pin/Unpin menu states rendered correctly; drag reorder completed and persisted the reordered IDs to ui-settings.json

Requesting changes for the profile-boundary preference-loss case noted inline.

wingleeio:
### Screenshot evidence for [review 4943659297](https://github.com/zeronsh/comet/pull/90#pullrequestreview-4943659297)

Synthetic isolated local-profile data; captured from head `1db0058`.

**Pinned section above the recency-sorted list**

Pinned section

**Pinned-session context menu showing Unpin**

Unpin context menu

**After dragging the first pin below the second**

Persisted reordered pins

The reordered ID sequence was also verified in `ui-settings.json` after the drag.

jsgrrchg:
Good catch, I'll fix this.

crates/ui/src/shell.rs
jsgrrchg: Solved in [3cad1c2](https://github.com/zeronsh/comet/pull/90/commits/3cad1c254aba2b72288b35e1ae5819563c95aa28)

jsgrrchg:
Fixed the profile-boundary issue by storing sidebar pins per active workspace identity. Pins now survive local ↔ synced and account switches, while pruning only affects the active profile after its initial chat snapshot. Added coverage for profile switching and isolated cleanup; all zeron-ui tests pass.

jsgrrchg:
I only found one acceptable edge case, pin changes use the existing 400 ms UiSettings debounce, so closing the window immediately after Pin, Unpin, or reorder could cancel the pending write. This behavior already applies to other UI preferences and seems acceptable for this PR.

jsgrrchg:
@wingleeio added remote sync of pins including ios

jsgrrchg:
@wingleeio  the iOS CI failure appears unrelated to the changes introduced in this branch.

The only failing test is `TranscriptLayoutTests.testToolToggleReversalWhileStreamingKeepsTailAttached`, where a timing-sensitive animation measurement reached 116.54 pt against a 100 pt threshold. The other 123 iOS tests passed, and this branch does not modify the transcript layout, animation, or test code involved.

The same test passed on `main` with the same implementation shortly before this run, so this looks like simulator timing variance or a flaky test. I suggest rerunning the failed job first. If it fails consistently, we should investigate or harden that transcript animation test separately from this PR.

crates/ui/src/shell.rs
```diff
@@ -2174,6 +2295,76 @@ impl Shell {
...
+                        if self
+                            .settings
+                            .sidebar_pinned_session_ids_by_profile
+                            .remove(&profile_key)
+                            .is_some()
+                        {
+                            self.schedule_save(cx);
+                        }
+                        let mut retained = preferences.pinned_session_ids;
+                        if spaces::retain_known_pins(&mut retained, &known_chat_ids) {
```
wingleeio: [P2] Prune pins only against a matching authoritative chat snapshot

`chats_synced` becomes true on *any* `WatchChats` frame, including the initial cached/empty value. Chats and sidebar preferences then arrive through separate RPC streams/tasks, so `preferences.synced == true` does not establish that `state.chats` represents that registry snapshot. If the authoritative preferences frame arrives first, this block treats still-undelivered live chats as deleted and writes the shortened/empty list back to the registry. This also affects migration above. I reproduced it by applying an empty initial chat frame, delivering initialized/synced preferences containing a live remote ID, and invoking `on_state_changed`: the pin is removed and a mutation attempted before the authoritative chat frame arrives. Move pruning to the engine under one registry snapshot, or include/match snapshot generations before destructive cleanup; add a reversed-stream-arrival regression test.

crates/ui/src/shell/spaces.rs
```diff
@@ -372,65 +1318,669 @@ impl Shell {
...
+            self.cancel_sidebar_session_transfer(cx);
+            return;
+        }
+        self.sidebar_session_transfer = None;
+        self.cancel_pinned_session_drag(cx);
+        if matches!(target, SidebarSessionDrop::Pinned(_)) {
+            self.pinned_open = true;
+        }
+        // Only pin preferences change. Regular rows keep their live activity sort.
+        self.replace_sidebar_pins(payload.profile_key.clone(), next, cx);
```
wingleeio: [P2] Apply the remote-readiness guard to drag transfers

The menu path rejects pinning when neither `sidebar_preferences.synced` nor `initialized` is true, but this drop handler calls `replace_sidebar_pins` without that guard. A synced profile can already display cached chats while its preferences have not arrived; dragging one into the temporary Pinned target then writes a replacement list built from the initial empty state. That offline/early write can replace an existing remote pin order once the registry connects. Reproduced with actual GPUI drag events: the menu reports “Pins are still syncing,” while a drag immediately sets `initialized = true` and pins the session. Native screenshots are in the review body. Enforce readiness centrally before optimistic state or RPC mutation, while retaining the intended cached-preference offline-edit behavior.

crates/ui/src/shell/spaces.rs
```diff
@@ -372,65 +1318,669 @@ impl Shell {
...
+        });
+        if !matches_drag || !self.sidebar_session_transfer_is_valid(payload, cx) {
+            self.cancel_sidebar_session_transfer(cx);
+            return;
         }
-        self.close_spaces_menu(cx);
-        self.schedule_save(cx);
+        let saved = self.sidebar_pins_for_profile(&payload.profile_key, cx);
+        let next =
+            sidebar_session_drop_pins(&saved, &payload.visible_ids, &payload.chat_id, target);
```
wingleeio, line 2418: [P2] Enforce the 200-pin limit before applying a drag result

Unlike `set_chat_pinned`, the drop path has no `MAX_SIDEBAR_PINS` check. Dragging a regular session into Pinned with 200 saved pins produces 201 entries. I reproduced that with the production drop helper and confirmed `RegistryDoc::set_sidebar_pinned_sessions` rejects the same result. In a remote profile the UI has already optimistically accepted the list and marks it initialized, but the RPC fails, so the displayed pin is not persisted/synchronized; local profiles also exceed the documented limit. Validate the proposed list before clearing drag state or applying the optimistic change, show the existing limit notice, and cover the full-capacity drag case.

wingleeio (changes requested):
Reviewed `e9e5ba8958b1db25873caf0617c0514340d3b770`. **Requesting changes for three reproducible preference-integrity issues**, detailed inline. No new cross-user/org access bypass, command execution, or filesystem access was found in this change. This is a PR-scoped audit, not a certification of the entire application.

The registry route checks the token's organization and derives the room from the authenticated user; the new preference uses that existing scope. I traced the desktop mutation/watch, registry overlay and validation, profile migration/cleanup, drag handlers, and iOS projection/write paths. The previous local/synced profile-loss issue is addressed by profile-keyed local preferences and runtime-state reset. However, synchronization readiness is not consistently enforced, and pruning consumes independently delivered snapshots.

Validation:

- Existing desktop pin tests: **17 passed** locally.
- Registry and engine preference tests: **5 passed** locally.
- Three audit-only reproductions: **all reproduced** the reported bugs (tests assert the observed incorrect behavior): preferences arriving before authoritative chats removes a live pin; unsynced dragging bypasses the menu guard; dragging constructs a 201-item list rejected by the registry. The first two exercise the actual Shell code with GPUI's test context; the limit reproduction exercises the production drop helper and registry validator.
- `git diff --check origin/main...HEAD`: passed.
- Inspected CI logs: **1,092 desktop UI tests passed**. iOS ran 124 tests with one failure in `TranscriptLayoutTests.testToolToggleReversalWhileStreamingKeepsTailAttached` (116.54 versus a 100-point threshold); both new `PinnedSessionsTests` passed. The failing transcript files are unchanged by this PR, but a flaky-test explanation is not established by a single run. No local iOS simulator run was possible on Linux.
- Inspected the author's screenshot and sampled the attached video. Captured the native Linux UI below using an isolated, offline fixture compiled against this head; all session/account data is synthetic. Normal drag reordering was verified in `ui-settings.json` as `chat-3, chat-1` after starting with `chat-1, chat-3`. No live account or remote registry was modified.

Screenshots are GitHub **user attachments**:

**Pinned section above the regular session list**

pr90-pinned

**Pinned-session context menu exposes Unpin**

pr90-unpin-menu

**After native drag reorder; saved order verified on disk**

pr90-reordered

**Unsynced fixture: the Pin menu correctly refuses with “Pins are still syncing”**

pr90-unsynced-menu-blocked

**Same unsynced fixture: dragging nevertheless creates a pin and reaches the mutation path. “Engine not connected” is expected because this fixture deliberately has no engine; this image demonstrates the UI guard bypass, not a live server write**

pr90-unsynced-drag-bypass

crates/ui/src/shell.rs
jsgrrchg: Pruning now runs in the engine against a single authoritative registry snapshot, not the separate UI streams. Migration uses the same protection and keeps the local copy until persistence succeeds. Added regression coverage for preferences arriving before chats.

crates/ui/src/shell/spaces.rs
jsgrrchg: Yep, dragging was bypassing the menu guard. Both paths now share the same validation before any optimistic update or RPC write. Cached preferences still work offline, and rejected drops animate back to their origin. Fixed and added regressions

crates/ui/src/shell/spaces.rs
jsgrrchg: Drops now validate the full list before accepting the move. A 201st pin shows the existing limit notice and the card returns to its original spot. Reordering and unpinning at capacity still work. Added coverage for both local and synced profiles.

wingleeio (approved):
Re-reviewed and security-audited `68306a172f74c87eef1772a756df1ef33ff45e74`. **No blocking findings in the reviewed changes; approving the code review.** The three issues from my previous review are addressed. Platform CI is still running; this review does not assert that all CI is green.

The implementation now uses one `sidebarPins/{sessionId}` record per pin with separate `pinned` and `orderKey` fields. A move writes only that pin's position, and an unpin writes false without deleting the row. Independent concurrent moves survive; a delayed move cannot restore membership. Same-item concurrent edits still resolve by field clocks. Queued UI operations are intents rebased onto confirmed state, not stale replacement arrays.

Security review covered the RPC mutation, registry write/projection/cleanup, edge validation and routing, desktop pending-operation lifetime, and the corresponding Swift implementation. The registry route still derives rooms from authenticated user identity and checks the organization claim; inbound identity headers are overwritten. Session IDs are validated in the Rust mutation, keys use bounded ASCII encoding, and edge validation restricts pin fields and retains the operation-size budget. No new command execution, arbitrary filesystem access, credential exposure, or cross-user/org access path was found. This is a review of this PR's changes, not an application-wide security certification.

Verified locally:

- Rust sidebar-filtered tests: 10 registry, 2 engine, 3 protocol, and 33 UI tests passed. The complete pin interaction group also passed: 29 tests, including the original native drag regressions, readiness/capacity checks, reversed stream delivery, profile/connection boundaries, rejection recovery, timeout handling and late acknowledgements. These test groups overlap.
- Registry tests verify independent concurrent moves, move/unpin races in both delivery orders, replayed operations, persisted keys, future-clock handling and concurrent capacity overflow.
- Edge typecheck passed; 14 registry-core tests and 2 real-SQLite workerd tests passed. Two additional audit-only validation tests passed for malformed membership, unknown fields, non-hex/trailing-newline keys, oversized keys and forbidden row deletion.
- Independent checks against the exact Rust allocator passed 20,000 insertions into varied gaps, asserting strict bounds and valid keys, plus malformed-input and key-exhaustion checks.
- `git diff --check origin/main...HEAD` passed.
- Native Linux fixture: reordered pins saved as `chat-3, chat-1`; dragging before remote preferences are known now returns the card and displays “Pins are still syncing.” Screenshots below are fresh captures from this head with synthetic local/offline data. No live account or remote registry was modified.

Remaining limits are explicit: 200 is an admission limit, so concurrent offline additions can temporarily exceed it without losing pins; 8192-byte key exhaustion returns an error; old development whole-list pins are deliberately not migrated. I reviewed the Swift implementation but did not run an iOS simulator locally. Current platform CI remains pending. Please update the PR description, which still describes the old whole-list row and migration behavior.

**Native pinned layout**

pr90-keys-pinned

**After drag reorder; saved local order verified**

pr90-keys-reordered

**Previously reported unsynced drag bypass is now rejected**

pr90-keys-unsynced-drop-rejected

PR #113, opened 2026-08-15
feat(ui): make active-agent Escape opt-in

jsgrrchg:
## Summary

Adds an opt-in Escape shortcut that interrupts only the currently selected active agent, while preserving all existing contextual Escape behavior.

The shortcut is disabled by default for both new users and existing users with older settings files.

## What changed

- Adds a device-local `Stop active agent with Escape` preference under Settings → Shortcuts.
- Defaults the preference to `false`.
- Persists the preference in `ui-settings.json`.
- Keeps Escape contextual behavior higher priority.
- When enabled, Escape interrupts only the selected chat if it is `Working` or `AwaitingInput`.
- Captures the selected `chat_id` before dispatching the durable interrupt command.
- Preserves per-chat idempotency and failure handling from the existing Stop button.
- Keeps interrupt state isolated between concurrent chats.
- Places the new preference below the keyboard shortcut list.
- Centers the existing shortcut helper text.
- “Restore defaults” also disables the Escape preference.

## Escape consumer protections

The active-agent interrupt is intentionally the final Escape fallback. Other Escape consumers remain protected through multiple layers:

- Shell-owned menus and overlays are handled during the capture phase before the focused child receives the event.
- Menus and dialogs that already use Escape to close themselves stop propagation, so they never reach the interrupt fallback.
- Blocking overlays consume Escape without triggering any agent action.
- Composer mention and slash completions retain ownership of Escape.
- The question wizard retains its existing Escape behavior.
- Pickers, rename dialogs, Changes ref/scope menus, and other contextual surfaces keep their existing cancellation behavior.
- The integrated terminal continues to receive Escape and sends the ANSI `0x1b` byte when appropriate.
- The fallback only runs on the Chat route, only when the selected chat is `Working` or `AwaitingInput`, and only when the opt-in preference is enabled.
- The selected `chat_id` is captured before dispatch, preventing navigation from redirecting the interrupt to another session.
- Per-chat in-flight tracking prevents duplicate interrupts while allowing independent interrupts for other chats.

This makes the agent interrupt a non-invasive fallback rather than a global Escape handler.

## User behavior

With the preference disabled, Escape behaves exactly as it did before this feature.

With the preference enabled:

- Escape stops the active agent in the selected session.
- Escape does nothing when there is no selected active session.
- Escape does not stop agents in other sessions.
- Escape does not override menus, dialogs, pickers, completion UI, or the terminal.

## Compatibility

- Existing settings files without the new field load with the preference disabled.
- New users also start with the preference disabled.
- No engine protocol, document format, sync behavior, or iOS behavior changes.
- Interrupts continue to use the existing durable `QueueCommand(Interrupt)` flow.

## Validation

- `cargo test -p zeron-ui --lib` — 432 passed
- `cargo test -p zeron-engine interrupt --test e2e` — 3 passed
- `cargo test -p zeron-harness interrupt` — 2 passed
- `git diff --check`
- Manual queue validation confirmed contextual Escape precedence and chat-scoped interruption.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/113"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789411948&installation_model_id=425430&pr_number=113&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F113&signature=2fa4d3994aa3650140e0513978bf535e2e60db522c828d67774e56d9fe6ef25b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (approved):
Reviewed, tested, and merging.

**Audit/review:** the fallback is genuinely opt-in (default false, older and partial settings files load as false, "Restore defaults" clears it), and the ordering is right: shell-owned menus/dialogs resolve in capture phase, every contextual Escape consumer now stops propagation, the terminal already swallows every keystroke it consumes (including Escape → 0x1b), and the wizard/completion popups keep ownership. With the preference off, the only observable change is that Escape no longer bubbles past an open shell menu/dialog, which is what we want. `interrupt_chat` reuses the Stop button's per-chat ledger, and the captured `chat_id` keeps navigation from redirecting the interrupt. The engine test proving interrupts stay chat-scoped is a good addition.

**Testing:** on the branch merged with current main: zeron-ui lib 810 passed, engine e2e 20 passed, 0 failures. I pushed a sync merge of main onto the branch (content unchanged, maintainer edits enabled) so CI would run against the current tree.

**CI:** ui-tests, macos-frame-recovery, ios-tests and linux-browser are all green on the head. Two flakes along the way, neither caused by this PR: `TranscriptLayoutTests` (iOS) failed on the pre-sync head and also failed on main today for a merge with no iOS changes; `linux-browser` failed once on the "outside menu click → page click" step (120ms menu exit vs a 150ms fixture pause) and passed on rerun. I also ran the same OpenGL-path fixture locally on this tree three times: all passed.

PR #115, opened 2026-08-15
Transcript: end-of-turn edit summary card with click-through to a turn-diff tab

SinaKhalili:
## What

After each agent execution settles, the assistant's reply now ends with a change-summary card — **"Edited N files · +X −Y"** with a per-file `+/−` stat list — the recap Codex/Cursor show at the end of a turn. Clicking the card reveals the turn's diff in the right-hand pane: closed → the pane opens onto a turn-scoped diff tab; open → the chat's existing "Latest turn" tab activates, or a fresh one is minted.

Screenshots (mock run, dark + light):

| | |
|---|---|
| card | header `pen · Edited 3 files · +5 −3`, per-file rows, chevron fold |
| interactions | whole card = click target with hover/press feedback and a group-hover "View changes" hint; chevron folds the list (stop-propagation) |

## How

- **`zeron-doc`** — `turn_edits()` aggregates a settled entry's tool parts into per-file counts: `diff_stats` first, pre-strip inline `diff`s next, else the edit-shaped call's own path with counts omitted (no fake `+0 −0`). Same-file edits sum in first-touched order; failed/unresolved calls don't count. Doc-only sourcing means historical turns, remote viewports, and offline all work; shell-side edits (an `Exec` that writes files) remain the changes pane's job.
- **`zeron-ui` transcript** — new `RowKind::TurnSummary` emitted as the last row of a settled entry that edited files (so it carries the timestamp strip). House rules kept: content-hash `version` (late stats re-splice), analytic heights (`summary_body_height`, 12-file cap + counted tail), the tool-group fold tween with its remount guard.
- **`zeron-ui` shell** — a new `TranscriptEvent::OpenTurnDiff` on the existing transcript→shell channel (shared with the subagent spawn-chip events). `show_turn_diff` opens the pane if needed, then reuses the chat's latest-turn tab or mints one via the new `Changes::for_scope` (the `for_commit` pattern); commit-pinned tabs can't false-match (their scope is `Commit`).
- **`zeron-harness` mock** — `ZERON_MOCK_EDITS=1` scripts diffed edits (plus one diff-less edit for the counts-unknown row) so dev-demo puts the card on screen.

## Testing

- 6 new unit tests (`turn_edits` aggregation ×4, row emission/splice, analytic heights); full suites green: 425 `zeron-ui` + 75 `zeron-doc`. fmt/clippy clean.
- Verified live via a seeded mock engine + headed app in both appearances.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/115"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789423348&installation_model_id=425430&pr_number=115&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F115&signature=6b01f4fe352e9e95c91369d7d803ab2bf1e80c08cdd4800637d30952c4733d07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/changes.rs
+++ b/crates/ui/src/changes.rs
@@ -1188,6 +1188,21 @@ impl Changes {
         changes
     }
 
+    /// A pane born onto a non-default scope (the transcript's turn-summary
+    /// card opens a "Latest turn" tab this way). Unlike a commit pin the
+    /// scope menu stays available — it is an ordinary pane with a head start.
+    pub fn for_scope(state: Entity<AppState>, scope: DiffScope, cx: &mut Context<Self>) -> Self {
+        let mut changes = Self::new(state, cx);
+        changes.scope = scope;
+        changes
+    }
+
+    /// The pane's current scope — lets the shell find an existing
+    /// latest-turn tab instead of stacking duplicates.
+    pub fn scope(&self) -> DiffScope {
+        self.scope
+    }
+
     /// The surface-tab title (contextual, user request): the pinned commit's
     /// subject (short sha for subject-less commits), else the scope's label.
     pub fn tab_title(&self) -> gpui::SharedString {
--- a/crates/ui/src/changes.rs
+++ b/crates/ui/src/changes.rs
@@ -28,8 +28,8 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use gpui::{
-    AnyElement, App, Context, Entity, FocusHandle, Focusable as _, ListAlignment, ListState,
-    SharedString, Subscription, Task, Window, div, font, list, prelude::*, px,
+    AnyElement, App, Context, Entity, FocusHandle, Focusable as _, ListAlignment, ListOffset,
+    ListState, SharedString, Subscription, Task, Window, div, font, list, prelude::*, px,
 };
 
 use zeron_proto::{Chat, CheckoutDiff, GitHistoryCommit};
@@ -952,6 +952,22 @@ pub fn body_rows(
     rows
 }
 
+/// The file index a transcript summary-card path refers to. Card paths are
+/// the tool calls' — often absolute — while diff paths are repo-relative,
+/// so fall back to suffix matching (the longest relative path wins when
+/// nested names collide).
+fn reveal_target_ix(files: &[FileDiff], path: &str) -> Option<usize> {
+    if let Some(ix) = files.iter().position(|f| f.path == path) {
+        return Some(ix);
+    }
+    files
+        .iter()
+        .enumerate()
+        .filter(|(_, f)| path.ends_with(&format!("/{}", f.path)))
+        .max_by_key(|(_, f)| f.path.len())
+        .map(|(ix, _)| ix)
+}
+
 /// Flatten all files into rows + each file's row span (header at
 /// `range.start`, body rows after it). `collapsed(ix)` folds a file to just
 /// its header. `comments` is the whole staged set; each file takes its own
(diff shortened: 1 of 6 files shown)


crates/ui/src/shell.rs
```diff
@@ -1702,6 +1703,53 @@ impl Shell {
...
+            Some(surface) => {
+                self.set_right_active(surface, cx);
+                match surface {
+                    RightSurface::Diff(id) => self.diffs.get(&id).cloned(),
+                    _ => None,
+                }
+            }
+            None => {
+                let changes =
+                    cx.new(|cx| Changes::for_scope(self.state.clone(), DiffScope::LatestTurn, cx));
```
wingleeio, line 1743: [P1] Preserve the turn that owns the clicked card. `LatestTurn` is backed by a single in-memory `chat_id -> most recent turn-start snapshot`, while `OpenTurnDiff` carries no entry/turn identity. After turn B completes, clicking turn A's still-visible summary therefore opens B's diff (and a file-row click may simply fail to find A's file). Either only make the newest card interactive or pass/persist a per-turn snapshot identity and fetch that exact turn.

crates/ui/src/changes.rs
```diff
@@ -1188,6 +1209,50 @@ impl Changes {
...
+
+    /// Scroll `path`'s file section into view — the transcript summary
+    /// card's per-file jump. Applies immediately when the diff is already
+    /// parsed, else parks until the next parse lands (a freshly-minted tab
+    /// captures its diff async). Consumed on the first parse either way: a
+    /// turn diff WITHOUT the file (a gitignored edit, say) must not scroll
+    /// some unrelated later diff.
+    pub fn reveal_file(&mut self, path: String, cx: &mut Context<Self>) {
+        self.pending_reveal = Some(path);
+        if self.parsed.is_some() {
```
wingleeio: [P2] Wait for the refreshed capture before consuming this reveal. When the existing `LatestTurn` tab is reused, `set_right_active` starts an async refresh but `parsed` still contains the previous turn, so this branch immediately applies and clears the new path against stale rows. A file that only exists in the just-finished turn will never be revealed when the new parse lands. Gate the immediate apply on the parsed key matching the current scoped capture, or keep the pending target through the in-flight refresh.

crates/ui/src/transcript.rs
```diff
@@ -3809,6 +3863,260 @@ impl Transcript {
...
+                div()
+                    .flex_none()
+                    .font_weight(gpui::FontWeight::MEDIUM)
+                    .text_color(theme.text_muted)
+                    .child(label),
+            )
+            // Totals cover only counted files; an all-blind turn (no diffs
+            // reached the doc) shows the file list alone rather than a
+            // misleading "+0 −0".
+            .when(edits.additions > 0 || edits.deletions > 0, |el| {
```
wingleeio: [P2] Do not label partial stats as the turn total. `turn_edits` deliberately permits `counts: None` and sums only counted files; in the supplied mock, `docs/notes.md` is unknown but the header still states `+5 −3` as if that covered all three files. The per-file omission is honest, but the headline is not. Hide or qualify the aggregate whenever any row has unknown counts (not only when every known count sums to zero).

wingleeio (changes requested):
The card looks polished in both appearances and the focused suites pass, but the click-through currently has two correctness problems and the aggregate can present partial counts as exact totals. These should be fixed before merge.

### Visual verification with user content

The captured turn uses the user prompt: “Add an end-of-turn summary that lists each edited file and opens the matching diff when I click it.”

| Dark | Light |
|---|---|
| Turn summary card in dark mode | Turn summary card in light mode |

PR #120, opened 2026-08-16
harness: discover slash commands per workspace, so project skills reach the popup

tobi404:
## Summary

Project-scoped slash commands never reach the composer's `/` popup. A repo with skills in `<repo>/.claude/skills` shows only the built-in and user-level commands.

The cause is one line. Command discovery asks the agent for a session in the wrong directory (`crates/harness/src/acp/mod.rs`, in `discover_commands`):

```rust
let cwd = std::env::var("HOME").unwrap_or_else(|_| "/".into());
let session = client.request("session/new", json!({ "cwd": cwd, "mcpServers": [] }))
```

ACP agents build the command list from the session `cwd`, so with `cwd = $HOME` the agent never sees the project's skills.

Measured against `claude-agent-acp` 0.66.0, varying only the `session/new` cwd:

| session cwd | commands |
|---|---|
| a project with installed skills | **110** |
| `$HOME` | 76 |

The 34 missing entries are exactly that project's skills.

The deeper shape: **ACP advertises commands per session, and a session is defined by its cwd. Zeron models them per harness**, with no workspace anywhere in the path. Three layers carry that mismatch, and fixing one alone changes nothing: the discovery cwd, the `ListCommands` params (which borrow `ListModelsParams` and have no workspace field), and two caches keyed by harness only.

A fourth fact shapes the design. A live session already holds the correct list, and Zeron drops it twice over: the run path emits `AvailableCommands` only from the **initialize** response, which is not cwd-scoped and which `claude-agent-acp` leaves empty; and the real `available_commands_update` arrives inside the `session/new` handshake, where `request_draining` discards every notification.

## Relationship to #160

#160 landed after this branch was opened. It is complementary, not overlapping. It gave the native `claude` and `codex` drivers a slash-command probe of their own (a `claude` `initialize` control request; a `codex` `skills/list` call), each cached **per harness**. It did not add a workspace dimension anywhere — `ListCommands` still carries only a harness id, and every project on a device still gets one shared list.

This PR adds that missing dimension, and follows it into the `claude` driver — because otherwise the branch would reach every agent except the one the bug was reported against. #136 had already moved `claude` off ACP, so before this commit project skills still never appeared for `claude-code`.

`claude` resolves `<cwd>/.claude/skills` relative to **where the process runs**, and has no `session/new` field to carry a cwd instead. Its probe never set `current_dir`. Measured on CLI 2.1.228 against a project holding two marker skills:

| probe cwd | commands |
|---|---|
| the marker project | **81** |
| a bare directory | 79 |

Through zeron, both read 79. Now they read 81 and 79.

`codex` is deliberately left per-harness. Its `skills/list` already answers with per-cwd **groups** under `data`, which `parse_skill_commands` flattens and dedupes, so scoping it means filtering those groups rather than setting `current_dir` — a different change, and the group's own cwd field could not be confirmed without the real binary. It takes the `cwd` and ignores it, documented in place, and the engine cache keys its answer per workspace regardless, so nothing regresses.

## Changes

- **Live command capture** (`crates/harness/src/acp/mod.rs`): `request_draining` gains an out-parameter and now captures `available_commands_update`, still discarding every other notification (a replayed `session/load` must not re-enter the doc). The run's emit gate becomes "we have a list" instead of "initialize said something". This alone fixes a silent hole: an agent that advertises only after `session/new` was invisible.
- **Workspace-aware probe** (`crates/harness/src/acp/mod.rs`, `crates/harness/src/lib.rs`): `Harness::commands` takes `cwd: Option<&str>`; the ACP per-harness `OnceCell` is deleted. The cwd rides `session/new` only, never `current_dir` — a missing directory there fails the spawn as `NotFound`, which this module reports as `NotInstalled`, a message the user cannot act on. A rejected cwd retries once from `$HOME` so built-ins still list. With a cwd supplied the probe always opens a session, or the workspace would be moot for any agent that answers initialize.
- **Workspace-aware claude probe** (`crates/harness/src/claude/mod.rs`): `discover_commands` takes the cwd and sets `current_dir` — the opposite of the ACP probe, for the opposite reason (ACP carries the cwd in `session/new`; claude has nowhere to put it but the process). An absent directory is dropped rather than passed on: `spawn` would fail it `NotFound`, which this driver maps to `NotInstalled`, telling the user claude is not installed. #160's per-harness `OnceCell` is deleted here too — it would pin whichever workspace probed first.
- **Codex stays per-harness** (`crates/harness/src/codex/mod.rs`): takes the `cwd` and ignores it, keeping #160's `OnceCell`. See the section above for why it needs a different change.
- **Engine cache** (`crates/engine/src/commands.rs`, new): keyed `(harness, cwd)`, 10 minute fresh TTL, 30 second negative TTL, single-flight, LRU 16. Two writers feed it: the probe on a miss, and a running session's own `AvailableCommands`. It lives in the engine because those two writers arrive in two different places.
- **RPC** (`crates/engine/src/rpc.rs`): `ListCommands` gets its own `ListCommandsParams` carrying an optional `cwd`. Absent means `$HOME`, so an older engine on a remote device ignores the unknown field and answers exactly as before. Skew degrades, never fails.
- **Run loop** (`crates/engine/src/sessions.rs`): the `AvailableCommands` arm writes `(harness, run_cwd)`. Note it keys by the request's cwd, not the event's — the opposite of the neighbouring `SessionStarted` rule a few lines up, because the popup looks up by `chat.cwd`.
- **Composer** (`crates/ui/src/composer.rs`): cache keyed `(harness, device, cwd)`. The device matters because every project-less chat on every device shares the path `~`. Rendering is stale-while-revalidate, and a fetch is issued on popup open or workspace change, never per keystroke.
- **Stale comment** (`crates/doc/src/parts.rs`): the note claiming `AvailableCommands` "feeds the engine's per-harness command cache" described a cache that did not exist. Now it does, and it is per workspace.
- **Three unrelated repairs.** Two are their own commits. `crates/harness/src/shell_env.rs`: `falls_back_when_interactive_attempt_hangs` failed on every macOS run, because the fixture's PATH omitted `/usr/bin` and the probe script runs `env`, which lives only at `/usr/bin/env` on macOS (Linux has `/bin/env`, so CI never saw it). `crates/sync/src/chat_client/tests.rs`: a `never_loop` correctness lint was a hard error on `cargo clippy --workspace --all-targets`, blocking the lint gate workspace-wide. The third rides the merge commit: that same shell-env test's 400 ms per-attempt budget starved whenever the machine was busy running the rest of the suite in parallel, so it is now 2 s — still nowhere near the 60 s hang it guards.
- **Docs** (`docs/slash-commands.md`): the design, including the measurements above.

## Tests

- **Engine cache** (`crates/engine/src/commands.rs`): key normalization, stale-as-miss, TTL and negative-TTL expiry, LRU bound with in-flight entries spared, single-flight, and a live write resolving waiters while a late probe is discarded. Plus a multi-threaded variant, the only flavor that could observe the newest-wins window.
- **Harness** (`crates/harness/tests/acp.rs`, `fake-acp.sh`): the handshake update reaching the run stream; `session/new` receiving the requested cwd; and the retry pinned to a `home-retry` marker the retried session alone can produce, so deleting the retry fails the test rather than passing on a bare count.
- **Engine integration** (`crates/engine/tests/command_cache.rs`): a running session fills the cache, verified by then reading it with a probe that always errors.
- **Claude driver** (`crates/harness/tests/claude.rs`, `fake-claude.sh`): the fixture advertises one extra command only when its own `$PWD` is the marker project, so `commands_discovery_runs_in_the_requested_cwd` fails if the child runs anywhere else — pinned by name, not by count. A second test drives a directory that does not exist and asserts the built-ins still list, rather than a "claude is not installed" error.
- **Composer**: `slash_cwd` and `slash_cache_key` as pure functions.
- On top of `main`: `cargo test --workspace` 1009 passed / 0 failed / 13 ignored, `cargo clippy --workspace --all-targets` at zero errors. Every hunk in this PR is `cargo fmt` clean; `cargo fmt --all --check` still reports drift in twelve files, none of it inside a hunk this PR adds. All of it is pre-existing on `main` under rustfmt 1.9.0 and left alone here, because reformatting those files would bury the change.

## Verified end to end

**ACP path, on the original branch.** Against a real `zeron headless` engine on an isolated profile, driving real IPC and the real `claude-agent-acp` adapter, back when `claude` still ran over ACP:

```
cwd = project      -> 111 commands
cwd omitted        ->  76 commands   (the old $HOME behavior)
project-only       ->  34, exactly that project's installed skills
```

A repo without those skills returns the same 76. Cold probe 1593 ms, same key again 1 ms.

**Native claude path, on this head.** Two throwaway projects, one holding two marker skills and one bare, against a `zeron headless` engine on an isolated profile and port. The probe sends the exact `ListCommands` params the composer sends:

```
project-with-skills   81 commands  1958ms   only here: zeron-marker-alpha, zeron-marker-beta
project-bare          79 commands  1988ms
project-with-skills   81 commands     0ms   (cache)
deleted-worktree      79 commands           (falls back, no "not installed" error)
cwd omitted           79 commands           (the $HOME behavior)
```

81 and 79 match what the claude CLI itself reports when run by hand in those two directories, so the engine now carries the CLI's own answer through unchanged. Before this branch, zeron returned 79 for both.

The popup itself was not driven: this environment has neither Accessibility nor Screen Recording permission, so the app could be launched and confirmed connected but not typed into or screenshotted. The layer below it is covered above, with the same RPC params the composer issues, and `slash_cwd` / `slash_cache_key` are unit-tested.

## Cost you should know about

A probe spawns a real agent process, runs that project's SessionStart hooks, and leaves a bare session directory under `~/.claude/projects/<slug>/`. Today's `$HOME` probes already do this — one machine had accumulated 666 such directories. This change moves that cost into real project directories, bounded by the TTL, single-flight, negative caching, and probing only on popup open. A running chat never probes at all. The clean fix belongs upstream in the adapter: a capabilities handshake that needs no `session/new`.

## Not in scope

Models keep the `$HOME` probe and their own `OnceCell`. `codex` stays workspace-blind, for the reason in the #160 section. `ListCommandsParams` and the cache key are shaped so both can join later without another interface change.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789482593&installation_model_id=425430&pr_number=120&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F120&signature=2b33b707c1650b18eb7624d4bf9dfe1fec2abf4716bbeaa7f9b164cf8a3a1cfa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
Superseded by merged PR #469. The replacement implementation scopes command and skill discovery to the selected project, passes the workspace to provider discovery, reads ACP session command updates, keys composer catalogs by device/harness/workspace, and refreshes catalogs whenever completion reopens. Thank you for documenting and validating the original workspace-scoping bug.

PR #122, opened 2026-08-16
iOS: single-flight token refresh, staggered room dials, frame-rate settle

wingleeio:
Follow-up to #107 — these fixes were mistakenly pushed to that PR's branch *after* it merged, so they never landed. Same commit, cherry-picked onto main.

## Problem

On-device testing after the #107 release showed a 5–10s connect spinner on every cold open and a transcript skeleton on every session open, cached or not. Three mobile-side causes:

1. **Token refresh had no single-flight** (`AppConfig.currentToken`). WorkOS refresh tokens are single-use (rotated per use — the desktop grew `auth.rs`'s `refresh_gate` for exactly this), and `HomeView`'s `preloadSessions()` dials every non-archived chat at launch: N+1 concurrent `currentToken` calls raced N+1 refresh POSTs with the same token. One won and rotated it; the rest failed, fell back to the expired access token, dialed sockets the edge rejected, and sat in backoff until the winner's tokens propagated. This stalls on any network, good wifi included.
2. **Dial stampede**: N simultaneous TLS handshakes at launch competed with the registry dial (the sidebar the user is looking at) for a thin uplink.
3. **The skeleton time was the settle loop's 50ms poll** in `TranscriptView` — a cached transcript that converges in a few frames still held the skeleton for multiples of 50ms, on every open.

## Changes

- `AppConfig`: refresh is one shared in-flight `Task` every caller awaits; it clears itself under the lock as its last act, so callers either join a live refresh or start a fresh one — never a second concurrent POST. Failure still falls back to the expired token (server rejects, rooms redial through the gate).
- `SessionStore`/`AppModel`: preload still hydrates every store from disk instantly, but room dials are held and released one per 300ms in attention order; opening a session releases its hold immediately. The foreground/path-recovery kick sweep gets the same stagger (registry instant, chats one per 200ms).
- `TranscriptView.settleToBottom`: poll at frame rate (16ms), same ~2s total budget — cached transcripts reveal in ~50ms.

## Validation

Swift doesn't compile on this box — @wing-anara validates on device. Expected: cached sessions render near-instantly; the spinner reflects one refresh + one dial.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/122"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789491545&installation_model_id=425430&pr_number=122&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F122&signature=1b948d3124135df406ead6d154824ec02c5bd793d765f5d4726d680720cef4cb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/ios/Zeron/App/AppConfig.swift
+++ b/apps/ios/Zeron/App/AppConfig.swift
@@ -20,6 +20,14 @@ final class AppConfig: @unchecked Sendable {
     private let lock = NSLock()
     private var tokens: AuthTokens?
     private var devBearer: String?
+    /// In-flight refresh shared by every caller (single-flight). WorkOS
+    /// refresh tokens are SINGLE-USE (rotated per use, desktop auth.rs
+    /// refresh_gate): without this, a cold launch's N room dials raced N
+    /// concurrent refreshes with the same token — one won and rotated it,
+    /// the rest failed, dialed with the dead access token, got rejected,
+    /// and every socket sat in backoff. That was the 5–10s "connecting"
+    /// stall on every app open past token expiry (~5 min).
+    private var refreshTask: Task<String?, Never>?
 
     init(edgeURL: URL, mode: Mode, userId: String, orgId: String,
          deviceId: String, deviceName: String,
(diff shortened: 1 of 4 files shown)


wing-anara (approved):
Verified: claims check out against the pre-PR code (unguarded per-caller refresh, N-dial preload at HomeView launch, 50ms settle poll), refresh_gate precedent confirmed in desktop auth.rs. holdDial lifecycle has no leak paths; single-flight ordering (tokens updated before the task clears itself) is correct. Compiles clean for iOS simulator and device. One non-blocking follow-up: refreshedToken could re-check stored tokens under the lock before starting a new refresh — a caller holding a pre-refresh snapshot that arrives just after the in-flight task clears can still burn the old refresh token (cost: one failed dial cycle, self-healing).

PR #127, opened 2026-08-16
diff-sync: stop reconcile churn spawning git diff back-to-back forever

danperks:
## What happened

On an idle checkout, the engine spawned `git diff` back-to-back forever — each pass burning ~100% of a core for 1–3s, a new one starting the moment the last exited. Enough to starve the machine (WindowServer at 99%, UI hover lag). Nobody was editing files in the checkout.

## Root cause

Two mechanisms combined into a self-sustaining loop:

1. `reconcile` runs on **every** workspace chat row change — including the `branch` and `checkoutId` writes `sync_entry` itself makes — and resolved each chat's checkout identity by spawning `git rev-parse` twice. Every publish fanned out into a storm of `git` spawns.
2. Under that fd pressure, one transient spawn failure (EMFILE) made the chat ungroupable, so reconcile tore its entry down and re-added it on the next pass. Every re-add kicks a full capture, whose row writes trigger the next reconcile.

Teardown discards the entry's checksum state, so the re-capture always republishes, which always triggers the next reconcile. On a large checkout each pass costs seconds of CPU, so the loop never runs out of work.

Separately, watcher setup ran inline on the runtime: the budget walk reads up to `MAX_WATCH_DIRS` directory entries and FSEvents registration stalls for seconds under contention, starving workspace watches and presence for every checkout whenever entries were rebuilt.

## The fix

Damp reconcile rather than treat the symptom — the `apply_numstat` comment records a previous fix for the same runaway symptom, so treating symptoms here is a known failure mode.

- **Memoize identities per cwd.** Chat-watch reconciles spawn no `git` at all; the repair tick revalidates. A failed fresh resolve keeps the memo unless the directory is actually gone.
- **Orphan grace period.** An entry whose chats vanish is only *marked*; teardown happens after a full `REPAIR_INTERVAL` of continuous absence. A single flapping chat-watch emission can no longer destroy a live entry.
- **Serialize reconcile passes.** Concurrent passes could both observe a checkout as missing and both `add_entry` it, the second insert silently discarding the first's checksum state.
- **Watcher setup moves to the blocking pool** in both diff-sync and spaces, attaching to the entry when ready with a kick to close the capture→attach gap.

Repair tick, watch budget bail-out and checksum-suppressed publishing are unchanged. Nothing is special-cased by path or file count — a pathological checkout made this visible, but any checkout can trigger it.

## Testing

`crates/engine/tests/diff_sync_churn.rs` covers all three dampers. Verified they actually pin the fix by mutating each damper away: with pre-fix behaviour restored, `entry_survives_transient_identity_failure` fails with exactly the incident signature (`updated_at` moves — the entry was torn down and re-captured), and `chat_flap_keeps_entry_and_sustained_absence_removes_it` fails on the entry being removed after a single missed pass. `deleted_checkout_is_evicted_after_grace` passes both ways by design: it guards that the dampers don't make legitimate eviction impossible.

Full `cargo test -p zeron-engine` is green (all suites, including 20/20 in `m5_repos_diffs_terminals`). `cargo fmt --check` clean; clippy adds no new warnings in the touched files.

One test deadline moved: `diff_sync_publishes_and_updates_chat_branch`'s watcher-path wait went 20s → 150s. FSEvents is permitted to drop events, and a dropped one converges only on the 120s repair tick, so the old deadline read normal event loss as failure.

## What this doesn't cover

- A checkout whose capture legitimately costs seconds still costs that on every real change; this removes the *spurious* passes, not the cost of a genuine one.
- Identity memoization means a checkout's identity changing underneath us (a worktree repointed at a different repo) is picked up on the repair tick rather than instantly.
- The blocking-pool move bounds runtime starvation but the budget walk itself is unchanged.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/127"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789508311&installation_model_id=425430&pr_number=127&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F127&signature=69b3bf85f94def0f35a0172e2db0d7fc0c6a4e431e9514322674b8765d3ac2df"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/engine/src/diff_sync.rs
+++ b/crates/engine/src/diff_sync.rs
@@ -19,6 +19,14 @@
 //! slow 2-minute repair tick because native watchers may coalesce or drop events.
 //! Snapshots carry a sha256 checksum; an unchanged checksum publishes nothing.
 //!
+//! Reconcile is deliberately damped, because it runs on *every* workspace chat
+//! row change — including the `branch`/`checkoutId` writes `sync_entry` itself
+//! makes. Checkout identities are memoized per cwd (chat-watch reconciles spawn
+//! no git; the repair tick revalidates), and an entry whose chats vanish is torn
+//! down only after [`REPAIR_INTERVAL`] of continuous absence rather than on the
+//! first pass that misses it. See [`resolve_identity`] for the incident this
+//! guards against.
+//!
 //! Beyond the working-tree watch, the service also answers one-shot
 //! `GetCheckoutDiff` captures for the Changes pane's scopes: *branch changes*
 //! (vs `merge-base(baseRef, HEAD)`, same capture path with the base overridden)
@@ -108,10 +116,20 @@ struct CheckoutEntry {
     chats: Mutex<Vec<Chat>>,
     /// Last published checksum — unchanged snapshots publish nothing.
     checksum: Mutex<Option<String>>,
+    /// Set when a reconcile pass finds no chats for this checkout; cleared the
+    /// moment chats reappear. The entry (watchers, checksum state, published
+    /// diff) is only torn down after `orphan_grace` of *continuous* absence:
+    /// tearing down and re-adding restarts a full capture, which costs seconds
+    /// of CPU on a big checkout, so a single flapping chat-watch emission or
+    /// transient identity failure must never destroy a live entry.
+    orphaned_since: Mutex<Option<std::time::Instant>>,
     /// Kick channel into the entry's debounce/sync task.
     kick_tx: mpsc::UnboundedSender<()>,
-    /// Keeps the recursive fs watchers alive; dropped on entry close.
-    _watchers: Vec<notify::RecommendedWatcher>,
+    /// Keeps the recursive fs watchers alive; dropped on entry close. Filled
+    /// asynchronously — watcher setup (budget walk + FSEvents registration) can
+    /// block for seconds, so [`add_entry`] does it off the runtime and attaches
+    /// the result here once ready.
+    watchers: Mutex<Vec<notify::RecommendedWatcher>>,
 }
 
 /// Working-tree snapshot recorded when a chat's turn dispatches — the diff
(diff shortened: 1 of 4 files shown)


wingleeio (approved):
Reviewed the diff in depth and verified it against current main (v0.2.19) locally — approving.

**Verification done:**
- Merged onto current main: clean; none of the touched files changed on main since this PR's base, so no semantic drift.
- All 3 new churn tests pass on the merge, and I confirmed they genuinely pin the fix by restoring pre-fix behaviour via mutation: `entry_survives_transient_identity_failure` then fails with exactly the incident signature (`updated_at` moves — teardown + re-capture) and `chat_flap_keeps_entry_and_sustained_absence_removes_it` fails on removal after a single missed pass. One note: the memo mutation alone doesn't fail the tests — the orphan grace covers the same scenario, so the two dampers are redundant for these cases. That's fine (defense in depth), just worth knowing the tests pin the combined behaviour rather than each damper individually.
- Full `cargo test -p zeron-engine` green on the merge (all suites, including 20/20 in `m5_repos_diffs_terminals`).
- `cargo fmt --check` clean on the touched files (the repo has unrelated pre-existing fmt drift elsewhere).

**Code notes (all non-blocking):**
- The reconcile gate, orphan grace, memoized identities, and blocking-pool watcher setup are each sound, and the capture→attach gap kick + checksum suppression close the races correctly.
- The `identities` memo is never pruned for cwds that stop appearing in any chat (only a failed fresh resolve on a vanished dir removes an entry). It's a few strings per distinct cwd ever used, so harmless, but a repair-tick sweep of unreferenced keys would tidy it if it ever matters.
- During the grace window an orphaned entry's watchers stay live, so an edit in a chat-less checkout can still run one capture; row writes for deleted chats no-op. Bounded at ~2× REPAIR_INTERVAL, acceptable.

Nice writeup — the root-cause analysis matched what the code actually does at every point I checked.

PR #133, opened 2026-08-17
ios: move "+" and the space filter into the bottom bar

tobi404:
Home and the space detail both put "new session" at the top-trailing corner,
the far end of a one-handed reach. Both move to the bottom bar's trailing
side, behind a flexible spacer (bottom-bar items lay out leading-first). On
Home the space dropdown follows it, so both ends of the screen's reach sit in
the same bar: filter on the leading side, "+" on the trailing. The account
menu keeps the top-trailing slot.

Both lists now scroll under the bar's glass, so the bottom edge gets the same
soft scroll-edge effect the top already had.

The connect spinner does not follow the dropdown down. It rode beside it only
to keep out of the list, and on the bottom bar it read as part of the filter,
so it takes the nav bar's top-leading status slot instead: bare, since the
item's shared glass is hidden, and level with the account button.

"+" is violet now, the one tinted control on the screen, the hue markdown
already gives inline code. Only `.buttonStyle(.borderedProminent)` gets the
tint onto the fill - `.glassEffect(.regular.tint())` and
`.buttonStyle(.glassProminent)` colour the glyph and leave the capsule
neutral, and a plain `Button` inside a `ToolbarItem` ignores the glass styles
outright.

`borderedProminent` paints a white glyph and does not darken the tint to suit
it, so the fill is violet-500, not the violet-400 the inline-code wash uses.
Measured off the simulator, violet-400 gave the glyph 2.5:1 - under the 3:1 a
control needs - and violet-500 gives 4.4:1. `Theme.violet` stays the paint
hue and `violetStrong` is the fill, matching the `accent`/`accentStrong`
split the file already keeps.

Labels are `Label`, not `Image` plus `.accessibilityLabel`: the toolbar
idiom, and it carries a title if the item ever folds into an overflow menu.

### Known rough edge

Under the All filter "+" is a `Menu`, and SwiftUI lays a menu-as-button out
4pt wider than the same `Button`, so that one state draws a 52x48 pill where
the others draw a 48pt circle. A toolbar item bridges to UIKit and drops
shape modifiers on the way, so `buttonBorderShape(.circle)`, `clipShape`,
`menuIndicator(.hidden)` and Label-vs-Image all no-op there. Only dropping
the `Menu` squares it up, and the system menu is worth more than the 4pt.

### Testing

Built and run on the iPhone 17 Pro simulator (iOS 26) in demo mode. Both
screens checked: the bottom bar lays out as described, the space filter and
"+" both work, and the "+" fill/glyph contrast was sampled from the rendered
pixels. No new build warnings.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/133"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789550151&installation_model_id=425430&pr_number=133&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F133&signature=064453c6401c465cdce4ecea6eedc6aa44eccf3f7797e592e79a5028a7531706"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/ios/Zeron/Theme/Theme.swift
+++ b/apps/ios/Zeron/Theme/Theme.swift
@@ -33,6 +33,8 @@ enum Theme {
     // ---- paint: accents ----
     static let accent = oklch(0.673, 0.182, 276.935)        // indigo-400
     static let accentStrong = oklch(0.585, 0.233, 277.117)  // indigo-500
+    /// violet-400 — the create/compose hue (the "+" glass, inline code's wash).
+    static let violet = oklch(0.702, 0.183, 293.541)
     static let danger = oklch(0.704, 0.191, 22.216)         // red-400
     static let dangerSoft = oklch(0.808, 0.114, 19.571)     // red-300
     static let warning = oklch(0.828, 0.189, 84.429)        // amber-400
@@ -45,7 +47,7 @@ enum Theme {
 
     // ---- paint: markdown inline code (violet family) ----
     static let inlineCodeText = oklch(0.811, 0.111, 293.571)  // violet-300
-    static let inlineCodeWash = oklch(0.702, 0.183, 293.541).opacity(0.12) // violet-400 @ 0.12
+    static let inlineCodeWash = violet.opacity(0.12)          // violet-400 @ 0.12
 
     // ---- paint: syntax tokens (soft, paint-only) ----
     static let tokenKeyword = oklch(0.709, 0.129, 20.0)   // soft rose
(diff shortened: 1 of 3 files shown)


tobi404:
Screenshot 2026-08-17 at 1 21 50 PM

wing-anara:
closing because we are re-writing the mobile app, thank u for contributing 🙏

PR #134, opened 2026-08-17
registry: gate default agent enablement on the installed probe

### LINKED ISSUE #128: Claude Code is enabled by default and selectable in the main UI even when the CLI is not installed
## Bug

On a machine where the Claude Code CLI is **not installed**, Claude Code still shows up as **enabled by default** in Settings → Agents and remains **selectable in the main harness picker**.

## Environment

- OS: macOS
- Fresh zeron install; no `claude` binary anywhere in PATH, `~/.claude/local/claude`, or the npm global bin dirs (`which claude` → nothing)

## Steps to reproduce

1. Install zeron on a machine without the Claude Code CLI
2. Open Settings → Agents: **Claude Code is toggled ON**
3. Open the main harness picker: **Claude Code is listed as a selectable agent**
4. Picking it only fails later at run time with a NotInstalled error

## Expected behavior

The default-enabled set should respect the installed probe: when the `claude` CLI is absent, Claude Code should default to **disabled** (like every other opt-in harness), or at least be shown as "not installed" and unselectable in the main picker.

Notably, `set_enabled()` already refuses to *manually* enable a harness whose CLI is missing ("CLI is not installed on this device") — but the out-of-the-box default set bypasses that gate entirely.

## Root-cause pointer

In `crates/engine/src/registry.rs`:

- `default_enabled()` unconditionally returns `[ClaudeCode, Codex]`
- `descriptor_enabled()` falls back to `default_enabled()` membership whenever the user hasn't touched the setting, without consulting the `installed` probe — even though `descriptors()` refreshes that probe on every call

So on a machine without `claude`, the descriptor reports `installed: false` yet reads as enabled from the default set. (Codex is in the same default set, so it presumably has the same behavior when its CLI is missing.)

danperks:
## Summary

On a machine without the Claude Code (or Codex) CLI, the agent still showed up **enabled by default** in Settings → Agents and stayed **selectable in the main harness picker** — picking it only failed later at run time with `NotInstalled`. The out-of-the-box default set bypassed the `installed` probe that `set_enabled()` already enforces for manual toggles.

This makes enablement follow **detection**:

- `enabled_set()` now offers every registered harness whose CLI probe passes, minus the user's explicit opt-outs — instead of returning a fixed `[ClaudeCode, Codex]` set.
- `descriptor_enabled()` falls back to `installed && !mock` when a catalog predates the `enabled` field, so a descriptor reporting `installed: false` never reads as enabled.
- `harness-prefs.json` switches from an allow-list (`enabled`) to an opt-out list (`disabled`). A legacy allow-list is migrated once on load: anything registered and absent from it was a deliberate "no", so it stays off — while an agent installed later turns itself on with no trip to Settings.
- The mock harness is excluded from auto-enable (it "resolves" everywhere) so the last real agent can't be switched off, leaving the composer with nothing runnable.

Net effect: no `claude` binary in PATH → Claude Code defaults **off** and is unselectable, exactly like every other opt-in harness.

## Test plan

- [x] `cargo test -p zeron-engine --lib registry::` (8 passing)
- Added coverage:
  - `descriptor_without_new_fields_parses_with_fallbacks` — unknown enablement follows detection (found → on, missing → off)
  - `enablement_stamps_guards_and_persists` — detection-driven flags + gate + reload
  - `newly_found_harnesses_enable_themselves_without_reviving_opt_outs`
  - `detection_never_enables_the_mock`
  - `legacy_allow_list_migrates_to_opt_outs`

Resolves #128

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/134"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789552379&installation_model_id=425430&pr_number=134&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F134&signature=c57b85a8a90bd91e939c93ada2bd66a2cf5b9d13dc7cf64877bd06557b2bc575"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (changes requested):
Thanks — this is a clean take on #128, and the opt-out prefs shape (with the one-time legacy migration) is the right call for letting later installs surface on their own. I ran the branch through the unit suites and a live headed pass on Linux; the core fix works exactly as advertised. One CI-breaking miss and one residual gap below.

## Verified working

- `cargo test -p zeron-engine --lib` — all 58 pass, including the 8 registry tests. Clippy clean for the touched code.
- Toggling in Settings persists the new shape (`{"disabled": ["codex"]}`) and the composer rail follows it live.
- **Machine with claude/codex present** (plus cursor-agent and grok): all four detected agents default ON, Hermes/Pi OFF with install hints — the new auto-enable working:

  

- **Machine without claude/codex** (hidden via a mount namespace; cursor-agent + grok still present) — the exact #128 scenario. Both default OFF, greyed, with "Install the claude/codex CLI to enable", and the picker only offers the detected agents; the composer default heals to Cursor:

  
  

## Blocking: a zeron-ui consumer test wasn't updated

`cargo test -p zeron-ui --lib pickers::` fails on this branch:

```
pickers::tests::offered_harnesses_follow_the_catalog_enabled_flags
  left:  [ClaudeCode, Codex, Grok]
  right: [ClaudeCode, Codex]
```

`crates/ui/src/pickers.rs:3995` still asserts the old fallback (a catalog with `enabled: None` → default-set membership, Claude + Codex only). Under the new `descriptor_enabled`, the all-`None` catalog with `installed: true` now offers Grok too — which is the intended new semantics, so the test (and its comment) needs updating to match, not the code. Worth running the ui suite before pushing since that file is the main consumer of `descriptor_enabled`.

## Gap worth a decision: zero CLIs installed

`offered_harnesses` has an empty-rail fallback (`crates/ui/src/pickers.rs:3349`): when nothing is enabled it offers everything visible. With this PR, a machine with **no** agent CLIs at all produces exactly that state — so the composer still defaults to Claude Code and every agent stays selectable, failing at run time with `NotInstalled`:



(Live run with every CLI hidden — chip says Opus 5 / Claude Code.) The Settings half is fully fixed in this state (everything OFF with install hints), but the picker half of #128 survives on the fresh-machine-with-nothing-installed shape, which is plausibly the reporter's own setup. That fallback is pre-existing UI code, not this PR's — so I'm fine landing this as-is with the test fix, but the fallback probably wants a follow-up (an "install an agent CLI" empty state in the composer rather than resurfacing unrunnable agents). Flagging so it doesn't get lost when #128 auto-closes.

## Non-blocking notes

- The migration marks every registered-but-unlisted harness as a deliberate opt-out, including ones that were merely uninstalled when the legacy file was written — those users won't get auto-enable when they later install that CLI. One-time and conservative; the test comment acknowledges the tradeoff, just confirming it's intended.
- `set_enabled(Mock, true)` now returns `Ok(())` while silently doing nothing (the `auto_enabled` filter wins). Dev-rig-only surface; a comment or an explicit error would save the next person a confused half hour.

danperks:
Addressed in `aacb621`:

- Updated the stale picker expectations.
- Removed the empty-rail fallback that resurfaced uninstalled agents.
- Added a clear “No agents available” state and blocked new-chat sends without a runnable agent.
- Made `set_enabled(Mock, true)` return an explicit error.

Verified with all 441 UI tests and 8 registry tests passing.

wingleeio (approved):
Re-reviewed `aacb6210c` — all three findings from the last round are addressed, and going after the empty-rail fallback itself rather than papering over it was the right call. Approving.

## Verified

- `cargo test -p zeron-engine --lib` (58) and `cargo test -p zeron-ui --lib` (441) both green, including the updated `offered_harnesses_follow_the_catalog_enabled_flags` and the new all-disabled / all-uninstalled empty cases. Clippy has nothing new in the touched code.
- Live headed pass with **every agent CLI hidden** (mount namespace): the composer now boots to a muted terminal-icon "No agents available" chip with no effort chip, the picker shows the guidance note with no rail, and send stays inert — typed a message, clicked send, no session minted. This closes the zero-CLI gap from the last round for real:

  
  

- Normal machine (CLIs present): chip resolves to Claude Code / Opus 5 as before, send active — the new `send_blocked` clause can't trip when anything is offered, and `effective_harness` already drops a remembered pick the catalog no longer offers, so the gate composes correctly with stale sticky defaults.
- `set_enabled(Mock, true)` now refuses loudly. 

## Non-blocking nit

The new empty-list note keys off `descriptors.is_empty()`, but `rail_descriptors` is also empty while the catalog is still **loading** — open the picker before `ListHarnesses` returns and you get a flash of "Enable an installed agent in Settings…" where the loading state used to be. Invisible locally; could show for real seconds on a retargeted remote device with a slow link. Gating the note on `self.harnesses.ready().is_some()` (or folding it into the `effective_models` match) would kill it. Fine as a follow-up.

One rig note for anyone re-testing the zero-CLI state: the grok CLI self-repairs its install into `~/.grok/bin` when launched, so a reused "empty" bind-mount quietly grows a working grok back — burned twenty minutes on a phantom "Grok 4.6" before spotting it. Fresh mount targets per run.

wingleeio (approved):
Re-approved at `dff80b1` — a merge with `main` that also rescopes the PR now that #145 landed:

- The UI half (no-fallback offered set, no-agents chip/card, send gating) shipped on `main` via #145 in equivalent form, so the merge takes `main`'s versions wholesale — notably `main`'s `send_blocked` only concludes once the catalog is loaded, so offline sends stay possible.
- What this PR now contributes is the registry's detection-based enablement: `enabled_set` = installed minus opt-outs, the `disabled` prefs shape with the one-time allow-list migration, mock excluded from auto-enable, and the picker catalog tests updated to the new `descriptor_enabled` fallback. #145's fresh-machine test is rewritten for the new model: uninstalled defaults are never enabled at all, and toggling them off is a clean no-op; the last-enabled guard goes back to unconditional since everything enabled is now runnable by construction.
- The enabled-but-uninstalled settings hint and picker filter stay as cross-version defenses — catalogs from older engines can still stamp that combination.

Startup order is safe for the migration: every call site builds `default_registry()` (full registration) before `assemble_*` calls `load_prefs`, so the fold sees the complete registered set. The earlier loading-flash nit is moot — `main`'s empty state only renders on a Ready catalog.

Verified on the combined tree: workspace check clean, zeron-engine lib 124/124 (registry suite includes detection, migration, mock-exclusion, and mid-session-install coverage running real probes), zeron-ui 480/480.

PR #136, opened 2026-08-17
Native harness drivers (Claude/Codex/Cursor), Grok ACP hardening, per-subagent transcript visualization

wingleeio:
Native drivers for Claude/Codex/Cursor, hardened Grok ACP, and per-subagent transcript visualization. Commits are phased in review order (A1→A5, B, then fixes); each driver was validated against the live agent CLI on this machine before landing.

## Policy

ACP is kept ONLY for agents built ground-up on it — **Grok** (`grok agent stdio`) and **Hermes** (`hermes acp`) — plus pi via `pi-acp` until a native driver exists. Claude, Codex, and Cursor move to native drivers: the adapters held prompt turns open for background work the CLIs settle eagerly, manufacturing the done-status bug class (phantom Working, quiesce patchwork) by construction.

## A1 — Claude over stream-json (validated on 2.1.228)
- **Permissions**: the stdio control channel — `--permission-prompt-tool stdio` + `can_use_tool` control requests, re-validated live (request/response shapes captured under `tests/fixtures/claude/`). Tools auto-allow (unattended parity); `AskUserQuestion` round-trips through the input bridge. The MCP-tool channel was the rejected alternative (needs a server process).
- **Eager done**: the CLI's own `result` frame ends the turn; a background subagent never holds it. The CLI's native wake turn (second `init`, same session id, second `result` — live-verified) flows through the same stream and the engine's existing parked-resume path renders done→Working→done with zero engine changes.
- **Subagents**: frames tagged `parent_tool_use_id` wrap as the new `AgentEvent::Subagent` — never folded into the parent feed. A verbatim live capture replays as a fixture test.

## A2 — Codex over app-server JSON-RPC (pinned 0.146.1, experimental API)
- Done = `turn/completed` (plus `turn.status == "failed"` and the nested `params.error.message` shape, both live-verified). Approvals via the requestApproval RPCs; the experimental `item/tool/requestUserInput` reaches the input bridge.
- **Child threads route through an explicit table** (pure, tested): item lifecycles AND message/reasoning deltas become tagged events — live capture shows child threads DO stream deltas on this wire, so codex subagent transcripts are token-level live with no rollout tailing. Child turn/thread bookkeeping is consumed (it must never settle the parent turn) and **unknown methods fall through to the parent — fail open, never silent loss** (the reference implementation shipped two bugs from a catch-all swallow). The root's own activity marker never registers as a child (a third shipped bug elsewhere). Two-child fan-out capture archived as a fixture.

## A3 — Cursor via the pinned `@cursor/sdk` shim (NOT ACP)
- First step per plan was determining what wire the SDK wraps: **it doesn't wrap `cursor-agent` at all** — it's Cursor's agent runtime bundled in-process, speaking proprietary protobuf/ConnectRPC to their backend with client-side tool execution. No stdio wire to drive from Rust, so the driver is a thin zeron-owned Node shim (JSONL over stdio) around the pinned `@cursor/sdk@1.0.28`, materialized into a managed npm install.
- Done = the SDK run's terminal result. Subagents: `tool-call-delta.taskUpdate` streams the FULL nested transcript tagged by the spawning task's callId. The permission-free print surface is never used.
- Shim validated live against the real SDK (ready + clean actionable auth error end-to-end).

## A4 — retirement
claude-agent-acp/codex-acp specs, prewarm entries, the `_session/turn_ended` upstream dependency (adapter PR #997), the Claude cost-frame settle machinery, and Cursor's ACP extension surface are gone. Engine: new `Harness::deterministic_turn_end` — native drivers return true and `drive_run` stops arming the 120s/20s quiesce watchdogs for them (`ZERON_TURN_QUIESCE_MS` still forces one for diagnostics); ACP agents keep the backstop.

## A5 — Grok hardening (stays on ACP)
- Spawn: `grok --no-auto-update agent --no-leader stdio` — flag *placement* verified against 1.0.4 (top-level vs subcommand; not the spellings the field notes guessed). **Field-hang repro**: leader on a private socket + SIGSTOP → attach gets no initialize response for 10s then dies (1.0.4's client timeout; older builds hung forever — the reported signature); same wedged machine with `--no-leader`: ~1s clean turn.
- `_x.ai/session/prompt_complete` is the authoritative turn end (live-confirmed, echoes a zeron-minted `_meta.promptId`): it settles the turn through the same response arm via a synthesized response; the possibly-hung RPC drains discarded. Stale/foreign/out-of-order completions can never settle a newer turn. Prompt response stays as fallback — first wins, exactly once.
- Prompt-stall watchdog (30s of TOTAL wire silence after a prompt; `ZERON_ACP_PROMPT_STALL_MS`) surfaces a visible error chip instead of indefinite Working. Grok's `_meta` usage tokens are now captured.

## B — subagent visualization
- **Main doc stays light**: subagent content never enters the parent chat doc. The spawn chip is the index — `subagent_ref` (the subagent doc id), `subagent_status` (running/done/failed — distinct from `resolved`, which eager-done fires early), and a one-line live tail.
- **One doc + one ChatRoom per subagent** (`{chatId}--sub--{toolUseId}`, hashed to fit the edge `ID_RE`), folded/flushed on the same coalesced tick as the parent, **frozen to a static R2 blob at completion** (`blob/{chatId}/{subDocId}`) so viewers of finished subagents never wake the room; the sink drop unpins the doc and the room idles. Every failure path degrades to chip-only.
- **Right-pane `Subagent` tab** beside Diff/Terminal, opened from the chip: strictly read-only Transcript instances (no composer/steering/permission UI). Running → doc watch over the unchanged `WatchDocMessages` RPC; finished → frozen blob first, doc watch fallback. Tab close drops the watch and unpins the LRU.
- Feeds: Claude token-level (tagged wire), Codex token-level (child deltas on the wire), Cursor token-level (SDK nested deltas).

## Validation
- Fixture tests from captured real frames (claude background-subagent replay; codex multi-agent capture archived), fake-CLI scenario suites per driver, `#[ignore]`d live-CLI smokes (claude + codex pass live on this machine), unit tests for chip folding, room-id derivation (ID_RE safety), and the child routing table. Full workspace suite: 50 binaries green.
- Media (rig screenshots + the two videos) to follow in this PR thread.

## Honest gaps
- **Cursor SDK auth is separate from `cursor-agent login`** (verified live): runs need `CURSOR_API_KEY` (or an SDK browser login). Existing cursor users must re-auth once; the error chip carries the exact fix. An authed live run could not be exercised on this machine — the shim is validated to the auth boundary; first authed run may surface mapping gaps (they degrade, not error).
- **Cursor SDK beta churn**: public beta, ~weekly releases, pinned at 1.0.28. Unknown update kinds are ignored by design — a drifted SDK degrades the stream, never errors the chat. Revalidate the shim on every bump.
- **Cursor `askQuestion` is disallowed** at agent creation: the SDK exposes no public answer channel (its `request` message carries only an id), so a question would block the run forever. Known regression vs the ACP `ask_question` bridge until Cursor ships one.
- **Grok subagents have no tab in v1**: grok ≤1.0.4 streams the subagent interior inline UNTAGGED over ACP (docs promise tagging; unimplemented upstream), so it stays in the parent transcript — visible, but duplicated conceptually and not separable. The file-tail harvester (`~/.grok/sessions/…/chat_history.jsonl`) and the small upstream `_meta` tagging ask are the follow-ups.
- **Vendor-format fragility**: codex app-server is experimental (pinned 0.146.1, revalidate on bump); all subagent plumbing fails soft to chip-only.
- **Claude permission channel**: `--permission-prompt-tool stdio` is undocumented (absent from `--help`) but is the same transport the Agent SDK drives; re-validated live this cycle and pinned by fixtures.

## Media

Captured on a headless rig (sway + real engine from this branch, real `claude` 2.1.228 and `codex` 0.146.1 CLIs, offline dev-mode edge).

### Video 1 — Claude background subagent, full flow

Three background-subagent launches in one session: prompt → Working → eager `LAUNCHED` + turn settles to **Done while the subagent still runs** → spawn chip shows **· running** → "Open subagent ▸" opens the read-only right-pane tab → the third launch's tab is opened mid-run and **streams the subagent's transcript live** (command chips accumulating) → the wake turn flips the session **Done → Working → Done** again.

*(Recorded pre-7fd861e: the chip vocabulary it shows — the “· running” suffix and “Open subagent ▸” row — has since been redesigned; the stills below are current.)*

https://github.com/user-attachments/assets/5f3f012a-8537-4d1d-8f58-0e250abdd886

### Video 2 — Codex fan-out (alpha + beta in parallel)

Codex multi-agent fan-out over the app-server wire: both `Agent: alpha` / `Agent: beta` spawn chips fold from `subAgentActivity` items, both children run in parallel (alpha sleeps 15s; both markers written), turn settles with `DONE`. Note: no "Open subagent" affordance appears on codex chips — see finding 2 below (since fixed; codex chips now carry the full designed lifecycle, see stills).

https://github.com/user-attachments/assets/cb5dd2eb-03f6-4bbe-97e1-e11f9f478354

### Stills

Recaptured at **7fd861e** (“subagent UX: real terminal statuses, bare tab titles, designed chip”) — live runs on this rig with real `claude` 2.1.228 and `codex` 0.146.1, `cargo build -p zeron` at 7fd861e. The chip now carries its subagent’s lifecycle **visually**: the app’s mini gradient dot loader pulses in the trailing slot while the subagent runs (no “· running”/“· done” text anywhere), done is the ordinary quiet chip. Subagent tabs are the **bare task description** (no “Agent: ” prefix) and the tab’s strip icon swaps to the mini spinner while the subagent doc streams, settling back when it finishes.

> **Update (06fe8d1):** the chip quieted down further — the one-line live tail is gone end to end (a running chip is the static task description + loader, nothing rewrites mid-stream) and the expanded card’s affordance row now reads **“Open subagent”** (was “Show subagent transcript”). The first table row below is recaptured at 06fe8d1 accordingly; the follow/fade round beneath the table covers the tab’s new scroll behavior.

| | |
|---|---|
| Collapsed spawn chip while RUNNING (recaptured at 06fe8d1) — static task description + mini loader before the chevron; no live tail anywhere in the header | Expanded running chip (recaptured at 06fe8d1) — quiet static header, body shows the **“Open subagent”** row |
| running collapsed | running expanded |
| Quiet DONE chip — no loader, no “· done”; the wake turn acknowledges completion in line | Subagent tab opened MID-RUN — bare task title (“audit the tele…”), strip icon swapped to the mini spinner, transcript streaming from the top |
| done chip | tab streaming |
| Same tab after the subagent finished — strip icon settled back from the spinner, chip quiet, full transcript frozen top-aligned | Codex staggered fan-out MID-TURN — **alpha’s chip flipped to quiet done the moment its child turn completed** while beta still runs with loader + live tail (chips no longer stick at running) |
| tab settled | codex mixed |
| Codex fan-out after the turn — both chips settled to the quiet done chip | Codex `alpha` subagent tab — bare title, the child thread’s own transcript, read-only |
| codex done | codex tab |
| Chip failed state (parent CLI killed mid-subagent → end-of-run drain; pre-refinement rig, kept for the failure flow — at 7fd861e a failed chip renders as danger-tinted label+detail with no “· failed” suffix; not re-captured) | Frozen transcript of the failed subagent (`Ran 2 commands · 1 failed`; pre-refinement rig) |
| failed | frozen failed |

### Follow/fade round (06fe8d1)

Live run at **06fe8d1** on the same rig (real `claude` 2.1.228, one background subagent emitting 18 progress reports with foreground sleeps so its transcript overflows the pane). The recording walks the whole scroll cycle in the right-pane Subagent tab:

1. tab opened early — transcript anchored at the **top** with the empty runway below, **no fade band**, no pill;
2. as it overflows while streaming the view **follows the end** (same stick-to-bottom glide as the main transcript) and the **top-edge fade** appears as content slides under the pane’s top edge — nothing at the bottom;
3. **wheel-up mid-stream releases follow**: the view holds its offset while segments keep landing below, and the main transcript’s “↓ Scroll to bottom” pill fades in over the pane bottom;
4. **clicking the pill** glides back to the end and follow re-engages (the tail keeps tracking through the subagent’s finish).

Throughout, the parent chip header stays the static task description + loader — nothing rewrites per delta.

https://github.com/user-attachments/assets/e99d1b6f-3e72-4599-a514-dcfb1375216d

| | |
|---|---|
| Wheel-up hold mid-stream — follow released, **top fade** over the scrolled-under line, “↓ Scroll to bottom” pill floating over the pane bottom (no bottom fade) | Tab opened early in the run — content top-anchored with the empty runway below, no fade band, no pill |
| hold with fade and pill | early top-anchored |

**Glass pill (1fdf149):** both “↓ Scroll to bottom” pills (main transcript overlay + subagent pane — one shared element) now frost on glass appearances: radius-clipped backdrop blur at the composer tier under the floating-card tint, hairline + shadow kept; hover fades the on-glass wash in **over** the tint (pill deepens, never goes see-through). Opaque appearances keep the raised surface. The pills in the earlier stills above predate this and render as the old opaque chip. Verified on the live rig in both hosts and both appearances — transcript text visibly blurs through the pill in every still:

| | |
|---|---|
| Light · main transcript — “into the Loro doc every 120ms” runs behind the pill, crisp outside, blurred through the glass | Light · subagent pane — same treatment over the pane’s transcript (“n the … ff to next segment” behind the pill) |
| glass pill light main | glass pill light subagent |
| Dark · main transcript — dark tint over the blur (“so the oplog stays RLE-merged” ghosting through) | Dark · subagent pane — identical pill in the pane host |
| glass pill dark main | glass pill dark subagent |
| Hover (light) — the on-glass wash fades in over the tint: the pill deepens without thinning see-through | |
| glass pill hover | |

**Bot icon (9a12547):** the Subagent tab and the spawn chip’s icon tile now carry a hand-drawn bot glyph (antenna + eyes + ears, Solar Linear style — the embedded set has no robot) in place of the chat bubble / generic widget glyph; chat-bubble usages that are actually chat keep theirs. Live claude rig at 9a12547 — one real background subagent; the right pane shows its tab (bot icon) beside a Terminal tab with the frozen transcript below, and the parent transcript shows the settled spawn chip with the matching bot tile:

bot icon — subagent tab beside terminal + spawn chip

Close crop of the tab strip — bot glyph on the subagent tab, terminal glyph beside it:

bot icon — tab strip crop

### What could not be captured, and why (found while filming — not faked)

> **Update (7fd861e):** all three findings are now fixed. Findings 1 (claude half), 2, and 3 were fixed in bed83ba/e713383; the codex half of finding 1 is fixed at 7fd861e — each child’s turn completion now settles its fan-out chip (no `close_agent` needed), verified live in the stills above (alpha flips to done while beta still runs; both quiet after the turn). The chip language itself was also redesigned at 7fd861e: loader while running, ordinary quiet chip when done, danger tint on failure — no status-text suffixes.

1. **A "· done" chip is unreachable on both harnesses in this rig.**
   - *Claude 2.1.228*: when a background subagent finishes, the CLI emits **no tagged terminal frame** — completion arrives only as untagged `system` frames (`task_started`/`task_updated`/`task_notification`/`background_tasks_changed`), which `claude/normalize.rs` drops. Verified with a raw `--output-format stream-json` capture. So the chip stays **· running** even after the wake turn confirms completion (visible in the media); the sink only ever closes via the end-of-run drain, which stamps **failed** (that is also how the failed-state still was produced).
   - *Codex 0.146.1*: child threads end with `turn/completed`, which `route_child_notification` deliberately consumes; the driver's tagged-Done source is `thread/closed`, which only fires via the collab `close_agent` tool — and the model reports that tool is not available here ("I can't close them because no `close_agent` tool is available"), so no tagged Done ever arrives.
2. **Codex chips never gain `subagentRef`/`subagentStatus`/`subagentTail` in the doc** (so no "Open subagent" affordance for fan-outs, and no status/tail suffix in video 2), even though the per-subagent docs are created and populated (verified by watching `{chatId}--sub--{callId}` over WatchDocMessages). Root cause: `update_part_fields` in `crates/doc/src/schema.rs` doesn't write the three new chip fields, so chip mutations made inside the **live** streaming segment are dropped by `SegmentWriter::sync`. The fields only persist via the parked in-place `update_subagent_chip` path — which is why the claude flow (tagged traffic arrives after the eager Done parks the session) shows ref + **· running** correctly.
3. **The chip's one-line live tail never populates for claude**: 2.1.228 emits **zero** tagged `stream_event` content deltas for subagent output (verified on the raw wire); subagent text arrives only as tagged full `assistant` frames, whose *text* blocks the normalizer drops (it assumes they already streamed as deltas). Subagent transcripts are therefore tool-chips-only, and the tail stays empty. Codex children *do* stream `item/agentMessage/delta`, but finding 2 keeps the tail from reaching the doc.

Also: the rig runs with a dead edge URL (offline), so the frozen-blob upload path can't be exercised end-to-end here — finished-subagent tabs exercise the designed fallback (doc watch) instead.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/136"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789571347&installation_model_id=425430&pr_number=136&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F136&signature=37a99355c55629b0705c7bd3653d2cc753af1864fc17896375f1d1fd2d9cce12"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
The media rig session doubled as a live E2E pass and caught three real gaps beyond the fixture suites — all fixed in `bed83ba` with regression tests:

1. **Live-segment chip mutations were dropped**: `update_part_fields` didn't carry `subagentRef`/`subagentStatus`/`subagentTail`, so a chip updating while its segment still streamed (the codex fan-out shape) lost them at `SegmentWriter::sync`; only the parked in-place path persisted them (which is why the claude captures worked). Codex fan-out chips now get the "Open subagent" affordance.
2. **Claude background subagents never reached `· done`**: 2.1.228 emits no *tagged* terminal frame — completion arrives as an untagged `system`/`task_notification` carrying the spawning tool's id (shape pinned from the live capture). It now maps to the subagent's tagged Done: chips flip done/failed, transcripts freeze.
3. **Claude subagent transcripts were tool-chips-only**: the wire streams no tagged partial deltas; subagent text arrives as full text blocks on tagged `assistant` frames, which were dropped. They now emit as attributed text — message granularity, which is all this wire offers (the PR body's feed note is corrected accordingly).

Full workspace suite remains green (50 binaries).

PR #144, opened 2026-08-18
Add projects from other drives: a Locations rail in the add-project palette

wingleeio:
## What

The add-project palette's folder browser always started at the device's home directory, so a project living on an external SSD or a second disk meant walking up to `/` and hunting for the platform's mount convention by hand. This adds a **Locations** section to the palette's right rail — directly under Devices, in the same row recipe — listing **Home** plus the device's mounted drives:

palette with Locations rail

Clicking a drive rebrowses the palette at its mount point. The row that owns the browsed path carries the selection wash (longest-mount-prefix match, home outranks the `System` row that covers everything), and the mount path folds into a breadcrumb named after the drive — exactly the way home folds into the device crumb:

browsing T7 Shield

Repos on the drive announce themselves as usual, and adding one works end-to-end (space created with the drive path, git detected):

repos on the drive
project added from the drive

## How

- **`ListDrives` RPC** (ControlRpc, relay-forwardable via `targetDeviceId` like `ListFolders`, so remote devices' drives work through the same path): returns `{drives: [{name, path}]}`.
  - **macOS**: enumerates `/Volumes`; the boot-volume symlink canonicalizes to `/`, so the system drive arrives under its real name ("Macintosh HD").
  - **Linux**: parses `/proc/mounts` for mount points under `/media`, `/run/media`, `/mnt` (octal escapes like `\040` unescaped — the screenshots' "T7 Shield" exercises this), plus a `System` row for `/` so drives mounted elsewhere stay reachable.
  - **Windows**: drive-letter probe.
  - Enumeration runs on the same disposable-worker + 6s wall-clock ceiling as `ListFolders` — a dead network mount fails the listing, not the runtime. Failures are silent in the UI: the section just stays at Home (the browser's own error row already covers "device didn't respond").
- **UI**: `AddSpaceFlow` loads drives alongside the initial folder listing (and on device switch). The rail scrolls if a mount-happy machine outgrows the fixed card body.
- **Icons**: hand-drawn `hard-drive` + `home` glyphs in the Solar Linear style (same precedent as the git-branch/info-circle ports; the set has neither).

## UX notes

Considered a separate drive dropdown and a root-picker step in front of the browser; both add a mode. The rail already established "pick where to browse" (Devices), so Locations extends that idiom — one flat surface, no wizard, and the existing keyboard flow (`←` up, `/`-descend, tab-ghost) is untouched. "Locations" over "Drives" because Home belongs in the list (it's the anchor you return to, and the active-row highlight needs a home to land on when you're not on a drive) — this also matches Finder's sidebar vocabulary.

## Tests

- `linux_drives` parse (media/mnt filtering, octal unescape, System-first sort), mount-point dedupe, unescape edge cases, plus a live `list_drives` smoke — all in `repos.rs`.
- Full `zeron-engine` / `zeron-proto` / `zeron-ui` suites green; clippy clean on the touched crates.
- Rig-verified on headless sway with two tmpfs mounts under `/media/ubuntu` (screenshots above): Locations rows, drive browse, breadcrumb fold, System root browse, and add-from-drive all exercised live.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/144"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789626102&installation_model_id=425430&pr_number=144&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F144&signature=8b7dd33017877ae4ad46b56cafe9878bf050c5fb4990e6a15c3224c78d808d43"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
Overflow check: rig-verified the rail with 12 mounted drives. The whole rail (Devices + Locations, headers included, Finder-sidebar style) scrolls within the fixed card body — clipped at rest, everything down to the info line reachable by wheel:

rail overflowing with 12 drives
rail scrolled to the bottom

rabden:
On Linux, drive partitions can be on any custom path; for example, my drive partitions are on the `/disk2` path. Can it handle that? I think a better option would be allowing users to type any path in the input and be able to detect it instead of saying "not found." Limiting to a set `/home/user` by default might not be the right choice; that can be the default, but not the only option.

wingleeio:
@rabden Good points, both addressed in c6ad564 (rig-verified with a real ext4 partition mounted at `/disk2`):

**Custom mount paths.** The Linux enumeration no longer assumes the conventional locations: any block-device mount (`/dev/*` source) at a depth-one non-FHS path — `/disk2`, `/data`, `/tank` — now shows up as a Locations row. squashfs/erofs images are excluded so snap loops stay out, and the system's own split partitions (`/boot`, a separate `/home` or `/var`) don't appear as drives since Home and System already cover them. Deeper custom mounts aren't enumerated (no reliable way to tell them from plumbing), but they're reachable two other ways — the System row, or:

/disk2 as a Locations row

**Typed paths.** The input now navigates when the query is path-shaped: typing `/` jumps to the root and each `segment/` descends shell-style (with the existing tab-ghost completion helping along the way), while a pasted absolute path — or ⏎ on one — jumps in one hop. `~` expands to the browsed device's home. So home is just the starting point, not a limit:

typed /disk2/repos/ jumped straight there

And on the "not found" point — a path that doesn't resolve now surfaces the real error instead of pretending the device is offline:

honest error for a bad typed path

PR #145, opened 2026-08-18
settings: fix harness toggles, Codex free-tier quota window, double auth tab

### LINKED ISSUE #128: Claude Code is enabled by default and selectable in the main UI even when the CLI is not installed

rabden:
Three fixes to the settings/accounts surfaces, all found while dogfooding on Linux.

## 1. Installed-but-uninstalled harnesses couldn't be toggled off fixes #128 

Default-enabled harnesses with no CLI installed (claude, codex) showed a dimmed-ON toggle that couldn't be turned off, and the composer's agent picker kept offering them. Now any enabled-or-installed harness toggles freely, the hint says what to do (`{cli} CLI not installed — turn it off or install it`), and the picker only offers harnesses with a CLI actually installed.

## 2. Codex free-tier quota window mislabeled "Week"

Queried the live `wham/usage` endpoint against a free account: it reports **one 30-day window** (`limit_window_seconds` 2_592_000) with `plan_type` at the top level. The old `span > 1 day ⇒ "Week"` label mislabeled it, and the reset formatter showed a weekday ~29 days out.

- Window label now derives from the span: ≥28d `Month`, ≥5d `Week`, else `Session` (Plus keeps its 5h `Session` + weekly `Week` pair).
- The usage probe returns the live `plan_type`, so the plan badge (`ChatGPT Free`) comes from the endpoint when a probe ran, falling back to the login-time JWT claim — plan changes surface without a re-login.
- `format_reset`: clock time within ~22h, weekday within a week, month + day (`resets Sep 14`) beyond that.

## 3. Add Codex account opened two identical auth tabs

`codex login` opens the authorization page itself (webbrowser → xdg-open) and the app opened the same URL again when `StartAgentLogin` returned it. The spawned CLI now gets `BROWSER` pointed at a no-op script (created idempotently under `agent-accounts/.noop-browser`, 0755); a failed open is advisory to `codex login` — it still prints the URL and serves the loopback callback. Verified against codex-cli 0.147: with `BROWSER` set, the CLI invokes exactly that binary and login proceeds normally. The dialog's manual "Reopen" link is untouched.

## Testing

- `zeron-ui`: 443 tests pass (incl. new reset-formatting cases)
- `zeron-engine`: agent_accounts unit tests + the m5c accounts integration suite pass (incl. new window-label and no-op-browser tests)
- Manual: Codex card shows `Month · 8% used · resets {month day}`; add-account opens exactly one tab

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/145"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789627963&installation_model_id=425430&pr_number=145&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F145&signature=d152101250387506c38d34259dbacc86b974465bd70456f8d3dca426486fdbf2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (changes requested):
Requesting changes for two blocking correctness issues:

1. `offered_harnesses_impl` falls back to the enabled set whenever the runnable set is empty. On a fresh machine where Claude and Codex are default-enabled but neither CLI is installed, the picker therefore still returns those unrunnable harnesses. That preserves the issue #128 reproduction and contradicts the claim that the picker only offers installed CLIs. Please let the offered set remain empty and surface an empty state / block new sends instead of reintroducing descriptors that can only fail with `NotInstalled`.

2. The last-enabled guard makes the remaining missing harness impossible to turn off. After disabling one of the two missing defaults, the other becomes `last_enabled`, so its toggle is inert even though the new hint says to turn it off. Please either make effective defaults installed-aware or permit the final uninstalled harness to be disabled while handling the empty-selection state safely.

The branch also currently conflicts with `main` in `crates/engine/src/agent_accounts.rs` after the login-child refactor, so it needs an update before the combined tree can be verified.

I verified the exact head: the engine/UI checks, 443 UI tests, five targeted engine tests, and nine account integration tests all pass. The blockers above are behavioral/integration issues rather than test failures.

wingleeio (approved):
Both blockers are addressed at head `a8ef0aa` (plus the `main` merge `040d7e3` resolving the `agent_accounts.rs` login-child-refactor conflict — the `BROWSER` shim now rides `wire_login_child`):

1. `offered_harnesses` no longer falls back to anything: the offered set is strictly enabled ∩ installed (mock opt-in aside). On the fresh-machine shape (Claude/Codex default-enabled, no CLIs) the composer now shows a "No agents available" chip, the model picker renders a guidance card (enable an installed agent in Settings → Agents, or install a CLI), and `send_blocked` gates new-chat sends until an agent is runnable. Loading/error catalogs conclude nothing, so offline sends aren't blocked; an existing chat's committed harness is still force-included in the rail.

2. The last-enabled guard — engine `set_enabled` and the settings toggle — now only protects a harness whose CLI is installed. Both CLI-less defaults can be turned off in sequence; the persisted empty set round-trips (`uninstalled_defaults_can_both_be_disabled`).

Verified on the combined tree: workspace check clean, zeron-ui 480/480, engine registry + agent_accounts unit tests, and the m5c accounts integration suite 13/13.

PR #164, opened 2026-08-18
attachments: 10x chunks, parallel window, overall deadline, live progress

wing-anara:
## Problem

Image sends to a remote device could effectively never finish (observed on 0.2.10): the upload leg is bounded per-*call* but not per-*upload*. Each chunk races a 30s timer with 3 attempts, and the send only aborts after 3 **consecutive** failures on one chunk — on a flapping link most chunks succeed on attempt 2 or 3, so the abort never trips. With 45KB sequential chunks, a ~3MB screenshot is ~70 relay round trips; worst case that's hours of lawful crawling, rendered only as "Sending…". Text-only sends are all-local and instant, which is why only image sends looked hung.

## Fix

- **~510KB chunks** (`UPLOAD_CHUNK_B64_CHARS` 60 000 → 680 000): ~10x fewer round trips, ~35% headroom under Cloudflare's 1MB WebSocket message cap, still a multiple of 4 so b64 slices stay independently decodable. The engine's per-seq slot files are size-agnostic — no engine change.
- **3 chunks in flight** (`try_for_each_concurrent`): seq slots are idempotent so order doesn't matter; the first *window* (not just seq 0) gets the 90s cold-dial allowance.
- **Whole-attachment deadline** (120s + 15s/chunk, capped at 15 min): a crawling upload now fails with the existing composer banner (draft + files handed back) instead of spinning indefinitely.
- **Progress ring on the thumbnail**: while an image crosses the relay, its thumbnail in the echo bubble dims and shows a radial progress ring with the percent centered (stroked-polyline arcs — gpui has no arc primitive). The `pending/` attachment refs + upload-in-flight state identify exactly the images uploading, so nothing else in the UI changes; the "Sending…" label stays as-is. Percent clamps at 99 so the commit+queue tail never shows a stuck 100%.
- Capture knob `ZERON_DEMO_UPLOAD=<pct>:<image path>` (same family as `ZERON_OPEN_DIALOG`): fabricates a sending echo with a frozen ring for styling passes — a real upload is too fast to pause.

## Screenshots

_Coming shortly (captured separately)._

## Testing

- New unit tests: chunk-plan tiling (empty/exact/partial), frame-budget + alignment invariants, deadline scaling/cap, progress percent clamping/clearing (447 UI tests pass).
- Not covered by automation: a real flaky-relay upload — needs a live lossy link. Manual check after deploy: send a screenshot to personal-metal and watch the ring fill; on a dead link expect the red banner within ~2–4 minutes instead of an eternal spinner.

### DIFF
--- a/crates/ui/src/attachments.rs
+++ b/crates/ui/src/attachments.rs
@@ -18,6 +18,7 @@ use std::time::{Duration, Instant};
 
 use base64::Engine as _;
 use base64::engine::general_purpose::STANDARD as BASE64;
+use futures::TryStreamExt as _;
 use gpui::{
     AnyElement, BackgroundExecutor, Image, ImageFormat, ObjectFit, SharedString, Size,
     StyledImage as _, div, img, prelude::*, px,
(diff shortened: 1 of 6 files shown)


wingleeio (approved):
Tested and reviewed at `a729ed4`. No blocking issues found — approving.

## Thumbnail loader

The ring stays centered and legible over detailed imagery. The track/arc render cleanly at 0%, 43%, and the intentional 99% cap:



The fixed white-on-wash treatment also reads well in both light and dark themes:



## Validation

- `cargo test -p zeron-ui` — 447 passed
- Engine upload/readback test — passed
- End-to-end attachment-threading test — passed
- Desktop binary build — passed
- `git diff --check` — passed
- Manual GPUI/X11 captures with llvmpipe at 0%, 43%, and 99%, in light and dark themes

One non-blocking UX note: a multi-image send currently shows the same aggregate percentage on every pending thumbnail. That is coherent as batch progress, but if per-image progress is the intended semantics, the counter would need to be keyed by attachment.

I did not have a live lossy remote relay available, so the flaky-link deadline path was covered by code inspection/unit tests rather than a real 2–4 minute failure run.

Formatting note: `cargo fmt --all -- --check` is not clean repo-wide and also reports two formatting differences in PR-touched hunks.

PR #171, opened 2026-08-19
ios: port the network-reliability train (PRs #159/#164/#165/#168/#170) to mobile

wing-anara:
None of the recent reliability work had reached the iOS client — it landed entirely in the Rust crates, and the phone's parallel Swift stack kept every incident class the desktop just fixed. This ports the client-side halves. The engine-host halves (worktree materialization at drain, pending-ref defer + rewrite, staging sweep fix, echo reply) are already deployed on hosts ≥ 0.2.12, so the phone gets them by speaking the same wire.

## Transport (PR #168)
- **OnlineBus** (wake.rs port): every room join broadcasts "network is back"; `NWPathMonitor` parks backoffs while the path is definitively down (30s safety recheck) and broadcasts on recovery — including satisfied→satisfied interface handovers; the foreground hook probes `{edge}/health` (3s) so a focus lands redials in ~1 RTT.
- **Room clients**: backoff cap 30s→16s, stability-gated reset (a join must survive 30s — reset-on-join lets connect-and-die sockets hot-loop at 250ms), backoff waits ride the bus.
- **DeviceRelayClient** (was the least-defended client: 30s blind ping, no liveness, no backoff): 10s keepalive / 25s transport lease / end-to-end `echo` frame with a 20s host-proof deadline (feature-detected; inbound RPC frames count). The DO's auto-pong answers from the edge, so a healthy client↔edge leg masked a dead edge↔host leg for 6 minutes in the Aug 19 zombie-peer-link incident — this kills that class. A timed-out call drops the link as suspect (next call re-dials); a registry-dark liveness verdict (positive evidence only: joined + 60s warm-up + freshest heartbeat ≥ 5min old) fails peer calls fast with zero dials.

## Send truth (PRs #165/#168/#170)
- **ConnectivityCenter**: graced connectivity — a source (OS path / registry / per-chat room) must be raw-degraded **4s continuously** before it reports; recovery is instant. Sub-second blips (room joins on navigation) never flash the UI.
- Home bar shows a quiet grayscale "Reconnecting…" / amber "Offline — sends are saved" only while degraded; rows get **Queued**/**Failed** corner pills that suppress the fake Working spinner; the session strip shows "Sending…" / "Queued — will send automatically" / "**Not delivered — tap to retry**" (2-minute undelivered grace; retry restarts the clock, kicks the room on fresh backoff, nudges the host, re-derives attachment escorts). The composer gets one caption line of pre-send honesty on a degraded path.

## Attachments (PRs #164/#168)
- Uploader parity with #164: **~510KB chunks** (was ~45KB sequential — 112 round trips for a 5MB photo), 3-wide parallel window, retries staggered by seq, whole-upload deadline `min(120+15/chunk, 900)s`, "Uploading… N%" narration.
- **Queued flow** (host ≥ 0.2.12, gated on the registry device row's `version`): the command queues immediately with `pending://{uploadId}/{name}` refs — a durable local write — bytes stash to disk (`UploadStash`, survives relaunch; re-derived from the doc's pending refs on start/retry) and an escort pushes them with retry-forever on the online bus (2s→30s, 15min cap), then nudges the drain. An image send no longer dies with its link. Old hosts keep the legacy upload-first flow, now bounded and narrated.

## New chats (PRs #159/#170)
- New-worktree sends attach `RunRequest.worktree {repoPath, base}` for the **host to materialize at drain time** — the blocking `CreateWorktree` relay RPC before send (the one new-chat path that could hang forever on a zombie link) is gone. Base defaults to `HEAD` so a branch list that never loaded over a flapping relay can't silently drop isolation. Old hosts ignore the field and run in the main checkout: degraded, never hung.
- On legacy hosts, attachments stage **before** the chat row exists — an upload failure aborts with nothing created (no stranded empty chat).

## Not ported (follow-ups)
- **RelayCommand second road** (engine forwards the queued entry over the peer link when rows can't reach the edge): the phone already has an HTTPS push/pull fallback per room, so the marginal win is small; needs a rows-flushed signal from ChatRoomClient if we want it.
- The attachment **read-back** cache keeps its own ungated relays (it inherits the new echo/timeout hygiene, not dial parking).

## Tests
- 12 new unit tests (`NetworkReliabilityTests`): version-gate parsing, pending:// round-trips, stash lifecycle, chunk-plan wire invariants (b64 %4, CF 1MiB cap headroom), deadline formula, echo frame codec, timeout `owns` gating, `RunRequest.worktree` wire shape (omitted-when-nil for old hosts).
- Full suite: 46 tests, 0 failures. Demo-mode smoke on the iPhone 17 Pro sim renders clean (no spurious connectivity pill).

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/171"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789740064&installation_model_id=425430&pr_number=171&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F171&signature=9e03a70471791b62a9c57d5e6fd654735b0078d5ef5d37c44a9be19c48df3580"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/ios/Zeron/App/AppModel.swift
+++ b/apps/ios/Zeron/App/AppModel.swift
@@ -20,6 +20,9 @@ final class AppModel {
     var phase: Phase = .signedOut
     var workspace: WorkspaceStore?
     var demo: DemoDataset?
+    /// Graced connectivity truth — one stream every consumer inherits calm
+    /// from (home pill, composer notice, Queued/Failed badges).
+    let connectivity = ConnectivityCenter()
     private var sessionStores: [String: SessionStore] = [:]
     private var config: AppConfig?
     @ObservationIgnored private var pathMonitor: NWPathMonitor?
@@ -254,7 +257,27 @@ final class AppModel {
         let store = WorkspaceStore(config: config)
         workspace = store
         store.start()
+        startConnectivity()
         phase = .ready
+    }
+
+    /// Wire the graced-connectivity recompute over the live stores (the
+    /// engine's 1s compute_connectivity, phone edition).
+    private func startConnectivity() {
+        connectivity.registryConnected = { [weak self] in
+            guard let self, self.demo == nil, let workspace = self.workspace else { return true }
+            return workspace.connected
+        }
+        connectivity.chatRooms = { [weak self] in
+            guard let self else { return [] }
+            return self.sessionStores.compactMap { id, store in
+                store.roomActive ? (id: id, connected: store.connected) : nil
+            }
+        }
+        connectivity.hasPendingSends = { [weak self] in
+            self?.sessionStores.values.contains { !$0.pendingSends.isEmpty } ?? false
+        }
+        connectivity.start()
     }
 
     // MARK: Unified data accessors (demo or live — one path for views)
@@ -510,9 +533,24 @@ final class AppModel {
     /// Foreground hook: kick every room NOW (see ChatRoomClient.kick) — after
     /// a suspension the workspace room in particular stayed dead while chat
     /// views reconnected on open, freezing sidebar rows and Working
-    /// indicators against perfectly live transcripts (2026-08-04).
+    /// indicators against perfectly live transcripts (2026-08-04). Also the
+    /// focus fast path (PR #168): probe {edge}/health (3s) and broadcast the
+    /// online event on success, so every PARKED backoff (not just the rooms
+    /// the kick reaches) lands a redial in ~1 RTT.
     func foregrounded() {
         kickAllRooms()
+        probeEdgeHealth()
+    }
+
+    private func probeEdgeHealth() {
+        guard let config, demo == nil else { return }
+        Task.detached {
+            var request = URLRequest(url: config.edgeURL.appending(path: "health"))
+            request.timeoutInterval = 3
+            guard let (_, response) = try? await URLSession.shared.data(for: request),
+                  (response as? HTTPURLResponse)?.statusCode == 200 else { return }
+            OnlineBus.shared.notifyOnline()
+        }
     }
 
     private func kickAllRooms() {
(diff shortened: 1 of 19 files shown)


wing-anara:
Pushed a third commit porting **PR #172** (merged upstream after this PR opened) — it matters most of anything here:

- **Cursor contiguity + gap repair + universal amnesty**: the Swift `ChatRoomClient` carried the identical `cursor = max(cursor, seq)` root cause of the random new-session forever-hangs (live row outruns the backfill → cursor stamps over the hole → dependents park invisibly in loro → empty doc under a "converged" cursor, surviving restarts). Now the cursor walks, never jumps; gaps hold the honest cursor and request a bounded `rowsReq` repair (3/session, then redial); checkpoint-less rooms amnesty to 0 on first join, healing any already-wedged chat on disk.
- **Retry re-issues dead attempts** (phone half): a send whose command went rejected/expired/pending-past-TTL gets a fresh command id with the same payload + messageId on the retry tap — the host's user-entry pre-write dedupes, so retry is per-attempt while exactly-once stays per-id.
- Status strip now narrates the escort's real relay transfer as "Uploading… N%" (#185's progress-truth analog).

47 tests green (14 in the reliability suite, incl. a re-issue exactly-once test).

Audited the rest of the last ~20 merges for mobile-relevant network work: #147 is edge-only (free for the phone), #180/#185 are desktop thumbnail polish (covered by the strip narration), everything else is harness/terminal/desktop-UI. Nothing further to port.

PR #146, opened 2026-08-18
ui: real frosted floats on Linux (wgpu backdrop blur)

rabden:
Linux currently gets no frosted floats: the wgpu renderer skipped `BackdropBlur` (only the Metal path rasterizes it), so the composer pill and popover recipes degraded to faint transparent smudges. This turns on real backdrop blur on Linux, ported at the renderer level.

## Changes

- **`Theme::is_frost()`** separates *scene-level* frost (floating surfaces blurring in-app content) from `is_glass()` *window* vibrancy (the blurred desktop behind a translucent window). Frost needs no compositor support — the renderer can do it in-app — so it is now enabled on Linux; window-level glass remains macOS-only (no blur-behind guarantee outside KWin).
- Composer pill + question panel, popovers/modals, and the settings card recipes paint their backdrop blur and translucent tints on Linux (and drop the opaque shadow slab that only made sense over an opaque fill). macOS behavior is unchanged (`is_frost == is_glass` there).
- **gpui rev** points at the renderer port branch until wingleeio/zed#2 ("wgpu: rasterize BackdropBlur") merges — then it should be rebased back to `wingleeio/zed`'s `comet/edge-fade-horizontal` tip.

## Testing

Dogfooded on GNOME Wayland (RADV/Vulkan, Radeon 610M) with the vendored renderer change: pill and popovers genuinely frosted, blur tracks transcript scrolling, resize small↔large stays clean, no GPU validation errors over a long session, idle CPU ~2% / ~214 MB with blur active. `zeron-ui`: 443 tests pass.

Depends on wingleeio/zed#2.

---
[...]

PR #173, opened 2026-08-19
composer: a full-width @-mention panel over an indexed file search

neo773:
Before

image

After
image

The `@` panel was a narrow card hanging off the caret, one long grey path per row, and every keystroke re-walked the whole checkout to fill it.

## The panel

It now spans the composer and stacks on its top edge, and a row reads `{filename} {directory}` — the name carries the match and never truncates, the dimmed directory disambiguates same-named files and is the side that gives up space. New `popover::full_width_menu_above` mounts it: an absolute layer inset to `left-0 right-0` of a `relative()` wrapper around the pill, deferred so it paints and takes hits over the transcript it overhangs. Mounting had to move from around the input to around the pill — the input's box is not the width the panel wants.

The per-row path tooltip is gone; the row already shows the whole path.

## The search

Two changes, both in `repos.rs`:

**One walk, not one per keystroke.** The walk is now `ignore`'s `build_parallel()` (ripgrep's own crate, already a dependency — `.gitignore`, `.ignore`, global excludes and `.git` pruning all unchanged) and its result is cached per checkout for 10s. A keystroke ranks that cache in memory. Visitors batch 512 entries per lock and flush their tail on drop, since the walker drops one visitor per thread. Stale indexes are evicted whenever one is rebuilt, so a long session does not accumulate a file list per folder ever searched.

**A matcher built for source trees.** Scoring was a hand-rolled subsequence distance that lowercased both sides per candidate per keystroke. It is now `nucleo-matcher` (Helix's) in `match_paths` mode, which weights the last path segment — the property that ranks `src/composer.rs` over `composer/docs/readme.md`. Each entry's haystack is pre-built at walk time (ASCII paths stay a boxed str, no UTF-32 expansion), and one matcher plus one parsed pattern serves the whole pass. Ranking keeps a bounded buffer, so `@` with no query never sorts the whole tree to pick 8 rows.

Score direction flipped — nucleo's is higher-is-better — so the comparator's score arm inverted; the featured-paths and empty-query tiebreaks are untouched.

## Testing

`cargo test -p comet-engine` and `cargo clippy` clean. Existing search tests carry over unchanged (gitignore, featured paths, filename-over-path ranking, cancellation), plus a new one pinning that a second keystroke ranks the cache instead of re-walking.

Driven in the running app.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/173"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789741409&installation_model_id=425430&pr_number=173&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F173&signature=3e3189ac5ebfa82ecb66cace1fd146764249dcb60c6e5754cca9b5773730dafd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4266,6 +4266,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "nucleo-matcher"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf33f538733d1a5a3494b836ba913207f14d9d4a1d3cd67030c5061bdd2cac85"
+dependencies = [
+ "memchr",
+ "unicode-segmentation",
+]
+
+[[package]]
 name = "num"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -9143,6 +9153,7 @@ dependencies = [
  "libc",
  "loro",
  "notify",
+ "nucleo-matcher",
  "portable-pty",
  "reqwest",
  "rusqlite",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,7 @@ pulldown-cmark = "0.12"
 unicode-segmentation = "1"
 reqwest = { version = "0.12", default-features = false, features = ["rustls-tls", "json", "stream"] }
 ignore = "0.4"
+nucleo-matcher = "0.3"
 
 [profile.release]
 # Distribution profile (scripts/package-linux.sh): thin LTO buys most of fat
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -33,6 +33,7 @@ tokio-util = { workspace = true, features = ["rt"] }
 sha2.workspace = true
 reqwest.workspace = true
 ignore.workspace = true
+nucleo-matcher.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
--- a/crates/engine/src/repos.rs
+++ b/crates/engine/src/repos.rs
@@ -39,6 +39,9 @@ const DRIVE_LIST_MAX_ENTRIES: usize = 50;
 const FILE_SEARCH_MAX_RESULTS: usize = 8;
 /// A dead network mount must not leave the composer search spinning forever.
 const FILE_SEARCH_TIMEOUT: Duration = Duration::from_secs(6);
+const FILE_INDEX_TTL: Duration = Duration::from_secs(10);
+const FILE_INDEX_MAX_ENTRIES: usize = 250_000;
+const RANK_BUFFER: usize = 1_024;
 pub const GIT_HISTORY_DEFAULT_LIMIT: usize = 100;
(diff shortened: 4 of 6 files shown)


wingleeio (approved):
Reviewed in depth, verified locally on current main, and drove it in the running app — approving. The branch was a week behind main, so I resolved the (mechanical) conflict and am admin-merging the result; no action needed.

**Verification done:**
- Merged onto current main. The repos.rs conflict was adjacent insertions (main's drive-listing functions from #144 vs this rewrite); resolution keeps both, drops the superseded `fuzzy_score`. Main made no search-side changes since your base, so the rewrite loses nothing.
- `nucleo-matcher` 0.3.1 checksum verified against crates.io.
- Full `cargo test -p zeron-engine` green on the merge (all suites, including the new cache-reuse test). fmt/clippy clean in touched files.
- Headless rig smoke on the merge (Linux, post-#146 frost): panel mounts full-width over the pill and paints/hit-tests over the transcript; empty `@` lists 8 root files shallow-first; `cmposer` ranks the `Composer` dir + both `composer.rs` files on top with dimmed directories; clicking a row inserts the mention chip. The layering survives main's frost changes that landed after your base.
- Ranking details check out: the bounded rank buffer keeps the true top-8 (sort+truncate retains best-so-far), and the comparator's final full-path tiebreak makes results deterministic despite the parallel walk's nondeterministic order.

**Non-blocking notes:**
- A file created less than `FILE_INDEX_TTL` (10s) after a walk won't appear in `@` until the TTL lapses — the PR states this; worth remembering if reports come in.
- On a huge checkout with fast typing, a cancelled walk doesn't populate the cache, so keystrokes during the very first walk each restart it — same as the old per-keystroke behavior, strictly better once one walk completes.
- Stale index eviction only runs when some rebuild happens, so a fully idle session keeps the last ≤TTL-old indexes in memory — bounded and harmless.

Nice work — the panel reads much better, and the search is night-and-day on a big tree.

wingleeio:
Merged to main as 6e81686 — the branch was conflicting with post-#144 main, so I resolved the (mechanical) repos.rs conflict and admin-merged locally; your commit 447b689 is the merge's second parent, authorship preserved. GitHub's merged-badge auto-detect didn't fire on the pushed merge, hence the manual close. Thanks again!

PR #182, opened 2026-08-20
fix(acp): support Grok Build model switching

rabden:
## Issue

When a user switched models during an existing Grok Build ACP session, the
selected model was not applied correctly. Grok Build advertises its models
through ACP's first-class `session.models` state and expects model changes via
`session/set_model`, rather than through a `configOptions` entry with
`category: "model"`.

The existing flow only handled the config-option path. As a result, the model
picker could appear to accept a new model while the running ACP session kept
using the old model. In failure cases, the prompt could stop without a useful
application-level response or diagnostic.

## Fix

- Detect first-class ACP model advertisements when no model config option is
  present.
- Validate the requested model against the agent's advertised models.
- Send `session/set_model` with the selected `modelId` before starting the
  prompt.
- Treat an explicitly rejected first-class model switch as a protocol error so
  the failure is surfaced instead of silently continuing with the wrong model.
- Preserve the existing config-option behavior for adapters that expose model
  selection through `configOptions`.
- Keep runtime routing aware of model and run configuration changes so a live
  harness is restarted/resumed when the existing process cannot apply the new
  configuration.
- Add fake-ACP integration coverage and unit tests for the first-class model
  path and runtime configuration behavior.

## Validation

- `cargo test -p zeron-harness --test acp resumed_first_class_model_is_switched_before_prompt`
- `cargo test -p zeron-engine live_routing_requires_the_same_runtime_configuration --lib`
- `cargo test -p zeron-harness first_class_models_use_session_set_model_without_config_option --lib`
- `sh -n crates/harness/tests/fixtures/fake-acp.sh`
- `git diff --cached --check`

No local low-RAM Cargo configuration or ignored development plan files are
included in this PR.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/182"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789791858&installation_model_id=425430&pr_number=182&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F182&signature=03c56617a889f840aa68b066529b5ded6ac2258909f9ee32580d0bb25a6a3180"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/engine/src/sessions.rs
+++ b/crates/engine/src/sessions.rs
@@ -67,9 +67,45 @@ struct HarnessSessionRef {
     cwd: String,
 }
 
+/// Configuration baked into a live harness runtime. The steering mailbox only
+/// carries prompt text, so routing a request whose model/effort/sandbox changed
+/// would silently run it with the old process configuration. Such requests
+/// must replace the runtime and resume its harness-native session instead.
+#[derive(Debug, Clone, PartialEq)]
+struct RuntimeConfig {
+    harness_id: HarnessId,
+    model: Option<String>,
+    reasoning: Option<zeron_proto::ReasoningLevel>,
+    model_options: serde_json::Map<String, serde_json::Value>,
+    cwd: String,
+    sandbox: zeron_proto::SandboxLevel,
+    auto_approve: bool,
+    worktree: Option<zeron_proto::WorktreeSpec>,
+}
+
+impl RuntimeConfig {
+    fn from_request(harness_id: HarnessId, request: &RunRequest) -> Self {
+        Self {
+            harness_id,
+            model: request.model.clone(),
+            reasoning: request.reasoning,
+            model_options: request.model_options.clone(),
+            cwd: request.cwd.clone(),
+            sandbox: request.sandbox,
+            auto_approve: request.auto_approve,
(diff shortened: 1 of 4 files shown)


rabden:
@wingleeio recommended to include in next release as without this the whole grok agent is literally unusable.
well grok build is one of the few well built cli made in rust. its compact, solid and optimized just like zeron so this agent deserves a bit of care for zeron at least

wingleeio (approved):

PR #183, opened 2026-08-20
fix: harden OpenCode model discovery

rabden:
## Problem

OpenCode model discovery used a fixed 10-second timeout even though the ACP runtime handshake allows 120 seconds. On plugin-heavy installations, OpenCode can take tens of seconds or minutes to initialize its plugin/MCP graph, causing the model picker to cache a timeout or static fallback rather than the actual catalog.

## Fix

- Give OpenCode one shared, configurable five-minute startup budget for both the ACP handshake and model discovery (`ZERON_OPENCODE_STARTUP_TIMEOUT_SECS`).
- Surface OpenCode discovery failures instead of silently substituting static catalog rows.
- Serialize concurrent discovery probes so opening/rendering the picker cannot spawn duplicate OpenCode processes.
- Retry transient OpenCode catalog failures twice in the picker, preserving already loaded rows while refreshing.
- Add a thin floating model-list scrollbar which appears only while its scrollable list is hovered; it supports track clicks and captured thumb dragging without layout reflow.

## Validation

- `cargo test -p zeron-harness`
- `cargo check -p zeron-ui`
- `cargo test -p zeron-ui --lib`

### DIFF
--- a/crates/harness/src/acp/mod.rs
+++ b/crates/harness/src/acp/mod.rs
@@ -56,6 +56,23 @@ use normalize::{map_update, parse_commands, preferred_allow_option};
 use subagent::SubagentTracker;
 use subagent_opencode::OpencodeTracker;
 
+const DEFAULT_HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(120);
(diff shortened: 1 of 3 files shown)


rabden:
@wingleeio without this opencode models dont even load.

wingleeio (approved):

PR #192, opened 2026-08-20
Fix tool rendering: only agent calls render as subagent chips

wingleeio:
## Summary

Fixes a rendering regression where regular tool calls (like Run commands) were being rendered as clickable subagent chips. After PR #189 separated agent spawn chips from ordinary tools in the grouping logic, the rendering code needed clarification about which tools should actually render as subagent spawn UIs.

The fix ensures that only actual agent calls (matching the "Agent" naming pattern) with a document reference render as subagent spawn chips. Regular tools like Run, Read, etc. now correctly render as normal tool chips even if they have a subagent_ref field.

## Root Cause

The rendering check at line 3808 was only checking `tool.subagent_ref.is_some()` without verifying that the tool was actually an agent call. This could lead to non-agent tools appearing as clickable subagents that opened empty panels.

## Test Plan

- [x] Verify Run commands no longer appear with subagent chip styling
- [x] Verify actual Agent calls still render as subagent spawn chips when docs are available  
- [x] Verify mixed tool groups still render correctly (collapsed)
- [x] Code compiles without errors

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/192"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789853106&installation_model_id=425430&pr_number=192&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F192&signature=a693ef679c2bab0b24a3a062308569a0808954ca0b8911f8346e71c26017a7e8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/doc/src/parts.rs
+++ b/crates/doc/src/parts.rs
@@ -389,10 +389,16 @@ pub fn fold_event_into_parts(out: &mut Vec<MessagePart>, event: &AgentEvent) {
             for p in out.iter_mut() {
                 if let MessagePart::Tool {
                     id,
+                    call,
                     subagent_status,
                     ..
                 } = p
                     && id == parent_tool_use_id
+                    // Genus gate: only a SPAWN call ever carries subagent
+                    // lifecycle. A driver keying bug (claude's background
+                    // shells settled through the subagent subtype,
+                    // 2026-08-20) must not decorate an ordinary chip.
+                    && call.is_subagent_spawn()
                 {
                     match status {
                         Some(s) => *subagent_status = Some(s),
@@ -961,6 +967,42 @@ mod tests {
         }
         // Content never leaked into the parent parts.
         assert_eq!(parts.len(), 1);
+    }
+
+    #[test]
+    fn subagent_events_never_decorate_a_non_spawn_chip() {
+        // Mis-keyed tagged traffic (claude's background shells settled
+        // through the subagent subtype, 2026-08-20) must not stamp lifecycle
+        // onto an ordinary tool chip — the genus gate is the CALL.
+        use zeron_proto::DoneStatus;
+        let mut parts = Vec::new();
+        fold_event_into_parts(
+            &mut parts,
+            &AgentEvent::ToolCall {
+                id: "toolu_bash".into(),
+                call: ToolCall::Exec {
+                    command: "git clone …".into(),
+                },
+            },
+        );
+        fold_event_into_parts(
+            &mut parts,
+            &AgentEvent::Subagent {
+                parent_tool_use_id: "toolu_bash".into(),
+                event: Box::new(AgentEvent::Done {
+                    status: DoneStatus::Completed,
+                    result: None,
+                    error: None,
+                    session_id: None,
+                }),
+            },
+        );
+        match &parts[0] {
+            MessagePart::Tool {
+                subagent_status, ..
+            } => assert_eq!(*subagent_status, None),
+            other => panic!("{other:?}"),
+        }
     }
 
     #[test]
--- a/crates/doc/src/schema.rs
+++ b/crates/doc/src/schema.rs
@@ -554,6 +554,24 @@ impl SessionDoc {
                     Some(loro::ValueOrContainer::Value(LoroValue::String(s))) if s.as_str() == part_id
                 );
                 if is_tool && id_matches {
+                    // Genus gate: subagent lifecycle only ever lands on a
+                    // SPAWN call. Mis-keyed tagged traffic (a driver bug —
+                    // claude's background shells settled through the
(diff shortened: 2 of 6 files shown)


wingleeio:
Pushed 18987da on top with the root cause — the UI gate alone hid the symptom, but the data was genuinely poisoned upstream.

## Why Run chips had a `subagent_ref` at all

Claude Code settles background **shell** tasks (`Bash` with `run_in_background`, incl. auto-backgrounded long commands) through the **same `task_notification` subtype** as background subagents, carrying the Bash call's own `tool_use_id`. The normalizer gates `task_started` on `subagent_type` (shell tasks lack it) — but `task_notification` had no such gate: every terminal notification became a tagged subagent `Done` keyed to the Bash tool id. The engine then stamped `subagent_ref = chat--sub--<bash id>` onto the Run chip, while the `done_only` path (correctly) never created that doc — hence the empty panel on click.

Latent since subagent viz landed: pre-#189 the poisoned chip hid inside the collapsed "Ran N commands" fold; #189's always-visible spawn rows surfaced it.

## What the commit adds

- **claude driver (root):** `task_notification` only settles a *known spawn* — `Agent`/`Task` tool_use ids recorded from their own assistant frames (plus `task_started`'s agent pairing) gate the tagged Done. Background shells now produce no subagent traffic at all (no bogus ref, no bogus `subagent_status`, no false `settled_subagents` entry).
- **engine (guard, all drivers):** a `Done` with no sink is chip-only *for real* — it no longer stamps a ref linking to the never-created doc.
- **ui:** kept your gate, expressed as `is_spawn_link` (no `unwrap`), and applied it to the detail/invocation gates too — so a poisoned old-doc Run chip regains its expandable output; `is_agent_tool` now keys on the call alone, so those chips also fold back into "Ran N commands" instead of splitting into always-open rows.

## Tests

- `shell_task_notification_never_settles_a_subagent` (regression, wire shapes from the live captures)
- `task_notification_settles_the_subagent_with_a_tagged_done` updated to seed the spawn frame first (matches the live wire — the spawn always precedes its notification)
- `stray_subagent_ref_on_a_run_chip_stays_an_ordinary_tool` (old-doc healing)

`zeron-harness` / `zeron-ui` / `zeron-engine` suites green.

wingleeio:
Added 569793e: harness-wide hardening per follow-up discussion.

**Audit of every driver's subagent keying** (how tagged traffic gets bound to a parent chip):
- **codex** — keys by the `subAgentActivity` item id, which *is* the Agent chip; the child `thread/started` fallback keys to the child thread id (matches no tool part → benign). ✅
- **cursor** — tags by the SDK's own `parent` linkage; spawn Done/opening-prompt additionally gated on `name == "task"`. ✅
- **grok** — binds only through the `spawn_subagent` pending queue. ✅
- **opencode** — registers pending chips only for the `subagent_type`+`prompt` rawInput shape; settlement requires a known chip. ✅
- **claude** — was the one mis-keyer (background-shell `task_notification`), fixed in 18987da. ✅

All were individually safe, but the guarantee lived in five separately-maintained keying rules plus a UI predicate duplicated from the drivers' naming convention. Now it's structural:

`ToolCall::is_subagent_spawn()` (proto) is the single genus definition — the `Agent[: desc]` shape every driver decodes its spawn tool into — enforced at every boundary:
- **engine**: the folded-segment ref stamp skips non-spawn parts;
- **doc fold**: tagged lifecycle never sets `subagent_status` on a non-spawn;
- **doc schema**: `update_subagent_chip` refuses non-spawn parts (covers the flushed-chip path);
- **ui**: `is_agent_call` delegates to the shared predicate.

A future driver keying bug now degrades to an orphan subagent doc at worst — never a Run/Read chip masquerading as an openable-but-empty spawn link. New tests: `subagent_events_never_decorate_a_non_spawn_chip` (doc fold), `update_subagent_chip_refuses_non_spawn_parts` (doc schema, also proves the loro→JSON call round-trip the gate relies on). All suites green across `zeron-proto`/`doc`/`engine`/`harness`/`ui`.

PR #193, opened 2026-08-20
Spawn chips name their subagent's model; stop the killed-subagent spinner

neo773:
image

Two things, one about subagent spawn chips.

## The chip now says which model the child runs on

A fan-out of spawns rendered identically whether the children were `haiku` or `opus`. The only way to find out was to open each subagent tab.

The model was already on the wire — a spawn's input carries it whenever the call named one (claude's `Agent` takes `model`, grok's `spawn_subagent` a `model_id`). What was missing:

- `ToolCall::subagent_model()` reads it off the call, gated on the existing `is_subagent_spawn` genus so an ordinary tool with a stray `model` argument is never mistaken for a spawn. Four key spellings, since drivers disagree on it.
- `sanitize_tool_call` was dropping every `Unknown`/`Mcp` input wholesale, which is what had stranded the model host-side. It now keeps a whitelist of two short identifiers, for spawns only. The prompt still never leaves the host's run journal.
- The chip renders it in the trailing slot as bare faint text — not suffixed onto the description (that's the truncating slot, and it's exactly what gets cut in the fan-out this is for), and not a filled pill (the tiles either side are affordances; giving a passive label the same chrome made the loudest thing in the row the one thing you can't click).

A spawn that named no model renders nothing — the child inherits the parent's, which the chip already implies.

## Killing a subagent left its chip spinning forever

The CLI puts two frames on the child feed when a subagent is stopped, in this order:

```
done{interrupted}                       <- the chip settles
user "[Request interrupted by user]"    <- the chip goes back to running
```

A tagged user frame means "the parent steered its subagent" — it announces more work, and is therefore the one event allowed past the fold's no-regress guard onto a settled chip. The interruption marker announces the opposite, and nothing follows it, so the chip the `done` had just settled spun for the rest of the session.

It's filtered now, one line up from where `<system-reminder>` injections were already filtered: CLI bookkeeping riding a user frame is not conversation. Prefix-matched, since the CLI ships at least two spellings.

## Testing

Ran against a live claude session: spawned two haiku subagents, confirmed both chips read `haiku`, then killed them and confirmed both settle instead of spinning.

New tests cover all four model-key spellings, MCP-shaped spawns, non-spawn rejection, the prompt still being stripped, sanitize idempotency, both interruption-marker spellings, and — importantly — that a real steer on the identical frame shape still gets through.

One pre-existing failure on this tree, unrelated and present on a clean checkout: `shell_env::unix::tests::falls_back_when_interactive_attempt_hangs`.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/193"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789858872&installation_model_id=425430&pr_number=193&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F193&signature=fbdd549135686365d602db306fe072c2638ffc81826ae887bfff03c72298a875"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/doc/src/parts.rs
+++ b/crates/doc/src/parts.rs
@@ -5,7 +5,7 @@
 
 use serde::{Deserialize, Serialize};
 
-use zeron_proto::{AgentEvent, ToolCall, ToolDiff, UserInputQuestion};
+use zeron_proto::{AgentEvent, SUBAGENT_INPUT_KEEP, ToolCall, ToolDiff, UserInputQuestion};
 
 use crate::constants::MSG_INLINE_MAX;
 
@@ -426,7 +426,6 @@ pub fn fold_event_into_parts(out: &mut Vec<MessagePart>, event: &AgentEvent) {
     }
 }
 
-
 /// Stamp sidecar keys onto resolved tool parts that have sidecar content.
 ///
 /// Separate from the fold because the fold is chat-agnostic and pure; the
@@ -485,7 +484,9 @@ pub fn sidecar_payload(event: &AgentEvent) -> Option<SidecarPayload> {
 
 /// Render-only privacy policy — strip heavy/sensitive tool inputs before a call enters the doc.
 ///
-/// Keeps: command / path / pattern / url / query / todo items / server+tool names.
+/// Keeps: command / path / pattern / url / query / todo items / server+tool names,
+/// and a subagent spawn's model/type (see [`spawn_badge`] — a couple of short
+/// identifiers the chip names the child by).
 /// Drops: WriteFile content, EditFile old/new strings, WebFetch prompt, Mcp/Unknown input.
 /// Full inputs remain only in the host's local run journal. Idempotent.
 pub fn sanitize_tool_call(call: &ToolCall) -> ToolCall {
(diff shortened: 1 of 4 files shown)


wingleeio (approved):
No blocking findings. I traced the Claude normalization, tool-input sanitizer, protocol accessor, and chip render path end to end.

Verified:

- `done{interrupted}` followed by either known `[Request interrupted…]` marker no longer resurrects a settled chip, while a real steer on the same tagged-user frame shape still passes.
- The spawn model survives sanitization while the prompt and ordinary-tool inputs remain stripped; all four key spellings, MCP-shaped spawns, non-spawn rejection, and sanitizer idempotency are covered.
- A headed mock run with distinct `haiku` / `opus` model inputs kept the model labels stable through running and settled states; both lifecycle spinners cleared without moving the rows.

Running:

Running subagents show distinct model labels

Settled:

Settled subagents retain model labels and clear their spinners

Validation: `cargo test -p zeron-proto -p zeron-doc -p zeron-harness -p zeron-ui` — 727 passed, 0 failed; `cargo build -p zeron` also passed. The reported shell fallback test passed here. A separate workspace-wide rerun was runner-limited by the `/tmp` quota during duplicate linking, not by a test failure.

PR #195, opened 2026-08-21
Preserve transcript viewport across chat switches

jsgrrchg:
## Summary

Preserve each chat's transcript viewport when switching between conversations, preventing the UI from unexpectedly jumping back to the bottom.

This works for existing chats with replayed history, new chats with optimistic messages, and chats that are actively streaming.

## Problem

The transcript list is shared across chat tabs and resets whenever a different chat is attached.

Because transcript replay and layout measurement happen asynchronously, the list could lose the user-owned scroll position and re-engage bottom-follow behavior. This caused chats to jump to the bottom after tab switches or while the active turn changed the transcript geometry.

## Changes

- Add session-local viewport restoration for primary chats.
- Store up to 256 viewport snapshots using bounded LRU eviction.
- Anchor restoration to stable row and entry identifiers instead of relying only on row indexes.
- Distinguish pending, empty, and populated transcript replay states.
- Delay fallback restoration until the authoritative transcript reset arrives.
- Guard deferred restoration with chat-generation and layout-revision tokens.
- Preserve active-turn runway state when switching away from a streaming chat.
- Retire stale runway state when an optimistic prompt disappears after a failed send.
- Cancel pending restoration when the user explicitly scrolls or navigates.
- Hand viewport ownership to rail navigation consistently in animated and reduced-motion modes.
- Recalculate distance-based viewport state after the restored layout has stabilized.

## Behavior

- Chats that were following the tail continue following the tail.
- Chats where the user had scrolled upward restore the previous visible content.
- Returning to a cached chat marks it as the most recently used viewport.
- Subagent transcripts do not consume entries from the primary chat viewport cache.
- Viewport state remains memory-only and is not persisted across application restarts.

## Performance

The cache stores only small viewport anchors and metadata, not transcript contents.

Streaming adds no permanent polling loop. Layout finalization runs only while restoration or active-turn geometry is unsettled, and the cache is capped at 256 primary chats.

## Testing

- Added coverage for stable row restoration across transcript replay.
- Added entry and row-index fallback coverage.
- Added pending replay and optimistic echo coverage.
- Added active-turn runway restoration and failed-send cleanup coverage.
- Added LRU capacity and recency coverage.
- Added stale generation and unstable layout finalization coverage.
- Added rail navigation coverage for direct and animated scrolling.
- `cargo test -p zeron-ui --lib` (498 passed)
- `rustfmt --edition 2024 --check crates/ui/src/rail.rs crates/ui/src/state.rs crates/ui/src/transcript.rs`
- `git diff --check`

## Formatting note

The repository-wide `cargo fmt --check` currently reports pre-existing formatting differences outside the files changed by this PR. The three modified files pass an isolated `rustfmt --check`.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/195"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789868707&installation_model_id=425430&pr_number=195&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F195&signature=cd9846bd986b80e3ef40cd4b7f71d777a5e713f8683acbd4faaee44a5af1ccf0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/ui/src/transcript.rs
+++ b/crates/ui/src/transcript.rs
@@ -26,7 +26,7 @@
 //! whole turn is already visible, so there is nothing to scroll to.
 
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
 use std::ops::Range;
 use std::rc::Rc;
 use std::sync::{Arc, Weak};
@@ -61,6 +61,8 @@ pub const STICK_THRESHOLD_PX: f32 = 70.0;
 pub const OVERDRAW_PX: f32 = 320.0;
 /// Show the scroll-to-bottom button beyond this distance from the end.
 pub const SCROLL_BUTTON_THRESHOLD_PX: f32 = 320.0;
+/// Bound session-local viewport memory independently of total chat history.
+const MAX_SAVED_VIEWPORTS: usize = 256;
 /// Text-selection edge scrolling runs only during a drag. A 24 ms cadence is
 /// smooth enough to track text while avoiding a permanent animation-frame loop
 /// on low-end devices.
@@ -1464,8 +1466,9 @@ struct FoldState {
 /// hard-tracks the pad's stale bottom on every commit — rig-traced). Wheel
 /// input releases the hold, leaving the reservation as plain scrollable
 /// space. The anchor retires once the reply overflows the reservation (pad
-/// ~0, height-neutral) and on explicit navigation / chat switches (revisits
-/// start at the bottom).
+/// ~0, height-neutral). Chat switches snapshot its runway with the viewport
+/// and restore it released, so revisiting never resumes hidden auto-follow.
+#[derive(Clone, Debug)]
 struct OwnTurnAnchor {
     chat_id: String,
     message_id: SharedString,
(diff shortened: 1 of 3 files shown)


crates/ui/src/transcript.rs
```diff
@@ -5066,6 +5541,263 @@ mod tests {
...
+    fn viewport_row(id: &str, entry_id: &str) -> Row {
+        Row {
+            id: id.into(),
+            version: 0,
+            turn_start: true,
+            kind: RowKind::ErrorChip {
+                message: SharedString::default(),
+            },
+            entry_id: entry_id.into(),
+            timestamp: None,
```
wingleeio, line 5688: PR #200 added the required `Row::copy_text` field on current `main`, so this new initializer now fails with E0063 and blocks `cargo test -p zeron-ui --lib`. Please update the helper with `copy_text: None,` (and refresh the branch against current `main`). I verified that exact addition makes the merged tree compile and all 504 UI library tests pass.

wingleeio (changes requested):
The viewport restoration logic itself looks coherent, but the current-main merge result does not compile its UI test target. I reproduced this at the exact reviewed head against main 9c8e213 with `cargo test -p zeron-ui --lib`. Adding the missing neutral field to the new test helper makes all 504 UI library tests pass.

jsgrrchg:
Updated branch and updated and fixed the test.

wingleeio (approved):
Re-review complete against current `main` at `9c8e213`. The requested `Row::copy_text` initializer is present, and the upstream merge tree matches Git’s clean automatic merge.

Validated with focused `rustfmt --check`, `git diff --check`, `cargo test -p zeron-ui --lib` (504 passed), `cargo build -p zeron`, and a fresh headed two-chat viewport-switch scenario. The transcript content pane before switching and after returning had zero differing pixels.

Before switching away from the scrolled chat:

Scrolled viewport before chat switch

After switching back, with the same viewport restored:

Restored viewport after chat switch

PR #208, opened 2026-08-23
Keyboard shortcuts archive the current session and jump to one by its sidebar place

tobi404:
## The problem

Two everyday moves have no keyboard path. Archiving the open session means
reaching for its row and using the context menu, and there is no way at all to
switch sessions from the keyboard: every switch is a click in the sidebar, even
when the row you want is the second one down.

t3code answers the second one with thread jumps, and that is the shape used
here.

## The fix

**Archive the current session.** `Mod+Shift+A` archives the open session. It
routes through the same call the context menu makes, so both paths leave the
archived session on screen and only drop its row out of the active list. The
shortcut archives and never unarchives: on an already archived session, or with
nothing open, it does nothing. `Mod+A` is the composer's Select all, so this
takes the shifted combo.

**Jump to a session by its place in the list.** `Mod+1` through `Mod+9` open the
sidebar row at that position. Holding the modifier draws each reachable row's
combo as a key-cap chip, so the numbering is visible rather than memorised.

The rows, the chips and the jumps all read one new `AppState::sidebar_chats`,
so no numbering can drift from the list on screen. A slot past the end of a
short list opens nothing, and only the active list is counted: the archived
shelf has no places.

The hints need an EXACT modifier match, so `Mod+Shift+4` never flashes the
overlay, and they read the live keymap so a rebind moves the trigger with it.
Window deactivation clears them, since an app switch swallows the key-up.

A chip takes the row's whole corner, above the status word and the hover pill.
The corner carries a time only on idle rows, so replacing just the time would
leave holes on exactly the busy rows worth jumping to.

## Scope

All nine slots and the archive shortcut are rebindable in Settings, next to the
existing five, and inherit their conflict detection and per-row Reset. The
table grows from 5 rows to 14; that page already scrolls.

Combos stay platform-neutral in the settings file, as `"mod-1"` rather than
`"cmd-1"`. `platform_combo` binds Cmd on macOS and Ctrl elsewhere, and
`display_combo` prints `Cmd+1` or `Ctrl+1` to match, so the same settings file
works across platforms. This only reuses the convention the existing five
shortcuts already follow.

The nine jump combos are held as one `jumpSession` list rather than nine
fields. A short, long or absent list heals on load. A fixed-length array would
instead fail the parse, and `UiSettings::load` discards the whole file on a
parse error, taking every unrelated setting with it.

One `JumpSession(slot)` action carries the slot, rather than nine
near-identical action types.

**Why the composer commit rides along.** The question panel claimed every 1-9
key press, modifiers included. Without that fix `Mod+1` fires twice: once as
the jump and once as an option selection. Only a bare digit selects now.

## Verification

- `cargo test -p zeron-ui` - 513 passed, 0 failed. New tests cover slot
  resolution against the drawn list and under a project filter
  (`jump_slots_count_the_rows_the_sidebar_draws`), the archive target rules
  (`archive_shortcut_only_targets_an_open_active_chat`), the storage heal
  (`short_or_long_jump_lists_heal_to_the_slot_count`,
  `a_malformed_jump_list_heals_without_losing_other_settings`), the hint
  predicate (`jump_hints_need_an_exact_modifier_match`), and that all 14
  defaults parse on this platform and none collide
  (`every_shortcut_default_is_unique_and_bindable`).
- `cargo clippy -p zeron-ui --all-targets` - no new warnings; the remaining
  ones are identical to `main`'s, only line-shifted.
- `rustfmt --check` - no formatting changes to the touched code. The two
  pre-existing diffs in `shell/spaces.rs` and the one in `shell.rs` are present
  on `main` unchanged and are deliberately left alone.
- Exercised by hand in the app built from this branch, via
  `scripts/dev-demo.sh`.

## Screenshots

Settings -> Shortcuts, in a build of this branch, captured from
`scripts/dev-demo.sh`. **Archive session** on `Cmd+Shift+A` sits below the four
existing rows, and **Jump to session 1** onward follow on `Cmd+1` and up. Every
one of them records and resets like the originals.

Settings -> Shortcuts showing Archive session on Cmd+Shift+A and the jump slots on Cmd+1 and up

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/208"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790057550&installation_model_id=425430&pr_number=208&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F208&signature=0b524f7998fced3d3a01f0b258fa3aac4a9f90216b588769f9b54e7f172ada48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (approved):
Reviewed the diff, ran the suite, and exercised the build headed. Approving — the design and the tests are solid. One required chore and two small notes below.

## Verified

- `cargo test -p zeron-ui` on the branch: 513 passed, 0 failed, matching the PR.
- The branch is based on v0.2.23 and now **conflicts with `main`** — #125 (Ctrl+Tab session cycling) landed in the same four files. I resolved the 15 hunks locally as a straight union (both features' enum variants, keymap fields, defaults, bindings, and tests side by side) to check the integration: the merged tree builds, **526 tests pass**, clippy shows only `main`'s pre-existing warnings, and `every_shortcut_default_is_unique_and_bindable` holds across all 16 shortcuts (`mod-1..9` / `mod-shift-a` don't collide with `ctrl-tab`/`mod-tab`).
- Headed smoke test of the merged build (Linux, sway rig): holding Ctrl draws the `Ctrl+1..5` key-cap chips in the row corners and releasing clears them; `Ctrl+2` opens row 2; `Ctrl+Shift+A` drops the open session to the Archived shelf while it stays on screen; a second press and `Ctrl+9` past the end are both no-ops; after the archive, `Ctrl+2` opens the renumbered row 2; #125's `Ctrl+Tab` still cycles correctly from the now-archived selection.

## Required

- **Rebase onto `main`.** The union resolution is mechanical. One spot needs a decision: in `render_active_rows`, `main` now filters via `tabs::in_space_filter` while this branch moves the same predicate into `AppState::sidebar_chats`. They're semantically identical today, but two copies can drift — after rebasing, consider pointing `cycle_session` (tabs.rs) at `sidebar_chats` too, so cycling, jumping, and the drawn list all read one function.

## Notes (non-blocking)

- Post-#125, macOS stores recorded Ctrl-chords as literal `ctrl-…`. `combo_modifiers` only parses `mod`/`alt`/`shift`, so a jump slot rebound to a Ctrl-chord on macOS binds and jumps fine but never shows its hint chip (and `on_modifiers_changed` doesn't track ctrl separately from `platform` there). Worth a follow-up `ctrl` leg in both if macOS rebinds matter.
- `docs/PARITY.md` §1.4 got a line for the cycling shortcuts; the jump slots and archive shortcut deserve the same one-liner on rebase.

wingleeio:
Rebased this onto `main` (maintainer edit) and pushed — the branch now carries the three original commits re-applied over #125's session cycling, plus three follow-ups:

- **`27c499f` Cycling reads the one sidebar-list function.** The rebase left the sidebar's space-filter predicate in two places: #125's `tabs::in_space_filter` (used by `cycle_session`) and this PR's `AppState::sidebar_chats`. Cycling now reads `sidebar_chats` too, so the drawn rows, the jump slots and the cycle order all come from one function and can't drift.
- **`15d9267` Shortcuts page groups its rows.** With the jump slots the table had grown to 14 flat rows. It now renders three labeled cards — **Panels**, **Sessions**, **Jump to session** — using the label-over-card pattern the Appearance page established. `group` is a total match over `ShortcutId` and `every_shortcut_lands_in_a_rendered_group` keeps the two halves honest, so a future shortcut must pick a group and lands on the page by construction.
- **`849c0f5`** PARITY §1.4 records the jump and archive shortcuts.

Conflict resolution was the straight union from the review: both sides' enum variants, keymap fields, defaults, bindings and tests side by side; the two overlapping upgrade-path tests merged into one (`a_keymap_missing_newer_shortcuts_keeps_its_customizations`, asserting both `NextSession` and `ArchiveSession` defaults).

Verified on the rebased branch: `cargo test -p zeron-ui` — **526 passed, 0 failed**; clippy and rustfmt unchanged from `main`; and the flows below exercised by hand in the built app.

**Settings → Shortcuts, grouped:**

Shortcuts page grouped into Panels, Sessions, and Jump to session cards

**Holding the modifier draws the key-cap chips (archived shelf stays unnumbered):**

Sidebar rows wearing Ctrl+1 through Ctrl+4 key-cap chips while Ctrl is held

**Ctrl+2 jump, then Ctrl+Shift+A — the session drops to the shelf but stays on screen:**

The archived session in the Archived shelf, still selected and on screen

wingleeio:
Two more commits, per review feedback:

- **`acce630` The jump hint wears the app's badge.** The bespoke key-cap chip is gone; the corner now renders `widgets::badge` — the same pill the Accounts plan badge uses — replacing the status/time-ago slot while the modifier is held.

  Sidebar rows wearing Ctrl+1..3 as standard badges in the status corner

- **`675c0fa` Session-nav shortcuts go quiet under an open popover.** The model selector quick-picks its rows on primary-modifier digits — the same chords the jumps claim — and gpui delivers a matched binding *and* the popover's own `on_key_down` for one press, so `Ctrl+2` with the picker open switched sessions underneath it. A new `Shell::overlay_owns_keyboard` (add-space palette or any composer picker popover, via `Pickers::is_open`) now gates jump, archive **and** #125's cycling — the palette guard cycling already had, widened and shared. The hint chips obey the same gate, re-checked at render so they drop the frame a popover opens.

  Verified in the built app: with the model selector open, holding Ctrl draws no chips and `Ctrl+2` leaves the session and the popover exactly where they were; Esc, then `Ctrl+1` jumps again.

  Model selector open, Ctrl+2 pressed — session and popover untouched

`cargo test -p zeron-ui`: 526 passed, 0 failed.

wingleeio:
**`10d1889`** restyles the jump hint once more, per review: the corner now renders `popover::kbd_hint` — the borderless translucent chip the model picker's ⌘N hints already use, at the sidebar PR badge's scale (no border, `ink(0.05)` fill, 10px mono muted text) — still replacing the status/time-ago slot while the modifier is held.

The label follows suit: a new `badge_combo` spells macOS combos as their key glyphs in canonical ⌃⌥⇧⌘ order with no separators — **⌘1** rather than **Cmd+1** — matching those picker chips, while Linux/Windows keep the textual `Ctrl+1`. A ctrl-rebound slot on macOS reads ⌃, so rebinds stay truthful. Covered by `badge_combos_use_mac_glyphs_and_linux_text`; `cargo test -p zeron-ui` — 527 passed, 0 failed.

Captured on Linux (so the chips read `Ctrl+N`; on macOS they read `⌘N`):

Sidebar rows with the borderless keyboard-hint badges in the status corner

Zoom of the sidebar rows showing the badge style

wingleeio:
**`4ce66a9`** sizes the jump hint to the sidebar PR badge exactly. `kbd_hint` was the picker-popover scale — its height floats off `py(1)`, landing taller than the row's PR badge, in normal-weight text. The chip is now cut from `pull_request_badge`'s Sidebar values: pinned **16px** tall, `px(4)`, `rounded(4)`, the same 0.08-fill / 0.85-text opacity formula on a neutral tone, and the label in the badge's mono at 10px **medium** — the same font treatment as the PR number. One badge system on the row.

The macOS ⌘-glyph label from `10d1889` is unchanged (chips read `⌘1`; Linux shown below with `Ctrl+N`).

Sidebar rows with the PR-badge-scale jump chips

Zoom of the resized chips

`cargo test -p zeron-ui`: 527 passed, 0 failed.

PR #221, opened 2026-08-24
Model picker: harness tabs, tab-scoped search, merged traits

wingleeio:
## Summary

- rebuild the model popover around harness tabs across the top (favorites star first), replacing the left icon rail
- scope search to the viewed tab: the query re-scopes when tabs switch instead of flattening every harness into one mixed list
- fold the reasoning ladder and model options into the same popover as a pinned tray, and merge the separate Traits chip into the model chip — one button, one card
- compact single-line rows on harness tabs, and menu-shaped skeletons (varied-width ghost labels) while the catalog or a tab's models load

## One trigger, one card

The composer previously carried two chips — the model chip and a Traits chip whose narrow-composer overflow needed its own width-measuring machinery. They merge into a single combined trigger: brand mark + model name, with the effective traits summary ("High · 200K", "Agent · Balance") as the chip's muted second tone, brightening only when something departs from its default.

Combined chip: brand icon, model name, and the muted effective traits summary

The popover stacks harness tabs, the scoped search, the tab's model list, and the pinned traits tray:

| Claude tab with reasoning tray | Loading (cold catalog) |
| --- | --- |
| Popover with harness tabs on top, compact model rows, and the pinned reasoning ladder | Tabs and search stay put while varied-width skeleton rows pulse below |

## Search stays inside the viewed tab

The old search spanned every harness and hid the rail while typing. Now the tabs never hide: the same query re-scopes as you switch tabs, and each row no longer needs a harness subline — on a harness tab the rows are single-line (provider attribution rides inline, which keeps opencode's 64-providers-of-the-same-model rows distinguishable). Selecting a foreign-harness row from favorites still switches harness first, exactly like clicking its tab.

| "opus" on the Claude tab | Same query, Cursor tab |
| --- | --- |
| Query opus scoped to Claude models | The same query re-scoped to Cursor's catalog, with its Optimize for options tray |

The right shot also shows the tray adapting per selection: Cursor renders its "Optimize for" model options where Claude/Codex render their reasoning ladders (each with its Default badge; picks in the tray keep the menu open for multi-adjust, a model pick still closes it).

| Favorites tab | opencode attributions |
| --- | --- |
| Favorites tab mixing starred models from different harnesses with brand sublines | opencode tab with provider attribution inline on compact rows |

The favorites tab keeps the two-line rows with brand sublines — it's the one list that mixes harnesses — and its search ranks the starred set only.

## Polish (from review)

Even 6px insets around the list band, tray rows restyled locally to the model rows' scale (the shared `menu_row` geometry is untouched for other popovers), highlight radius softened 8 → 6, and the divider after the favorites tab removed — one even run of tabs. Screenshots above show the final styling.

## Removed

- `PickerKind::Traits`, the Traits chip, and the `compact_traits` width machinery (`Composer::set_available_width` no longer feeds the pickers). `ZERON_OPEN_PICKER=traits` still boots the merged popover so existing rigs keep working.
- `skeleton_rows` stays for other surfaces; the picker uses a new `skeleton_menu_rows` (ghost-label widths on a deterministic ladder — no randomness, so reopen paints identically).

## Behavior kept

- ⌘1–⌘9 row jumps, Enter/arrows keyboard nav, the floating scrollbar, stars-first ordering inside a harness tab, virtualized lists for 7k-model catalogs, per-open catalog revalidation, and the locked-harness treatment on sessions (other tabs dim).
- Sticky defaults, favorites persistence, and the no-agents empty state are untouched.

## Verification

- `cargo test -p zeron-ui`: 553 passed, 0 failed — includes new unit coverage for the scoped flatten (`scoped_model_rows`): tab search never leaves the viewed harness, favorites search ranks only starred rows, stars float first, and description/provider text stays searchable inside a tab.
- Exercised in the headless rig against the five real installed harnesses (screenshots above): tab switching, scoped search re-scoping live, favorites round-trip, skeletons on a cold daemon, and the combined chip label tracking model + effective traits.

### DIFF
--- a/crates/ui/src/composer.rs
+++ b/crates/ui/src/composer.rs
@@ -3466,14 +3466,9 @@ impl Composer {
     }
 
     /// Feed the stable conversation-column width into responsive composer
-    /// controls. The text input's own width is unsuitable here because it
-    /// changes when the Traits label is replaced by the overflow dots.
+    /// controls.
     pub fn set_available_width(&mut self, width: f32, cx: &mut Context<Self>) {
         let composer_width = width.clamp(0.0, COMPOSER_MAX_WIDTH);
-        let inner_width = (composer_width - 2.0 * Theme::SPACE_LG).max(0.0);
-        self.pickers.update(cx, |pickers, cx| {
-            pickers.set_composer_width(inner_width, cx);
-        });
         if composer_width_changed(self.last_available_width, composer_width) {
             self.last_available_width = Some(composer_width);
             // The shell renders before this child, so this queues one more
--- a/crates/ui/src/pickers.rs
+++ b/crates/ui/src/pickers.rs
@@ -2954,8 +2954,8 @@ impl Pickers {
         }
         let rows = self.model_rows(cx);
 
-        // ── tabs: the favorites star, a thin divider, one brand icon per
-        //    harness — ACROSS THE TOP (user request; was a left rail). The
+        // ── tabs: the favorites star, then one brand icon per harness —
+        //    ACROSS THE TOP (user request; was a left rail). The
         //    viewed tab wears a 2px accent bar sitting on the row's bottom
         //    hairline. Tabs never hide: a live search only filters the
         //    viewed tab's list, so switching tabs re-scopes the same query.
@@ -3003,14 +3003,6 @@ impl Pickers {
                         }),
                 )
                 .when(favorites_view, |el| el.child(tab_indicator(theme.accent))),
-        );
-        tabs = tabs.child(
-            div()
-                .flex_none()
-                .w(px(1.0))
-                .h(px(18.0))
-                .mx(px(2.0))
-                .bg(crate::theme::hairline(0.10)),
         );
         for (ix, descriptor) in descriptors.iter().enumerate() {
             let harness = descriptor.id;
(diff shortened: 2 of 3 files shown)


wingleeio:
Polish from review feedback (pushed): the list band's vertical inset now matches its 6px horizontal inset (compact rows +1px vertical padding), the traits tray rows are restyled locally to the model rows' scale (12.5px text, same padding — `menu_row`'s roomier app-wide geometry stays untouched for other popovers), and highlight/selected radius softened 8 → 6 across the card.

Restyled popover: even insets, tray rows matching model row scale, softer highlight radius

wingleeio:
Chip loading states (pushed): the chip never collapses to a bare icon anymore. While the selected model resolves, the label slot shows a pulsing ghost bar; while the harness itself is still unknown (catalog loading, nothing remembered), the icon slot wears the pixel-glyph loader instead of guessing a brand mark. Staged deterministically via a new dev knob `ZERON_SLOW_CATALOG_MS=<ms>` (these states are sub-second against a warm local daemon).

| Model resolving | Harness resolving |
| --- | --- |
| Chip with harness icon and pulsing ghost label | Chip with pixel-glyph loader and ghost label |

PR #251, opened 2026-09-04
feat(landing): add PostHog traffic and download tracking

SinaKhalili:
## Summary
- Capture `$pageview` events and `download_clicked` events with CTA placement and the current release version through the PostHog US ingestion API.
- Use a small first-party tracker with no SDK dependency, cookies, browser storage, autocapture, or session replay. Respect DNT/GPC and exclude URL queries, fragments, and referrer paths.
- Restrict tracking to production landing-page URLs and keep download navigation independent of telemetry delivery.
- Add 27 offline tests and run them before landing-page deployment in CI. The browser uses the public project token; `.env` is not committed.

## Measurement scope
- Download events represent button clicks, not completed transfers or installs.
- IDs reset on each page load, so unique/returning-visitor counts are not reliable.
- This does not certify consent compliance. Hosting region alone does not determine consent requirements.

#### Test plan
- [x] `node --test apps/landing/telemetry.test.mjs` — 27 tests pass with mocked requests; no live events sent.
- [x] `node --check apps/landing/public/telemetry.js`
- [x] `git diff --check`
- [x] Verify the US ingestion endpoint accepts a CORS OPTIONS preflight without sending a project token or event.
- [ ] After deployment, confirm pageview and download-click events appear in PostHog and downloads still work with tracking blocked.

Generated with [Devin](https://devin.ai)

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/251"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791150141&installation_model_id=425430&pr_number=251&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F251&signature=93cadadaa4e152efee3d0060808ea333dd3f6d12eb3554bfbd72ba75a48b1ce4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -93,6 +93,10 @@ jobs:
         if: steps.changed.outputs.changed == 'true'
         with:
           node-version: 22
+      - name: test
+        if: steps.changed.outputs.changed == 'true'
+        working-directory: apps/landing
+        run: node --test telemetry.test.mjs
       - name: deploy
         if: steps.changed.outputs.changed == 'true' && env.CLOUDFLARE_API_TOKEN != ''
         working-directory: apps/landing
--- a/apps/landing/public/index.html
+++ b/apps/landing/public/index.html
@@ -9,6 +9,7 @@
 <meta property="og:description" content="Control your coding agents from any of your devices." />
 <meta property="og:image" content="/assets/app-screenshot.jpg" />
 <link rel="icon" href="/assets/zeron-favicon-v3.png" type="image/png" sizes="128x128" />
+<script defer src="/telemetry.js"></script>
 <style>
   @font-face {
     font-family: "Geist";
--- a/apps/landing/public/telemetry.js
+++ b/apps/landing/public/telemetry.js
@@ -8,8 +8,7 @@
   ) return;
 
   const distinctId = crypto.randomUUID();
-  const sessionTimestamp = Date.now().toString(16).padStart(12, "0");
-  const sessionId = `${sessionTimestamp.slice(0, 8)}-${sessionTimestamp.slice(8)}-7${crypto.randomUUID().slice(15)}`;
+  let sessionId, sessionStartedAt;
   let referrer = "$direct", referringDomain = "$direct";
   try {
     const url = new URL(document.referrer);
@@ -22,6 +21,12 @@
   const capture = (event, properties = {}) => {
     if (optedOut()) return;
     try {
+      const now = Date.now();
+      if (!sessionId || now < sessionStartedAt || now - sessionStartedAt >= 24 * 60 * 60 * 1000) {
+        const sessionTimestamp = now.toString(16).padStart(12, "0");
+        sessionId = `${sessionTimestamp.slice(0, 8)}-${sessionTimestamp.slice(8)}-7${crypto.randomUUID().slice(15)}`;
+        sessionStartedAt = now;
+      }
       fetch("https://us.i.posthog.com/i/v0/e/?ip=0", {
         method: "POST",
         headers: { "Content-Type": "application/json" },
@@ -31,7 +36,7 @@
         body: JSON.stringify({
           api_key: "phc_yAQTUdUM9vQMcHpGmMcNJom6vfrN3r3eo5avDSnt8S9R",
           event,
-          timestamp: new Date().toISOString(),
+          timestamp: new Date(now).toISOString(),
           properties: {
             distinct_id: distinctId,
             $process_person_profile: false,
--- a/apps/landing/public/telemetry.js
+++ b/apps/landing/public/telemetry.js
@@ -1,0 +1,73 @@
+(() => {
+  const optedOut = () => navigator.globalPrivacyControl === true || ["1", "yes"].includes(navigator.doNotTrack);
+  if (
+    !["https://zeron.sh", "https://comet.zeron.sh"].includes(location.origin) ||
+    !["/", "/index.html"].includes(location.pathname) ||
+    optedOut() || typeof fetch !== "function" ||
+    typeof crypto === "undefined" || typeof crypto.randomUUID !== "function"
+  ) return;
+
+  const distinctId = crypto.randomUUID();
+  const sessionId = crypto.randomUUID();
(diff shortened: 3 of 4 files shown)


wingleeio:
Reviewed head `7272a9018799f030509ff13273eb20a81222ffef`. Two actionable findings:

### 1. [P1] Failing tests block landing-page deployment

In `apps/landing/telemetry.test.mjs`, lines 63 and 159 still read `payload.distinct_id`, but the tracker now sends the identifier as `payload.properties.distinct_id`. Running `node --test apps/landing/telemetry.test.mjs` on Node 22 reports **25 passed, 2 failed**. The new required test step in `.github/workflows/deploy.yml` therefore prevents landing-page deployment after merge.

Update all identifier assertions to use `payload.properties.distinct_id`, including line 98: the click-identity assertion currently passes by comparing `undefined` with `undefined`.

### 2. [P2] UUIDv4 session IDs exclude events from PostHog session aggregations

In `apps/landing/public/telemetry.js:11`, `crypto.randomUUID()` generates a UUIDv4 for `$session_id`. [PostHog requires custom session IDs to be UUIDv7](https://posthog.com/docs/data/sessions#custom-session-ids) and explicitly excludes events with nonconforming IDs from session aggregations. Raw events can still be recorded, but session-based analytics will not work as intended.

Generate a compliant UUIDv7 for `$session_id` and add a format assertion. The `distinct_id` can remain UUIDv4.

### Validation

- Ran the 27-test suite: 25 passed, 2 failed.
- Reproduced the UUID-version mismatch by importing the actual tracker with mocked browser globals and transport.
- `node --check apps/landing/public/telemetry.js` and `git diff --check` passed.
- No live telemetry was sent and no source files were edited.

wingleeio:
Clarification after checking the current PostHog docs and ingestion implementation:

**Sina is right about top-level `distinct_id`.** The [single-event API documentation](https://posthog.com/docs/api/capture#single-event) explicitly puts `distinct_id` at the top level, alongside `api_key`, `event`, and `properties`. [PostHog’s ingestion implementation](https://github.com/PostHog/posthog/blob/24a784e198fbbe76dcc2a5d1cc03c4a28f2dae68/rust/common/types/src/event.rs#L463-L482) accepts both top-level and nested `distinct_id`.

The earlier automated review claiming that top-level IDs cause every event to be rejected was incorrect. Moving the ID into `properties` was unnecessary and introduced the test failures observed at `7272a90`. Reverting that move is a valid fix—not just updating the assertions as my previous comment suggested. I should have made that clear.

**The separate UUIDv4 session-ID finding still stands.** The [custom session ID documentation](https://posthog.com/docs/data/sessions#custom-session-ids) explicitly requires IDs to be a valid UUIDv7 and says events with nonconforming session IDs are not included in session aggregations. [`crypto.randomUUID()` produces UUIDv4](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), so it is fine for `distinct_id`, but not a compliant `$session_id`. This affects session aggregations, not raw event ingestion.

Bottom line: the original top-level `distinct_id` is valid and can be restored; `$session_id` needs a separate fix for session analytics.

SinaKhalili:
@wingleeio fixed

wingleeio:
Re-reviewed `e263bdf911bdc86fbd0dd7d8ba33fbc9a6ddbc90`. The failing identifier assertions and UUIDv4 session-format issue are fixed. **29/29 tests pass**, along with syntax and whitespace checks.

### Remaining finding: [P2] Rotate the session ID after 24 hours

In `apps/landing/public/telemetry.js:11-12`, the UUIDv7 is generated once and reused for the entire page lifetime. If someone leaves the landing page open for 24 hours or longer and then clicks Download, that event retains an expired session ID.

[PostHog’s custom session ID requirements](https://posthog.com/docs/data/sessions#custom-session-ids) say the UUIDv7 timestamp plus 24 hours must be after the timestamp of the last event in the session. Events outside that window are excluded from session aggregations.

I reproduced this using the actual tracker with mocked transport and an advancing clock:
- 24 hours minus 1 ms: within the permitted window.
- 24 hours and later: outside the permitted window; the same session ID is still sent.

Generate a fresh session ID before capturing an event once the existing session reaches 24 hours. Add an advancing-clock test—the new tests freeze time, so they miss this case.

This affects session analytics, not download navigation or raw event ingestion. No other actionable findings in this pass, and no live telemetry was sent.

SinaKhalili:
Addressed the remaining 24-hour session lifetime finding in 154e932.

- Rotate the UUIDv7 before capturing an event when the session reaches 24 hours, or when the clock moves backward before the session start.
- Use the same clock reading for rotation and the event timestamp.
- Preserve the visitor ID and reuse the new session ID for subsequent events within its valid window.
- Added advancing-clock coverage for 24 hours minus 1 ms, exactly 24 hours, 24 hours plus 1 ms, a multi-day gap, clock rollback, and repeated rotation.

Validation: all 34 offline tests pass; JavaScript syntax and git whitespace checks pass. No live telemetry was sent.

SinaKhalili:
@wingleeio  updated

wingleeio (approved):
Reviewed 154e932c1fe366f36d1e6d4bdc44389884248df4. All previously raised findings are resolved, and no new actionable issues were found. All 34 offline tests pass, along with JavaScript syntax and whitespace checks. Independently verified session rotation at the 24-hour boundary, after multi-day gaps, and following clock rollback, while preserving the visitor ID. Live PostHog ingestion remains the planned post-deployment check; no live telemetry was sent during review.

PR #277, opened 2026-09-08
Add shared message queues with protected editing on desktop and iOS

jsgrrchg:
## Summary

Messages submitted while an agent is busy now have a shared, visible queue that can be reordered, edited, removed, or explicitly delivered from desktop and iOS. Pending messages stay out of the transcript until the chat host actually dispatches them, and edits are protected against concurrent delivery.

For example, a follow-up submitted during a turn can remain above the composer until that turn finishes. Editing it reserves its position and blocks delivery; saving replaces the queued content in place. Cancelling the active run freezes the remaining queue instead of immediately starting the next message.

## Shared document and protocol

- Add a per-session `LoroMovableList` for queued messages, separate from the immutable command ledger and host-authored transcript. Rows preserve stable IDs, text, committed attachment paths, author/timestamps, hold-until-turn-end policy, and delivery gates.
- Support enqueue, update, reorder, removal, and whole-list queue watches. Native move operations allow concurrent reorders to converge without delete-and-insert duplication.
- Add RPC methods for queue observation and mutation, explicit steering and send-now, and edit acquisition, renewal, and completion.
- Route delivery, removal, and protected editing to the chat's owning host, including through the device relay. Ordinary enqueue/reorder operations remain synchronized document writes.
- Advertise queue support through `EngineInfo` and device-registry capabilities: `message-queue-v1`, `message-queue-actions-v1`, `message-queue-attachments-v1`, `message-queue-clean-attachment-text-v1`, and `message-queue-edit-lease-v1`. Missing capabilities default to unsupported; clients retain the existing command-send path for older hosts.

## Host execution and recovery

- Drain queues on document changes and session status changes. Serialize draining, explicit delivery, removal, and protected editing with a per-chat mutex.
- Treat both `Working` and `AwaitingInput` as an active turn. Providers that accept mid-turn input can receive text through steering; other providers hold messages until the turn ends.
- Respect each row's hold policy. Attachment-bearing messages wait for a new turn that can accept files, and an editing/review gate blocks the head without skipping it.
- Offer non-interrupting steering and an explicit send-now path that interrupts before dispatching the selected message. Preserve the row when dispatch fails and prevent two clients from consuming the same row.
- Preserve the queue ID when promoting a message into the transcript. Rebuild attachment transport text only at dispatch and recognize legacy expanded attachment trailers to avoid duplication.
- Freeze automatic delivery after an explicit Cancel and when opening a document with recovered queued work. A successful explicit prompt or queue delivery resumes it. Freeze open queues before shutdown stops live runs.
- Keep chats discoverable by unarchiving them when new queued work is submitted.

## Protected editing

- Acquire a host-authoritative edit lease before loading a queued message into the composer. Persist the delivery gate before acknowledging acquisition so a host crash cannot make an acknowledged edit sendable again.
- Use 60-second leases with 20-second client renewal. Expiry transitions to `ReviewRequired`, never automatic release; restored documents rearm each outstanding lease's expiry independently.
- Validate lease generations and the original text hash when committing. Support commit, cancel, discard, and release-unchanged outcomes without letting a stale response resolve a newer edit.
- Commit text, optional attachment replacements, and gate release in one document operation. Desktop retains the edit buffer until acknowledgement and restores the previous composer draft and attachments afterward.
- Keep removals pending and inert until the host acknowledges them. Failed or unconfirmed mutations refresh the authoritative queue projection instead of leaving a misleading optimistic state.

## Desktop experience

- Add a queue tray above the composer with drag reordering, attachment indicators, editing/review/pending states, edit/remove controls, and a trailing delivery action.
- Use a consistent **Steer** label with tooltips explaining whether the selected action steers without interruption or sends immediately by interrupting. Availability follows resolved provider metadata and host capabilities.
- Edit queued text and attachments directly in the composer while the protected row retains its position. Escape cancels the edit after contextual completion handling; an entirely empty saved edit discards the row.
- Add persisted Enter versus Cmd/Ctrl+Enter submission settings and a Queue/Steer preference for active-turn submissions. Desktop preserves automatic steering as its default when supported.
- Cmd/Ctrl+Enter submits content, saves an active queue edit, or activates the queue head when the composer is empty. It never skips a blocked head or turns an empty modified submit into Stop. Holding the primary modifier reveals the head's shortcut hint.
- Reserve the modified-submit shortcut and migrate conflicting saved bindings while preserving unrelated customizations. Update shared input consumers, initialization, and the resource-profile example for the new composer event/configuration.
- Avoid optimistic transcript bubbles for queued submissions. Preserve the active reply's scroll runway and anchor a locally queued message only when its stable ID materializes in the transcript.

## iOS experience

- Project the shared queue alongside session entries and add a queue panel with drag/arrow reordering, attachment indicators, edit/remove controls, and provider-aware **Steer** or **Send now** actions.
- Add protected composer editing with lease renewal, actionable failure states, acknowledgement-based removal, and sync repair after failed actions. Discarding an actively edited row does not release its lease first.
- Add a persisted Queue/Steer toolbar preference, preserving Queue as the phone's default.
- Support Command+Return on external keyboards for draft submission, saving an active edit, or activating the first queued row. Attachment-only drafts count as content.
- Decode device capabilities and provider steering metadata, keep attachment transport details out of editable text, and update benchmark helpers for the combined transcript/queue projection.

## Documentation and validation

Add a message-queue reference covering ownership, delivery, recovery, edit leases, compatibility, and test coverage, and document the mobile workflow in the iOS README. Align the existing RPC transcript-reset test with the current `contextUsage` field.

Validation completed on Linux:

- `cargo check --workspace --all-targets` passed.
- Protocol, document, and engine unit tests plus the dedicated message-queue integration suite: **311 passed**.
- Device-routing integration suite: **7 passed**, including a local/remote race to consume the same queued row.
- Existing end-to-end, queued-attachment, relay-delivery, unarchive-on-send, and workspace-sync suites: **28 passed**, with **2 live-service tests ignored**.
- Desktop unit and headless UI suite: **767 passed with one test thread**, matching the repository CI's test-thread setting. An initial parallel run failed `selecting_live_tail_keeps_anchor_visible_during_a_stream_burst`; that test passed in isolation and in the complete serial run.

**1,113 Rust tests passed in total.** iOS queue tests are included, but the iOS app and XCTest suite have not been built or run here because Xcode is unavailable on this Linux machine. Native macOS/iOS validation remains outstanding.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/277"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791479805&installation_model_id=425430&pr_number=277&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F277&signature=6a2c690418bf25b4b499b9edcaf3c627a9fcb492c50634e433b38f83fdd8cbe4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


crates/engine/src/doc_host.rs
```diff
@@ -2151,6 +2341,742 @@ impl DocHost {
...
+                return;
+            };
+            handle.publish_queue();
+            if let Err(err) = self.dispatch_queued(handle, &item, send).await {
+                tracing::warn!(chat = %handle.chat_id, error = %err, "queued send failed");
+                // Back to the head it came from: a failed send must not reorder
+                // the queue behind the user's back.
+                let _ = handle.doc.insert_queued(0, &item);
+                handle.publish_queue();
+                return;
```
wingleeio, line 2995: [P1] Stop failed dispatches from retriggering their own drain

Restoring the failed row commits another document change, which wakes `chat_task` and immediately calls `drain_queue` again. A persistent failure therefore continually takes and reinserts the same row with no backoff or external event. I reproduced this by calling `queue_message` on an open, claimable chat whose workspace row/config has not arrived: its operation count grew from 7,885 to 25,021 in 200 ms while dispatch kept returning “no live run and no prior run config.” Besides CPU/log churn, every retry permanently expands the synchronized CRDT history. Preserve the pending row without generating a self-triggering retry, or gate retries on a bounded timer or a relevant state change.

crates/engine/src/doc_host.rs
```diff
@@ -2151,6 +2341,742 @@ impl DocHost {
...
+            }
+        }
+        // Same reading of "busy" as the drain: a turn parked on a question is
+        // still a turn, and it has to be stopped before this one starts.
+        if send == QueueSend::Interrupt && sessions.turn_in_flight(chat_id) {
+            sessions.interrupt(chat_id).await?;
+        }
+        let request = sessions
+            .last_request(chat_id)
+            .or_else(|| self.request_from_chat_row(chat_id, &prompt));
```
wingleeio: [P2] Apply current chat configuration when starting a queued turn

`last_request` always wins once this process has dispatched a turn, so a later model/reasoning/options change in the chat row is ignored for queued follow-ups. I reproduced this with a turn-boundary harness: dispatch with `old-model`, update the chat config to `new-model`, enqueue a follow-up, then finish the first turn. The second request still contains `old-model`. This also affects explicit send-now, even though the composer displays the new selection. Rebuild the next-turn request from current chat configuration (retaining any necessary runtime-only fields), or preserve the intended configuration on the queue row.

crates/ui/src/queue.rs
```diff
@@ -0,0 +1,1506 @@
...
+        let editing = self.editing_queued.clone();
+        let mid_turn_steering = self.pickers().read(cx).resolved_mid_turn_steering(cx);
+
+        let list_chat = chat_id.clone();
+        let drop_chat = chat_id.clone();
+        let rows = div()
+            .flex()
+            .flex_col()
+            .gap(px(ROW_GAP))
+            .children(items.iter().enumerate().map(|(ix, item)| {
```
wingleeio: [P2] Bound the queue tray and make excess rows scrollable

All rows are rendered at a fixed height into an unbounded column. In the native app, enqueuing 24 held messages in a 1280×900 window pushes the composer and final rows below the window and removes the transcript from view. Wheel input over the tray does not reach the hidden controls because there is no scroll container. Smaller windows hit this with fewer rows. Cap the tray relative to available height and scroll its contents while keeping the composer visible.

[Reproduction screenshot](https://github.com/user-attachments/assets/4f50d917-2250-43da-b260-0b6e09de2139)

apps/ios/Zeron/Composer/ComposerView.swift
```diff
@@ -437,7 +525,90 @@ struct ComposerView: View {
...
+    /// Save the retyped row. Saving it empty is how a queued message is thrown
+    /// away — nothing to send is a decision, not an error.
+    @MainActor
+    private func commitQueueEdit() async {
+        guard let lease = queueEditLease, !queueEditBusy else { return }
+        let edited = text.trimmingCharacters(in: .whitespacesAndNewlines)
+        queueEditBusy = true
+        let result = edited.isEmpty
+            ? await store.finishQueuedEdit(lease, action: "discard")
+            : await store.finishQueuedEdit(lease, action: "commit", text: edited)
```
wingleeio: [P2] Save or disable attachments while editing a queued message

The queue-edit path leaves the composer's attachment picker and removal controls active, but saves only `text`; `SessionStore.finishQueuedEdit` never sends replacement attachment paths. Consequently, attaching a photo while editing and tapping Save acknowledges the edit without adding that photo to the queued message. The photo remains in the ordinary draft and can be sent with a later unrelated message. Clearing the text also discards the row even when a newly staged photo is visible. Isolate the original draft attachments, load the queued attachments, and upload/commit replacements with the edit, or disable attachment changes for text-only queue editing. This finding is from the iOS source path; it was not run in a simulator.

apps/ios/Zeron/Composer/ComposerView.swift
```diff
@@ -437,7 +525,90 @@ struct ComposerView: View {
...
+            queueEditLease = nil
+            text = stashedDraft ?? ""
+            stashedDraft = nil
+            uploadError = nil
+        case .conflict:
+            uploadError = "This message changed on another device; your edit is still here."
+        case .missing:
+            uploadError = "The queued message was removed; your edit is still here."
+        case .lost:
+            uploadError = "The edit lease changed; your text is still here."
```
wingleeio: [P2] Provide a local escape from a missing or superseded edit lease

When another device removes the row during an iOS edit, Save returns `.missing` but leaves `editingQueuedId` and `queueEditLease` set. Every subsequent Send still calls `commitQueueEdit` for that nonexistent row, while the queue row containing “Stop editing” has disappeared. A superseded lease similarly makes both Save and Cancel return `.lost` forever. The only recovery is navigation, whose `onDisappear` restores `stashedDraft` and drops the edited text. Preserve the edit as a normal recoverable draft and clear the terminal lease state, or expose an independent local cancel/reacquire action. This is a source-level iOS finding.

wingleeio (changes requested):
Request changes on `70ff28886008d827434f0dcf808096cec1c02fae`. The review found a self-triggering failed-delivery loop, stale model selection for queued turns, an overflowing desktop queue, and two iOS edit-state problems. Details and reproduction steps are in the five inline comments.

Validation on this PR revision:

- `cargo build --locked -p zeron` and `cargo check --workspace --all-targets --locked` passed.
- **1,113 existing Rust tests passed**, with two live-service tests ignored: 102 document, 25 protocol, 163 engine unit, 21 message-queue integration, 7 device-routing, 28 other relevant engine integration, and 767 desktop unit/headless UI tests. Tests ran with one test thread.
- Added two temporary regression checks in the isolated review checkout. Both fail on this revision: a permanently failing queued dispatch keeps growing the CRDT operation log without external input; a queued turn uses `old-model` after chat config changes to `new-model`.
- Exercised the actual Linux/X11 desktop app with an isolated mock harness. Opening a protected edit and saving it with Ctrl+Enter preserved the row ID, replaced its text, and released its gate. A 24-row queue pushed the composer and bottom rows outside a 1280×900 window; scrolling over the tray did not recover them.
- iOS findings are from source review. Xcode, iOS XCTest, and native macOS validation were unavailable here.

Screenshots below are captures of this PR's native app, uploaded as GitHub user attachments. The mock harness uses isolated local test data.

<details>
<summary>Normal queue above the composer</summary>

Native desktop with three queued messages

</details>

<details>
<summary>Protected editing in the composer</summary>

Native desktop with the first queued row reserved while its text is edited

</details>

<details open>
<summary>Reproduced failure: 24 rows push the composer off-screen</summary>

Native desktop queue overflow at 1280 by 900

</details>

wing-anara:
Corrected screenshots: the actual native iOS queue and queue-editing UI from the local feedback-fix build (iPhone 17 simulator, iOS 26.3).

| Three queued messages, including an attachment | Editing the attachment row in the composer |
| --- | --- |
| iOS queue with three pending messages and an attachment indicator | iOS queue editing with Editing below, attachment indicator, and independent Stop editing control |

Captured through XCTest using seeded offline queue data and a stubbed edit-lease response; these demonstrate native UI states, not live cross-device relay behavior. The capture check verifies the editing state, the independent stop control, and absence of the photo picker during editing. Temporary capture hooks were removed afterward. Local fixes remain unpushed.

crates/ui/src/composer.rs
```diff
@@ -1291,12 +1408,50 @@ pub fn init(cx: &mut App) {
         bindings.push(KeyBinding::new(&format!("{prefix}-x"), Cut, ctx));
         bindings.push(KeyBinding::new(&format!("{prefix}-v"), Paste, ctx));
     }
+    bindings
+}
+
+/// Bind the composer keymap. Call once at app boot.
+pub fn init(cx: &mut App, send_behavior: ComposerSendBehavior) {
```
wingleeio, line 1415: [P1] Update the browser callers when adding the send-behavior argument

The current merge result with main fails both CI jobs because `crates/ui/src/browser/mod.rs:522` and `crates/ui/examples/browser-fixture.rs:109` still call `composer::init(cx)`. Both report E0061 after this signature change. Update the test caller with a default behavior and the fixture caller with its configured behavior (or otherwise retain compatibility). I reproduced the UI test compilation failure on merge commit `98b5a664`; the macOS log confirms the second call site. The PR-head suite passes because those browser files are not in the head checkout yet.

wingleeio (changes requested):
Retested `25d4428d6e9e8663242aa1cf3527438f61342f05` with the real authenticated accounts in isolated fixture projects. **Still requesting changes:** the merge result fails both CI jobs, and OpenCode has an intermittent queued-turn completion failure.

The original dispatch hot-loop and stale-model regressions now pass. The desktop tray stays bounded with 24 messages, scrolls to its last row, reorders correctly while scrolled, and preserves the prior draft after cancelling an edit. At 900×600 the composer remains visible. The two original iOS findings are addressed in source: attachment controls are disabled during a queue edit, and a missing/lost lease has an independent copy-and-exit action. iOS runtime validation is still outstanding.

**Remaining findings**

1. **[P1] Update both browser callers of `composer::init`.** GitHub's merge commit `98b5a664` contains callers added on main that still pass one argument: `crates/ui/src/browser/mod.rs:522` and `crates/ui/examples/browser-fixture.rs:109`. The new signature requires `ComposerSendBehavior`. Both fail with E0061. I reproduced the UI-test compile failure locally against the merge commit. [Linux CI](https://github.com/zeronsh/comet/actions/runs/34323302752/job/102374668860) / [macOS CI](https://github.com/zeronsh/comet/actions/runs/34323302752/job/102374668764).

2. **[P1] Investigate OpenCode retiring a held prompt without an answer.** With Big Pickle, run an initial `sleep 8`, queue two held prompts, protect the head for editing, and release it after the initial turn ends. In chat `9995226b-7196-4314-aed0-c2b0d4b9ab84`, the two stable queued IDs entered the transcript 61 ms apart. Only `QUEUE_EDITED_OK` was answered; `QUEUE_FIRST_OK` disappeared from the queue, with the session idle and no answer after 150 seconds. The journal ends with `Steered → AssistantMessageCompleted → Done` and no content for the next segment. OpenCode's own persisted session also has the final user prompt without an assistant answer. Another run failed similarly; a subsequent repeat passed. This is an intermittent integration failure, not a claim that every OpenCode run fails. The exact race still needs isolation and a regression test.

OpenCode: empty queue and idle session, with the second queued prompt unanswered

**Validation completed**

- `cargo check --workspace --all-targets --locked` passed on the PR head.
- `cargo test --workspace --all-targets --locked -- --test-threads=1`: **1,522 passed, 0 failed, 13 ignored**, including the two review-only regressions. A filesystem-watcher failure under exhausted inotify instances disappeared after stopping the baseline test engine; the full suite was then rerun successfully.
- Four normally ignored integration tests passed against a real local edge worker: two-engine workspace convergence, device-room relay, registry churn, and two-client registry convergence.
- Edge unit/workerd tests: **47 passed**. Edge TypeScript check passed.
- macOS release application check passed in CI; the subsequent fixture compile failure prevented native browser/frame validation. No Xcode/iOS simulator is available locally.

| Real harness | Queue/edit/reorder/remove | Cancel + Send now | Queued image |
|---|---|---|---|
| Claude Code · Haiku 4.5 | Pass | Pass; mid-turn steering also passed | Pass |
| Codex · GPT-5.6 Luna | Pass | Pass; mid-turn steering also passed | Pass |
| Cursor · Composer 2.5 | Pass | Pass | Pass on retry¹ |
| Devin · GPT-5.6 Luna | Pass | Pass | Pass with Sonnet 4.6² |
| Grok · 4.6 | Pass | Pass | Pass |
| OpenCode · Big Pickle | Intermittent failure above; repeat passed | Pass on retry | Pass with MiMo V2.5 Free² |

¹ Cursor initially answered RED for the blue fixture, then BLUE on retry; upload bytes matched the fixture in both runs. ² Devin's initial Luna image response refused the task; Sonnet passed. OpenCode's catalog marks Big Pickle as not supporting images, so the image test used an image-capable model. Initial OpenCode startup timeouts were retried separately. Hermes and Pi are not installed/configured here and were not tested. These are harness checks, not coverage of every model offered by each provider.

All screenshots below are actual captures of the updated native Linux app, uploaded as GitHub user attachments.

<details>
<summary>Desktop queue: top, bottom, scrolled reorder, protected edit, and small window</summary>

24-message tray: top
24-message tray: bottom, composer visible
Scrolled drag moved row 24 ahead of row 22; verified through RPC
Protected editing in the composer
900×600: prior draft restored after cancelling the edit

</details>

<details>
<summary>Real-account screenshots for all six available harnesses</summary>

Claude Code queue lifecycle
Codex queue lifecycle
Cursor queued image
Devin queued image
Grok queue lifecycle
OpenCode queued image with an image-capable model

</details>

wingleeio:
Fixed both remaining findings in c5b25dfb1fc358f1987f66fe274931ee619944ce, after merging the current `main`.

- Updated both browser callers of `composer::init`; the fixture uses its configured send behavior.
- OpenCode now ignores trailing idle events from the previous turn until the queued turn starts, fails, or is explicitly aborted. A deterministic HTTP/SSE regression failed before the fix and passes afterward with both idle-event orderings. Early provider errors and cancellation are also covered.

Validation: **1,529 workspace tests passed**, 13 ignored; the app and browser fixture build. Real-account queue lifecycle checks passed with Claude Code, Codex, Cursor, Devin, Grok, and OpenCode. OpenCode passed three consecutive lifecycle runs, cancellation/Send now, and queued-image delivery. The isolated test processes have been stopped.

[Linux and macOS CI](https://github.com/zeronsh/comet/actions/runs/34326069385) both passed: 775 Linux UI tests, the macOS application check, native browser fixture/tests, and native frame recovery. iOS runtime coverage remains unavailable locally.

The updated app now answers both queued prompts in order:

OpenCode queue after the fix

wingleeio (approved):
The reported findings are addressed in `c5b25dfb1fc358f1987f66fe274931ee619944ce` (including the preceding fixes in `25d4428`). I implemented the remaining browser-call and OpenCode fixes, then verified the resulting branch.

- 1,529 workspace tests passed; 13 ignored. The new OpenCode wire regression reproduces the old failure and passes with the fix.
- Real-account queue lifecycle checks passed on all six available harnesses. OpenCode passed three consecutive runs, cancellation/Send now, and queued-image delivery.
- [Linux and macOS CI passed](https://github.com/zeronsh/comet/actions/runs/34326069385), including 775 Linux UI tests and native macOS browser/frame checks.

The iOS fixes were verified in source; no iOS runtime tests were available locally. My prior request for changes is satisfied.

Both queued OpenCode prompts now receive answers in order

PR #286, opened 2026-09-09
Add Linux sidebar browser with GPUI compositing

wingleeio:
Linux browser tabs now display inside the sidebar on X11 and Wayland. A lazily started WebKitGTK helper renders offscreen frames into the GPUI scene, so resizing, sidebar transitions, tooltips, and frosted overlays compose with the rest of the app.

Includes navigation/history, shared ephemeral tab cookies, keyboard and mouse input, Unicode clipboard and IME composition, GPUI select/context menus, high-DPI sizing, and load-error handling. WebKit remains a distribution-managed runtime; only the small helper executable is embedded. Installation and rendering costs are documented in `docs/reference/linux-browser.md`.

The real-shell fixture passes locally on **X11 and Wayland with both Vulkan and OpenGL**, including OpenGL at **2× scale**. It checks resize/CSS reflow, interrupted sidebar toggles, rapid tooltip hover, dark/light/opaque overlays, outside-click isolation, clipboard and composition, select/context menus, tab lifecycle, navigation, and load failures. Evidence uses Xvfb/Openbox and nested Weston with real GPUI/WebKit processes and software rendering. The videos are continuous native-mouse recordings; media is uploaded as user attachments and is not committed.

[Final CI passed](https://github.com/zeronsh/comet/actions/runs/34386293995): 777 UI tests, four Linux native runs (X11/Wayland with automatic adapter selection and forced OpenGL), macOS browser and frame-recovery regressions, and iOS tests. Ubuntu 24.04 selected OpenGL in both modes; Vulkan was additionally validated locally. The full Linux application check also passed.

Depends on https://github.com/zeronsh/zui/pull/6 and https://github.com/zeronsh/gpui-component/pull/3. These fix a Linux portal startup borrow panic and a real OpenGL blur failure found by Ubuntu 24.04 CI. Surfaces without framebuffer-copy support now use a copyable target for backdrop effects instead of skipping blur. The unchanged pixel assertion passes with the fix: dark checkerboard edge contrast dropped from 10.16 to 0.055.

Frame analysis found no black page frames during the hover/resize interval in either attached recording.

Recordings below: **X11/Vulkan**, then **Wayland/OpenGL**. Each exercises input/popups, hover and resizing, sidebar toggles, then blur/appearance checks.

https://github.com/user-attachments/assets/9c7c9172-6981-43b7-ae7e-f47510764712

https://github.com/user-attachments/assets/ff6aae6f-a7f5-4c83-8b37-bc774ae2179e

Linux sidebar browser

OpenGL: dark frosted overlay over the live browser

OpenGL: light frosted overlay over the live browser

Wayland OpenGL at 2× scale after resizing

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
             libxkbcommon-dev libxkbcommon-x11-dev libwayland-dev \
             libx11-dev libxcb1-dev libx11-xcb-dev \
             libfontconfig1-dev libfreetype-dev libasound2-dev \
-            libvulkan-dev pkg-config cmake
+            libvulkan-dev pkg-config cmake libwebkit2gtk-4.1-dev libjson-glib-dev
       - name: rust toolchain (rust-toolchain.toml)
         run: rustup show active-toolchain || rustup toolchain install
       - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/ui-tests.yml
+++ b/.github/workflows/ui-tests.yml
@@ -147,7 +147,11 @@ jobs:
         run: cargo build --release --locked -p zeron-ui --example browser-fixture --features browser-fixture
       - name: X11 and Wayland browser regressions
         timeout-minutes: 5
-        run: scripts/test-linux-browser.sh "$RUNNER_TEMP/linux-browser"
+        env:
+          RUST_LOG: gpui_wgpu=info
+        run: |
+          scripts/test-linux-browser.sh "$RUNNER_TEMP/linux-browser/default"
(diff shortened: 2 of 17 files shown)


jsgrrchg:
I tested this PR on Fedora 44. The build initially failed because the new Linux browser helper requires the WebKitGTK and JSON-GLib development packages, but the error message only mentions the Debian/Ubuntu package names.

On Fedora, I had to install:

```bash
sudo dnf install webkit2gtk4.1-devel json-glib-devel
```

After installing them, the build completed successfully.

Could we add Fedora instructions to `docs/reference/linux-browser.md` and possibly make the build error more distribution-neutral? It may also be worth clarifying that release users need the WebKitGTK and JSON-GLib runtime packages, since the Linux installer does not currently install or validate them.

PR #287, opened 2026-09-09
Fix file tree flicker and preserve browsing state during refreshes

jsgrrchg:
Files could flash blank, jump to the top, and collapse expanded folders when directories loaded or filesystem events arrived. The periodic watcher repair also reset the entire browser, and opening another file tab could make existing subscribers incorrectly report missing events. This change preserves the visible tree during refreshes and fixes watcher sequencing.

## Improvements

- **Preserve the viewport during tree updates.** Replace unconditional list resets with a splice of the changed row range. Anchor scrolling to the first visible path and its pixel offset; fall back to a visible ancestor or neighboring row if that path disappears. Unchanged rows do not trigger list mutations, and file rows use stable path-based element IDs. Generation changes still reset the list when the underlying tree is explicitly reset.
- **Keep cached content visible while loading or recovering from errors.** Distinguish an initial load from a refresh, including previously loaded empty directories. Root reloads no longer replace the tree with a blank panel, and background loads do not insert loading rows into cached directories.
- **Reconcile directory contents instead of rebuilding every subtree.** Retain loaded descendants, expansion, and selection for surviving entries. Remove genuinely missing subtrees and handle entries changing between files and directories. Partial pages cannot establish that an unseen child was deleted; track children across pages before pruning.
- **Refresh cached pages together.** Collect enough pages to cover previously cached children before publishing the updated listing. Initial loads remain lazy, and normal refreshes stop once cached paths are accounted for. If a cached child is missing, read to the end to confirm its removal. Later-page failures and repeated cursors return an error without publishing a partial refresh. Confirming missing children can therefore require additional page requests in large directories.
- **Make watcher resyncs preserve browsing state.** The existing 120-second repair, startup resync, and sequence-gap recovery now invalidate and reload directories without clearing the tree. Changes arriving during a load mark it for a follow-up reload. Closed directories retain their cache but become stale so reopening them fetches current contents.
- **Fix subscription and lag-recovery sequencing.** Only broadcast events advance the shared sequence. Serialize publication with receiver creation and baseline capture so new subscriptions cannot create artificial gaps in other tabs. Lagging subscribers get a fresh receiver and current baseline, discarding obsolete queued frames without consuming a shared sequence number.

## Validation

- `cargo test -p zeron-ui -p zeron-engine --lib files::` — **102 passed**: 75 UI Files tests and 27 workspace-files engine tests.
- Added regressions for scroll anchors and fallback, unchanged rows, retained expansion/selection, empty and failed refreshes, paginated reconciliation, deleted/type-changed nodes, invalidation during loads, lazy pagination, later-page failures, new subscriptions, and lag recovery.
- Targeted `rustfmt --check` on all six changed files and `git diff --check` passed.
- Interactive visual verification in the running app remains pending.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/287"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791569705&installation_model_id=425430&pr_number=287&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F287&signature=5119509083c7e37cee7661a9628e13de9de6779160217eb6ae18edc7a6cec54e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/engine/src/workspace_files.rs
+++ b/crates/engine/src/workspace_files.rs
@@ -6,7 +6,7 @@
 use std::collections::{HashMap, HashSet};
 use std::io::Write;
 use std::path::{Component, Path, PathBuf};
-use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, Weak};
 use std::time::{Duration, Instant};
 
@@ -65,7 +65,8 @@ struct WorkspaceFileKey {
 struct CheckoutWatch {
     checkout_id: String,
     root: PathBuf,
-    sequence: AtomicU64,
+    // Serializes publication with subscription baselines, including lag recovery.
+    sequence: Mutex<u64>,
     subscribers: AtomicUsize,
     changes_tx: broadcast::Sender<WorkspaceFileChanges>,
     cancel: CancellationToken,
@@ -548,7 +549,7 @@ impl CheckoutWatch {
         let watch = Arc::new(Self {
             checkout_id,
             root,
-            sequence: AtomicU64::new(0),
+            sequence: Mutex::new(0),
             subscribers: AtomicUsize::new(0),
             changes_tx,
             cancel,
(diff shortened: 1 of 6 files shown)


wingleeio (approved):
Reviewed tree reconciliation, cached pagination and error handling, viewport anchoring, invalidation during loads, and watcher subscription/lag sequencing. No blocking findings. Local validation passed: 102 Files tests (75 UI, 27 engine), targeted rustfmt checks, and git diff --check. All three CI checks passed. Interactive visual verification was not performed.

PR #309, opened 2026-09-10
Fix false completion notifications when sending queued messages

jsgrrchg:
Sending a queued message immediately interrupts the current response, briefly changing the session from Working to Idle before the replacement starts. The notification detector treated that transition as a successful completion and played the chime and posted “Run finished”.

This change uses a synchronized completion marker to distinguish completed responses from interruptions and steering handoffs. Normal queued turns still notify when they finish, even when the next turn starts before the UI observes Idle. Input requests and responses settled by the engine’s existing quiescence watchdog retain their notifications.

- Persist the last completed turn across session updates and both workspace storage paths; consume stale or duplicate markers silently.
- Track explicit queue sends until the host acknowledges or rejects delivery, including when the user switches chats.
- Serialize steering mailbox acceptance with its pending-delivery ledger in both explicit steering and warm dispatch, preventing a fast confirmation from leaving an orphaned pending entry.

The new session field is optional when reading older rows. Completion notifications from a remote host require that host to emit the new metadata.

Validation: 184 tests passed across the engine lifecycle/queue suites, protocol and document unit tests, and notification tests; one existing engine test was ignored. `cargo check --workspace --all-targets` and `git diff --check` passed. Native sound/banner presentation was not manually exercised.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/comet/pr/309"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791652742&installation_model_id=425430&pr_number=309&repository=zeronsh%2Fcomet&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fcomet%2Fpull%2F309&signature=a0c237aebf12cb264ead8dbf7f5f5b42f076aaf4d20c2493e5ce84179c1ec24b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/doc/src/registry.rs
+++ b/crates/doc/src/registry.rs
@@ -1159,6 +1159,7 @@ impl RegistryDoc {
             ("chatId", json!(session.chat_id)),
             ("deviceId", json!(session.device_id)),
             ("status", serde_json::to_value(session.status)?),
+            ("lastCompletedTurn", json!(session.last_completed_turn)),
             ("startedAt", opt_ms(session.started_at)),
             ("updatedAt", json!(session.updated_at.timestamp_millis())),
         ]);
@@ -1294,6 +1295,7 @@ impl RegistryDoc {
                     ("chatId", json!(session.chat_id)),
                     ("deviceId", json!(session.device_id)),
                     ("status", serde_json::to_value(session.status)?),
+                    ("lastCompletedTurn", json!(session.last_completed_turn)),
                     ("startedAt", opt_ms(session.started_at)),
                     ("updatedAt", json!(session.updated_at.timestamp_millis())),
                 ]),
--- a/crates/doc/src/registry/tests.rs
+++ b/crates/doc/src/registry/tests.rs
@@ -292,6 +292,7 @@ fn space(id: &str, device_id: &str, path: &str) -> Space {
 
 fn session(chat_id: &str, device_id: &str, status: SessionStatus) -> Session {
     Session {
+        last_completed_turn: None,
         chat_id: chat_id.into(),
         device_id: device_id.into(),
         status,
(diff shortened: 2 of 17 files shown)


wingleeio (approved):
Audited and reviewed the diff, ran the engine/doc/proto/ui suites locally on the branch (1260 passed, 0 failed), CI is green.

The approach is sound: publishing `lastCompletedTurn` atomically with the settled status, and gating the Done chime on the marker changing rather than on a Working→Idle edge, correctly separates interrupts and steer handoffs from real completions. The ledger-under-lock change in both steer paths closes the fast-`Steered`-before-push race cleanly, and nothing in the harness crate reads `SteerMessage.message_id`, so always minting it is inert. iOS decodes session rows field-by-field, so the new key is ignored there.

Two behaviour notes, neither blocking:
- A viewer on this build watching a host on an older build will no longer get a Done chime at all (the row never carries the marker). Acceptable since devices auto-update together, but worth knowing when triaging "no chime" reports for a few days.
- A steer accepted for the next boundary suppresses the chime for the turn it lands on, and only the steered continuation notifies. That matches the "steer owns the continuation" intent; just flagging it as a deliberate change.

Merging.

PR #10, opened 2026-08-02
Add light mode

neo773:
Adds light mode. Dark mode is unchanged.

Screenshot 2026-08-02 at 8 20 13 PM

Screenshot 2026-08-02 at 8 45 53 PM

## How it works

Copied Zed's approach:

- `Theme` is a gpui global with two instances. Components read it with `Theme::of(cx)`.
- The OS appearance is watched with `observe_window_appearance`. On a change we swap the global and call `cx.refresh_windows()`. Colors are read at paint time, so `notify()` is not enough.
- Window vibrancy is re-applied after every theme swap, because gpui's macOS backend drops the `NSVisualEffectView` otherwise.

## Colors

Light is not an inversion of dark:

- Text tones are matched to their dark counterpart by contrast ratio, not by lightness.
- Accents move from the 400 step to 600. amber-400 measures 1.9:1 on white.
- Light chrome is opaque. A translucent light tint picks up the desktop colour and goes muddy.
- Elevation reverses: on dark a faint white wash means raised, on light that reads as recessed, so light lifts with white plus a border.

Tests check the contrast ratios rather than eyeballing them.

## UI

Settings → Appearance picks System / Light / Dark. Same options in the View menu. The preview cards use a new `option_card` widget that takes any preview element, so it is reusable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #313, opened 2026-09-10
Add native Windows support and portable releases

hoangvu12:
Adds native Windows x64 source builds and portable release ZIPs, covering application storage and locking, agent discovery and launching, terminal process cleanup, desktop behavior, and in-app updates. Installers and background services remain unsupported.

Windows agents and terminals use creation-time Job Object membership. npm `.cmd`/`.bat` launchers use hardened argument escaping; unsafe line breaks and expansion syntax in executable paths are rejected. Terminal discovery fails closed when a bare shell cannot be resolved. Generated-image imports pin their Windows source and directory handles, reject reparse points during traversal, and reject concurrent writers or replacement attempts. Portable updates require HTTPS and SHA-256 verification, verify a replacement copy before swapping, and restore the previous installation when the final rename fails.

The pinned ZUI revision includes [ZUI #7](https://github.com/zeronsh/zui/pull/7) and [ZUI #9](https://github.com/zeronsh/zui/pull/9). [ZUI #10](https://github.com/zeronsh/zui/pull/10) is not included: Windows uses native Acrylic window chrome while in-app panels stay opaque until that renderer work is ready. The lockfile also updates Rustls to 0.23.45 for RUSTSEC-2026-0285.

See [Windows development](https://github.com/zeronsh/zeron/blob/0cc2f0778d136d8477c6ef265baba928eb4cd896/docs/reference/windows-development.md) for setup, supported behavior, and remaining acceptance work. CI covers Windows release compilation, shader layouts, synthetic native agent/process tests, ConPTY, locking, auth fixtures, UI, and startup. Native font geometry is tested with DirectWrite; deterministic UI tests use their own animation scheduler, and a separate test exercises the Windows high-resolution timer. Live authenticated provider runs and broader GPU/DPI coverage remain manual acceptance work. See the review for validation of the final commit.


crates/update/src/windows.rs
```diff
@@ -0,0 +1,244 @@
...
+/// Replace the running executable. The library handles Windows image locks and
+/// deferred cleanup; the desktop relaunch waits for the old engine to exit.
+pub fn apply(staged: &Path, directory: &Path, relaunch: bool) -> anyhow::Result<()> {
+    let installed = directory.join("zeron.exe");
+    ensure!(
+        std::env::current_exe()?.canonicalize()? == installed.canonicalize()?,
+        "update must run from its installation"
+    );
+    verify(staged)?;
+    self_replace::self_replace(staged).context("replacing Windows executable")?;
```
wingleeio, line 124: [P1] Preserve a launchable installation when replacement fails

`self-replace` 1.5.0 is not transactional on Windows: its `self_replace` renames the running executable away and schedules its deletion **before** copying the staged executable back to the installation path. If the subsequent copy/rename fails (for example, insufficient disk space for that extra copy or an antivirus/file-sharing error), this `?` returns without restoring `zeron.exe`. The UI then reports an update failure, but the normal launch path is gone and the cleanup helper may delete the old executable when the app exits. Please retain a recoverable backup and restore the installed path on every post-rename failure, or use a replacement procedure with those guarantees. Add a failure-injection test after the old image has been moved. This follows directly from the pinned dependency's Windows `self_replace` implementation; the existing checksum tests only cover failures before replacement.

crates/harness/src/claude/mod.rs
```diff
@@ -64,43 +63,14 @@ use wire::{ControlRequestFrame, Frame, allow_response, control_response_line};
...
-    );
-    candidates.into_iter().find(|p| p.exists())
+    let mut extra = Vec::new();
+    if let Some(home) = crate::executable::home_dir() {
+        extra.push(home.join(".claude").join("local").join("claude"));
+        extra.push(home.join(".local").join("bin").join("claude"));
+    }
+    extra.push(PathBuf::from("/opt/homebrew/bin/claude"));
+    extra.push(PathBuf::from("/usr/local/bin/claude"));
+    crate::executable::find_on_paths("claude", extra)
```
wingleeio, line 73: [P2] Keep Claude availability consistent with the executable override

This refactor removes `CLAUDE_CODE_EXECUTABLE` handling from `resolve_claude_executable`, but `ClaudeHarness::installed()` still calls this function rather than `self.resolve_executable()`. When Claude is available only through a valid custom override, the launch resolver accepts it while the registry reports `installed = false` and leaves it disabled/unavailable in the composer. This also regresses Linux/macOS. I reproduced it with the override pointing to an existing executable and the normal discovery locations hidden. Please use the validated launch resolver for `installed()` as this PR already does for Codex, and cover override-only availability.

crates/engine/src/agent_accounts.rs
```diff
@@ -589,13 +589,13 @@ impl AgentAccounts {
             .root_dir()
             .join(format!(".login-{login_id}"));
         std::fs::create_dir_all(&home)?;
-        let mut command = tokio::process::Command::new("codex");
+        let mut command = zeron_harness::process::Command::new("codex");
```
wingleeio, line 592: [P2] Reuse native Codex discovery for account login

For the Windows npm layout newly supported by `executable::find_on_paths`, PATH contains `codex.cmd` while the real `codex.exe` lives inside the platform package. The catalog finds that payload, but this login path constructs `Command::new("codex")`; the new Windows command resolver searches for a top-level `codex.exe` and neither follows the npm payload nor honors `CODEX_EXECUTABLE`. Consequently Add account fails with “CLI was not found” for an installation advertised as supported, even though the harness can run it. Resolve the same native executable and compose its child PATH here, then test the account-login path with the npm fixture and an override-only installation.

wingleeio (changes requested):
Reviewed a48e1fb2347657ab3afd9ba3661da4968559438e. Requesting changes for the three issues below: a failed Windows update can remove the launchable installation, Claude availability no longer honors its environment override, and Windows npm Codex discovery is not used by account login.

Validation on Linux: 174 engine unit tests, 130 harness unit tests, 8 auth integration tests, and 5 updater unit tests passed (317 total). A separate isolated regression probe reproduced the Claude override failure. I inspected both pinned UI dependency changes and the new updater dependency. I could not run the Windows build, ConPTY/Job Object tests, GUI probes, or a real Windows update here; this head currently has no successful Windows CI result.

**WorkOS on Windows:** I found no Windows-specific blocker in the implemented authentication path. The desktop uses GPUI's Windows `ShellExecuteW` browser opener, then receives the OAuth callback on `http://127.0.0.1:<port>/callback`; exchange, refresh, and organization requests go through the existing edge API. It does not depend on registering a custom URL scheme. Session persistence uses the new application data directory, shared by GUI and CLI. WorkOS documents loopback redirects for native clients ([redirect requirements](https://workos.com/docs/reference/sso/get-authorization-url)); the configured client must allow the callback URI/port pattern.

That is a source-level compatibility assessment plus passing stub-server auth tests on Linux, **not confirmation of an authenticated Windows run**. Please include `cargo test --release --locked -p zeron-engine --test auth` in Windows CI (the current workflow skips that integration target), and verify browser sign-in, app restart/session restore, refresh, org selection, and logout on Windows before claiming authenticated support. Windows session-file privacy currently relies on inherited directory ACLs; the Unix-only 0600 enforcement does not apply there.

The PR description also needs updating: the current diff adds portable packaging, publication, and in-app updates, although the description still lists them as follow-ups.

crates/harness/src/acp/mod.rs
```diff
@@ -631,12 +609,14 @@ impl AcpHarness {
     fn resolve_launch(&self) -> Result<Launch, HarnessError> {
         let spec_args: Vec<String> = self.spec.args.iter().map(|a| a.to_string()).collect();
         if let Some(p) = &self.executable {
-            return Ok(Launch::Program(p.clone(), spec_args));
+            return crate::executable::validate_native_override(p)
```
wingleeio, line 612: [P2] Model discovery now errors instead of falling back for a missing override

`models()` starts with `self.resolve_launch()?`. On `main`, an `executable`/env override that does not exist passed through here and failed at spawn, so `discover_models()` returned `Err` and `models()` served `(self.spec.models)()`. With `validate_native_override` the `?` fires first, so a stale `GROK_EXECUTABLE`/`PI_ACP_EXECUTABLE`/etc. now fails model discovery outright. The pre-existing regression test for this path, `models_fall_back_to_the_static_catalog_when_the_probe_fails` in `tests/acp.rs`, fails on this head with `NotInstalled("/nonexistent/never-a-pi-acp does not exist")`.

Meanwhile `AcpHarness::installed()` (line ~1125) still returns `true` for any `Some(executable)` or non-empty env override, so the registry can advertise an ACP agent as installed whose `models()` and `run()` both fail. That is the exact inconsistency this PR fixed for Claude and Codex. Please pick one: keep the static-catalog fallback (return `Ok(static)` when `resolve_launch()` fails here), or make ACP consistent with the other harnesses by gating `installed()` on `resolve_launch().is_ok()` and updating the test to expect the error. Either way the existing test needs to reflect the chosen behavior.

crates/harness/src/executable.rs
```diff
@@ -0,0 +1,646 @@
...
+    fn windows_extra_candidates_resolve_through_pathext_variants() {
+        let temp = tempfile::tempdir().unwrap();
+        let bin = temp.path().join("bin");
+        std::fs::create_dir_all(&bin).unwrap();
+        std::fs::write(bin.join("tool.cmd"), b"@echo off").unwrap();
+        // Extras are file candidates: an explicit extension is taken as given.
+        assert_eq!(
+            find_on_paths_with(
+                "tool",
+                vec![bin.join("tool.CmD")],
```
wingleeio, line 457: [P2] Two new unit tests assume a case-insensitive filesystem

`windows_extra_candidates_resolve_through_pathext_variants` writes `tool.cmd` and expects the `tool.CmD` extra to resolve, and `windows_pathext_environment_reorders_extensions` (line 429) writes `agent.cmd` and expects the `.CMD` PATHEXT variant to resolve. Both go through `is_file()` on the real filesystem, so they pass on NTFS and default APFS but fail on Linux and case-sensitive macOS volumes:

```
windows_extra_candidates_resolve_through_pathext_variants: left: None, right: Some(".../bin/tool.CmD")
windows_pathext_environment_reorders_extensions: shim variant resolves
```

This breaks `cargo test -p zeron-harness --lib` on Linux, which is the suite the PR description cites as the shared regression check. Please either match the on-disk case in these two tests (keep the case-insensitivity claim to the `candidate_names`/`extra_variants` output assertions, which need no filesystem) or gate the case-dependent assertions with `#[cfg(windows)]`. Both were added in 6f5a40a2.

crates/engine/tests/codex_login_resolution.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// The npm layout — a `codex.cmd` shim on PATH with the payload buried in
+/// `node_modules` — must log in through the same resolution the harness uses.
+#[test]
+fn codex_login_runs_the_npm_cmd_shim() {
+    let dir = tempfile::tempdir().unwrap();
+    let bin = dir.path().join("bin");
+    std::fs::create_dir_all(&bin).unwrap();
+    // The shim plays both roles itself.
+    std::fs::copy(fake_codex(&bin), bin.join("codex.cmd")).unwrap();
```
wingleeio, line 137: [P2] `codex_login_runs_the_npm_cmd_shim` cannot pass on Unix

This binary is not `cfg(windows)`-gated, but `.cmd` discovery only exists through the Windows PATHEXT variants: on Unix `find_on_paths("codex", ..)` looks for a file literally named `codex`, so `bin/codex.cmd` is never a candidate. On a runner without codex the child fails with `NotInstalled` while the parent expects `true`. On a dev machine with `/usr/local/bin/codex` it is worse: the absolute fallback resolves the real CLI, `start_login` launches a real `codex login` (against the throwaway `CODEX_HOME`, so no credentials are touched), and the assertion fails on the URL instead:

```
left: "", right: "https://auth.openai.com/authorize?probe=fake"
```

Please gate this test (or the whole file) with `#[cfg(windows)]`, and add the same `/opt/homebrew/bin/codex` / `/usr/local/bin/codex` guard the `without_any_cli` test already has to the override-only test so it never reaches a real CLI.

crates/harness/tests/claude_availability.rs
```diff
@@ -0,0 +1,84 @@
...
+    let name = if cfg!(windows) {
+        "claude-custom.exe"
+    } else {
+        "claude-custom"
+    };
+    let exe = dir.path().join(name);
+    std::fs::write(&exe, if cfg!(windows) { b"MZ" } else { b"sh" }).unwrap();
+    // PATH without any claude: only the override makes Claude installed.
+    probe(dir.path(), Some(&exe), true);
+    probe(dir.path(), None, false);
```
wingleeio: [P3] Second probe has no guard for a system-wide `claude`

The `expected = false` probe here removes `HOME` and empties `PATH`, but `resolve_claude_executable` still checks the absolute `/opt/homebrew/bin/claude` and `/usr/local/bin/claude` fallbacks, so on any machine with claude installed there the child reports `installed = true` and this test fails. The sibling `claude_availability_reports_not_installed_without_any_cli` already skips in that case; please apply the same guard (or drop this duplicate probe, since the sibling covers it).

wingleeio (changes requested):
Reviewed 73b245ccf1c0e3040dd2d6fb85c36eb341d22f9c. The three earlier requests are addressed on this head (the swap now copies and verifies before moving the running image aside, Claude `installed()` uses the launch resolver, and Codex login goes through `codex::login_command`). Requesting changes again for one behavior regression and three new tests that cannot pass on Linux.

**Validation on Linux (this head):** `cargo check --locked --workspace --all-targets` clean; engine 174 unit tests, auth 8, updater 10, app 5, claude 9, codex 20, cursor 20, acp_quiet 1, acp_stall 6 all passing. Failing: harness unit tests 129/131, `tests/acp.rs` 26/27, `codex_login_resolution` 3/4, `claude_availability` 2/3 — details inline. The `terminals` integration test (`m5_repos_diffs_terminals`) passed 21/22 with the one failure being a pre-existing SHA-prefix fuzzy-match flake that also reproduces independently of this branch. I rebuilt against the pinned Zui/gpui-base revisions; both match their upstream `main` heads.

**CI:** I approved the fork workflow runs on this head. Preview networking and UI tests (all five jobs) are green. Windows tests: green on this head (build, shader layout, app/updater, engine+harness unit, native harness integration, catalog, auth, UI, and the no-HOME CLI startup probe; the native GUI job is dispatch-only and was not run). CI, a fresh release build, and the GUI probes are no longer "pending" for this revision..

**Nits (no change required to land):**
- `release.yml`: `publish` now requires the Windows job, so a Windows-only build failure (fxc discovery, runner image drift) blocks Linux and macOS releases. That is a reasonable "every platform" policy, just calling it out as a deliberate change from the previous comment's intent.
- `crates/update/src/lib.rs` `release_base`: `ZERON_RELEASES_URL` is honored on every platform and is not HTTPS-checked, unlike the Windows `zeron-update.json` feed. Consider the same `https://` guard.
- `process/windows/command.rs` `quote_batch_argument`: agent arguments passed to a `.cmd` shim are quoted but not `%`-escaped, so a `%NAME%` inside an argv element is still expanded by cmd.exe. Today every harness argv is program-controlled (flags, model ids, settings JSON), so this is not reachable from prompts; worth a comment or `^%` escaping so it stays that way.
- `terminals/windows.rs` `Process::wait_for`: terminating the job when the shell exits also kills anything the user detached from that shell (`start notepad`, background servers), which differs from the Unix terminals where orphans survive. If intentional, a doc note in `windows-development.md` would help.

wingleeio:
Pushed three commits to this branch (maintainer edit) so it can land without another round trip:

- **c3df6624** Merge `origin/main` — resolves the `Cargo.toml` / `Cargo.lock` / `crates/ui/Cargo.toml` conflicts. Main had already moved the Zui and gpui-base pins past this branch's bump (`aa009411` / `94c1bbaf`), so those are kept from main; the ui manifest keeps both the `windows-render-fixture` and the new `appshots-fixture` examples. The lockfile diff against main is now only this branch's `windows-sys`/`embed-resource` dependency entries.
- **61306e40** Addresses the four review findings: `AcpHarness::installed()` validates overrides the same way `resolve_launch` does (with the static-fallback test rewritten around an executable that exists but cannot speak ACP, plus a new missing-override test); the two `executable.rs` PATHEXT tests now match their on-disk casing so they pass on case-sensitive filesystems; the npm `.cmd` login test is `cfg(windows)`; the claude availability probe shares its sibling's system-wide-claude guard.
- **ed05dc45** Fixes a main-side test the first Windows CI run surfaced: `settings::tests::round_trip` hard-coded `appshot_destination: NewSession`, but that field is `serde(skip)` outside macOS/Linux and reloads as the default there. The expectation is now platform-aware; no production change.

Linux on the merged head: engine 177, harness 131, acp 28, claude_availability 3, codex_login_resolution 3, terminals 22, auth 8, updater 5, app 10 — all passing; `cargo check --locked --workspace --all-targets` clean. I've approved the CI runs on the new head. If you'd rather structure any of this differently, feel free to amend — nothing here changes the Windows-side behavior.

wingleeio (changes requested):
Security review of `01bcb8ee4cbe5323d3dc1329f49830754a2aa6a9`: **requesting changes; not cleared for approval/merge yet**.

I found no evidence of malicious behavior in the contributor's changes. I reviewed the complete diff against current main, Windows process/job/handle ownership, argument encoding, executable discovery, ConPTY, updater/packaging, build workflows, and related ZUI/dependency changes. This is a code review, not proof of anyone's intent or a guarantee of zero vulnerabilities.

**Remaining dependency blocker:** a clean resolution of the unchanged `@cursor/sdk@1.0.28` pin installs `@connectrpc/connect-node@1.7.0 -> undici@5.29.0`. Current npm audit reports twelve Undici advisories, with three affected package nodes after propagation and no compatible automatic fix. Connect uses Undici for a Headers polyfill, but importing it also initializes `Symbol.for("undici.globalDispatcher.1")` when absent. The SDK uses global fetch, including its local fetch executor. I therefore cannot dismiss all HTTP-client exposure as dead code.

The published SDK additionally bundles **Undici 6.23.0**, outside the npm audit graph. Its tree-shaken auth module imports ProxyAgent, and module evaluation likewise initializes the shared dispatcher when absent. An npm override alone does not replace this bundled copy. For example, [GHSA-35p6-xmwp-9g52](https://github.com/nodejs/undici/security/advisories/GHSA-35p6-xmwp-9g52) affects HTTP/1.1 keep-alive response handling below 6.27.0 and requires an attacker-controlled upstream. I have not reproduced an exploit against Zeron, and did not establish reachability for the reported high-severity WebSocket issues. A supported dependency update or tested transport mitigation is required for the requested whole-dependency clearance. The registry's latest SDK 1.0.31 still declares Connect Node ^1.6.1, so an SDK version bump alone is not an established fix. **This dependency debt already exists on main and is not evidence against this contributor.**

Fixes pushed during this review:
- Replaced unsafe Windows batch argument encoding with Rust's hardened strategy; added native injection-sentinel, round-trip, Unicode-path and rejection tests.
- Removed unresolved Windows shell fallback to current-directory executable search.
- Required HTTPS update feeds and rejected redirect downgrades; patched Rustls 0.23.43 to 0.23.45 for [GHSA-2mjx-qc3c-rqvc](https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc).
- Resolved main conflicts while preserving shared Codex login resolution; kept unsupported Windows in-app blur opaque.
- Added Windows login, startup and portable packaging checks; fixed the packaging version probe they exposed.
- Fixed the Linux browser test's menu-dismissal timing race while retaining both click-isolation assertions.

ZUI remains pinned to the same revision as main. Reviewed [ZUI #7](https://github.com/zeronsh/zui/pull/7) and [#9](https://github.com/zeronsh/zui/pull/9), including DirectX shader layout parity. Inspected [draft #10](https://github.com/zeronsh/zui/pull/10); it is not included or approved here. Grok/pi-acp metadata, archive integrity and relevant install code were checked; provider prebuilt binaries were not reverse-engineered.

Validation: workspace all-target check passed; 1,032 UI unit tests passed; broad integration invocation passed 589 tests with 16 skips (overlapping other unit runs). Final native X11 and Wayland browser checks passed locally. Cargo audit now has zero vulnerability findings, with the same pre-existing maintenance/soundness/yanked warnings as main. The final Windows native harness security stage passed; [Windows CI](https://github.com/zeronsh/zeron/actions/runs/35020819351) and [UI/platform CI](https://github.com/zeronsh/zeron/actions/runs/35020819152) are still running. Their completion will not by itself resolve the npm dependency blocker.

wingleeio (approved):
Reviewed the final PR changes, including the Windows generated-image import hardening added after native CI exposed a replacement race. Windows imports now hold the source and traversed directories open without write/delete sharing, reject reparse points, and release the handles after publication. The Windows regression attempts source overwrite/replacement and parent rename; Unix retains its mutation-detection regression.

Native font geometry now exercises DirectWrite on Windows, and deterministic UI schedulers use their own timers. A separate real waitable-timer test preserves native clock coverage. The full Windows UI stage passed after these fixes.

The previous audit also corrected Windows batch command injection, unresolved shell fallback, insecure update-feed overrides, the Rustls advisory, unsupported in-app blur, Codex login merge conflicts, portable packaging and browser-test timing. The ZUI pin is unchanged from main; related ZUI #7 and #9 were reviewed, and draft #10 is not included.

No evidence of malicious behavior or remaining PR-introduced security vulnerabilities was identified in the reviewed changes. This is a scoped review, not a guarantee that the application or third-party binaries contain no vulnerabilities.

Per the maintainer's explicit scope decision, the pre-existing Cursor SDK/Undici findings documented in the [earlier review](https://github.com/zeronsh/zeron/pull/313#pullrequestreview-5215548778) are deferred separately and do not block this Windows-support PR. Those findings have not been fixed or withdrawn. Existing Cargo maintenance/soundness warnings and the limitations of reviewing prebuilt provider binaries likewise remain documented.

Final reviewed commit: `0cc2f0778d136d8477c6ef265baba928eb4cd896`. All applicable CI checks passed: [native Windows tests/build/portable packaging/startup probes](https://github.com/zeronsh/zeron/actions/runs/35029384081), [UI/platform and session regressions](https://github.com/zeronsh/zeron/actions/runs/35029384083), and [preview networking](https://github.com/zeronsh/zeron/actions/runs/35029384162). Local workspace checks, Linux UI/browser tests, and targeted security regressions also passed. Opt-in native GUI/pixel probes and live authenticated provider runs were not executed.

This approval supersedes my earlier changes-requested review: the PR-specific issues are fixed, and the maintainer has accepted tracking the existing Cursor dependency findings separately.

PR #320, opened 2026-09-11
Add a configurable Zeron sound signature

gaelcado:
## Summary

- replace the existing completion and input-request chimes with a quiet, rounded sound pair
- add a restrained attention cue for run failures and durable connectivity outages
- add a master **Session sounds** switch plus independent completion, input, and attention preferences
- preserve each choice in device-local settings while keeping `ZERON_DISABLE_SOUND` as the global override
- give every Notifications switch a 40×40 target, keyboard focus/activation when enabled, and persistent accessible role/name/state while subordinate choices are unavailable
- send desktop banners for genuine durable connectivity outages, respecting the background-only preference
- keep the selected sounds and ten additional auditions reproducible from standard-library-only synthesis scripts with no external samples

## Listen

These immutable CDN links open in a browser’s native audio player without forcing a download:

- [Completion — 0.52 s](https://cdn.jsdelivr.net/gh/gaelcado/comet@770b5d217ee349c83558808a55eaf3506e26a948/crates/ui/assets/sounds/done.wav)
- [Input required — 0.60 s](https://cdn.jsdelivr.net/gh/gaelcado/comet@770b5d217ee349c83558808a55eaf3506e26a948/crates/ui/assets/sounds/request.wav)
- [Attention — 0.65 s](https://cdn.jsdelivr.net/gh/gaelcado/comet@770b5d217ee349c83558808a55eaf3506e26a948/crates/ui/assets/sounds/attention.wav)
- [Appshot capture — 0.67 s, supplied by #216](https://cdn.jsdelivr.net/gh/gaelcado/comet@ab51c3fbe1fa2c21bd629160c2f5adb871d4882d/crates/ui/assets/sounds/appshot.wav)

All four are stereo, 48 kHz, 16-bit PCM WAV files at deliberately quiet peaks. The complete audition set is documented under `docs/sound-design/auditions/`.

## Behavior

- session baselines seed silently, so boot and replay do not chime
- connectivity alerts arm only after the engine startup grace, suppressing both warm and cold boot-time outages
- completion follows a new explicit completion marker; interrupts, stale activity, and pending-send handoffs do not create false completion sounds
- input required follows the transition into `AwaitingInput`
- run failure plays attention once on the transition into `Errored`
- connectivity attention and desktop banners fire once per genuine durable outage; the engine already applies a four-second degradation grace before exposing `Offline` or `Reconnecting`
- attention requests are coalesced across independent session and connectivity watch callbacks, including multiple simultaneous run failures
[...]

PR #216, opened 2026-08-23
Add Appshots capture, queue integration, and iOS presentation

gaelcado:
## Summary

Add Appshots: capture another application's window with a global shortcut, then review and send the capture from Zeron. Each Appshot pairs a screenshot with its application/window identity and optional accessibility context. Capturing stages a draft; sending remains an explicit user action.

## User experience

- **Dedicated settings:** Settings → Appshots contains enablement, the configurable global shortcut, destination policy, capture sound, and platform-specific permission status. Controls support keyboard navigation and activation, accessible names/state, and visible focus.
- **Capture and review:** Invoke the shortcut while another application is focused. Zeron captures before taking focus and stages the result in the chosen composer. Cards show the window preview, source application and title, with full-image preview and removal before sending. Invoking the shortcut within Zeron does nothing.
- **Transcript:** Sent Appshots retain distinct source-labelled cards instead of becoming anonymous image attachments. Loading and unavailable states retain source information. Accessibility context is hidden from displayed message text.
- **Queue:** Image previews use compact cropped thumbnails and an additional-image count. Ghost actions and modifier hints occupy stable slots. Narrow layouts reduce the visible thumbnail count. Editing, failed sends and draft restoration preserve screenshot/context associations.
- **iOS:** Received Appshots render as a horizontal strip of source-labelled cards with full-image previews. Queue rows expose a gallery for additional images and touch-sized edit/send/menu controls. Editing a queued message preserves its original Appshot context and attachment references.

## Screenshots

Native desktop and iPhone simulator frames using neutral fixture content. These demonstrate presentation, not live capture or physical remote delivery. Light desktop exports omit the macOS compositor backdrop; the offline iOS host disables Send now.

**Desktop composer and queue**

Desktop composer and queue

**Distinct Appshot cards in the transcript**

Distinct Appshot cards in the transcript

**Dedicated Appshots settings**

Dedicated Appshots settings

**iPhone transcript and compact queue**

iPhone transcript and compact queue

**iPhone queued-image gallery**

iPhone queued-image gallery

<details>
<summary>Desktop light and narrow layouts</summary>

**chat light 1100**

chat light 1100

**chat light 700**

chat light 700

**chat light 390**

chat light 390

**chat light 320**

chat light 320

**transcript narrow 390**

transcript narrow 390

</details>

<details>
<summary>Settings and keyboard states</summary>

**settings light 1100**

settings light 1100

**settings dark 600**

settings dark 600

**settings enabled keyboard**

settings enabled keyboard

[...]

PR #322, opened 2026-09-11
Add a guided first-run onboarding journey

gaelcado:
Zeron welcomes new profiles with a six-step setup journey for workspace mode, appearance, coding agents, session defaults, automatic titles, and a project. Existing profiles continue into the workspace; setup can be reopened from the user menu.

The journey uses a consistent 620 × 680 portrait canvas. Entering and leaving setup uses the existing 420 ms new-thread motion curve with a fade-through around the window resize. Interrupted transitions resume from the presented size and opacity, reduced motion snaps to the destination, and initial startup mounts setup immediately so keyboard focus remains usable. macOS keeps the portrait aspect during setup and releases it for the workspace.

Content fills the space between the heading and pinned navigation. Agent and model lists expand with the available height and show edge fades only where content overflows. Setup choices stay device-local and use their existing settings owners; agent actions target this device.

Supports skip, defer/resume, Back, keyboard choice navigation, close confirmation, modal focus containment, and deterministic onboarding fixture routes.

## Validation

Latest polish commit: `93abcd43`.

- `cargo test -p zeron-ui onboarding --lib`: 23 passed.
- Portrait canvas regression: 1 passed.
- Checks include interrupted resize/restoration, initial focus, modal keyboard behavior, and footer placement at normal/minimum sizes; agent scrolling is checked against the footer bounds.
- `git diff --check`: passed.
- The full `zeron-ui` library suite terminated with SIGABRT before completion; it is not a full-suite pass.

## Visual evidence

Fresh native motion review and verified screenshots are pending. Earlier local captures were rejected after the user identified incorrect window targeting and are not presented as evidence. The capture workflow now protects the production/control instance and requires the expected accessibility view in the exact preview process before selecting a window.


gaelcado:
### CI diagnosis

The failing [`ios-tests` job](https://github.com/zeronsh/zeron/actions/runs/34609268138/job/103295354801) on head `c6850db` had 1 failure out of 103: `TranscriptLayoutTests.testToolGroupsRevealAndCollapseThroughIntermediateHeights`, at the intermediate-height timing assertion.

This PR changes no iOS files. I ran that exact test from the unchanged PR head locally with `-test-iterations 3`; all 3 iterations passed. The local machine has Xcode 26.3 while Actions used Xcode 26.6, so this does not prove identical runner behavior, but the zero iOS diff plus repeated local passes strongly identifies a pre-existing simulator timing flake rather than an onboarding regression.

Provenance: Wing introduced the test in `f9fbca6` / #270 and later relaxed its timing assertions in `c31b440` / #291. This is existing repository test reliability debt (maintainer/CI ownership), not an issue in this PR; the provenance is included to route follow-up, not to assign personal blame.

I attempted to rerun the failed GitHub job, but GitHub rejected it with `Must have admin rights to Repository`; no rerun occurred. A maintainer can rerun the failed job, or fresh CI can run after the documented rebase onto #320.

crates/ui/src/shell.rs
```diff
@@ -3051,43 +3076,813 @@ impl Shell {
...
+                            Some(row.id) == shell.onboarding.selected_harness
+                                && row.id != zeron_proto::HarnessId::Mock
+                                && crate::onboarding::harness_is_ready(
+                                    row,
+                                    &shell.onboarding.accounts,
+                                )
+                        })
+                    },
+                );
+                if !selected_is_ready {
```
wingleeio: **Default agent race.** This callback runs when `LIST_HARNESSES` returns, which is before `LIST_AGENT_ACCOUNTS`. At that moment `harness_is_ready` is false for Claude/Codex/Cursor (accounts still `Loading` → "Checking sign-in…"), so `selected_harness` falls through to the first harness that needs no account — Devin on my box. When accounts arrive, the accounts callback keeps it because Devin *is* ready, and Harnesses→Continue then persists it via `persist_onboarding_defaults`.

Reproduced twice on a fresh profile with Claude Code detected as Ready and listed first; `composer-defaults.json` ends up with `"harness": "devin"` and the Defaults step opens on Devin.

Suggestion: only auto-pick once **both** `harnesses` and `accounts` are `Ready` (a small `fn resolve_default_harness(&mut self)` called from both callbacks), and prefer registry order at that point. Alternatively keep `selected_harness = None` until the user reaches Defaults and pick there.

crates/ui/src/shell.rs
```diff
@@ -3051,43 +3076,813 @@ impl Shell {
...
-        self.settings.git_history_column_widths = crate::history::configured_column_widths(cx);
-        self.settings.git_history_column_order = crate::history::configured_column_order(cx);
-        self.settings.git_history_author_display = crate::history::configured_author_display(cx);
-        self.settings.theme_selection = crate::appearance::themes(cx);
-        self.settings.accent = crate::appearance::accent(cx);
-        self.settings.surface = crate::appearance::surface(cx);
-        self.settings.ui_font_family = crate::typography::requested(cx);
-        self.settings.ui_font_size = crate::typography::font_size(cx);
-        settings::replace(self.settings.clone(), SavePolicy::Debounced, cx);
+    pub(super) fn onboarding_remember_target(&mut self, cx: &mut Context<Self>) {
```
wingleeio, line 3957: **Project choice is lost on Finish.** This loads a fresh `ComposerDefaults` from disk and saves it, but `Pickers` holds its own in-memory copy (loaded at boot) and every `Pickers::save_defaults()` rewrites the whole file from that copy. Traced on a fresh run:

```
after "Continue without a project": noProject=True,  device=<local id>
after "Finish setup":               noProject=False, device=None   (modelLabels 413→437 = pickers catalog save)
```

Relaunching the finished profile then lands on "Add a project to get started" with no composer (screenshot in the review body), so the choice the step promised does not survive.

Suggestion: make `Pickers::remember_target` `pub(crate)` and call it through the pickers entity from here and from the `spaces.rs` hook, so there is a single owner of `composer-defaults.json`.

crates/ui/src/shell.rs
```diff
@@ -6180,15 +7128,393 @@ impl Shell {
...
         cx: &mut Context<Self>,
     ) {
-        if event.keystroke.key == "escape" && self.capture_escape_surface(cx) {
+        if event.keystroke.key == "escape" && self.capture_escape_surface(window, cx) {
             cx.stop_propagation();
         }
     }
 
-    fn on_key_down(&mut self, event: &gpui::KeyDownEvent, _: &mut Window, cx: &mut Context<Self>) {
+    fn onboarding_key_down(
```
wingleeio, line 7181: **Tab is never handled, so keyboard-only users cannot reach Continue.** GPUI does not move focus on Tab by itself — it only exposes `window.focus_next()` / `focus_prev()`, and nothing in the shell binds them. The `tab_stop(true)` handles therefore only matter for the close dialog, which handles Tab explicitly below. Verified: pressing Tab repeatedly leaves the focus ring on Skip; arrows move within a group but never out of it.

Suggestion: at the top of this function, when `event.keystroke.key == "tab"` and no ctrl/alt/platform, call `window.focus_prev(cx)` for shift or `window.focus_next(cx)` otherwise and return `true`. Also worth focusing `control(0)` on the first mount: `restore_mounted_focus` picks `unfocused` when nothing is focused yet, so arrows/Enter do nothing until the user clicks.

crates/ui/src/shell.rs
```diff
@@ -6158,6 +7096,16 @@ impl Shell {
             cx.notify();
             return true;
         }
+        if self.onboarding.active() {
```
wingleeio, line 7131: Escape with the theme dropdown open lands here (capture phase) and opens the "Continue setting up later?" dialog; the `MenuKey::Escape` branch in `onboarding_key_down` never runs. Suggest: `if self.onboarding.theme_menu.is_open() { self.onboarding_close_theme_menu(cx); return true; }` before the close-confirm logic.

crates/ui/src/onboarding.rs
```diff
@@ -0,0 +1,2290 @@
...
+            step.index() + 1
+        ))
+        .size_full()
+        .relative()
+        .track_focus(&ui.focus)
+        .text_color(theme.text)
+        .font_family(theme.font_sans.clone())
+        // Keep the dialog's accessibility tree and tab order isolated. The
+        // underlying journey is remounted when the dialog closes.
+        .when(!ui.close_confirm, |root| root.child(panel).child(title_bar))
```
wingleeio: Unmounting the panel and titlebar while the dialog is up makes the screen go blank behind the scrim (see screenshot), removes the window drag region, and re-runs the step's `fade_quick` on cancel. Overlaying the dialog on top of the journey (keep `panel`/`title_bar` mounted, gate only their focus/handlers) reads much better.

Related: "Keep setting up" hands focus to control 29 (Skip), so `Esc, Space, Space` skips the whole setup. Restoring focus to the control that had it, or to `control(0)`, is safer.

crates/ui/src/shell.rs
```diff
@@ -6180,15 +7128,393 @@ impl Shell {
...
+                            models
+                                .iter()
+                                .find(|model| {
+                                    Some(model.id.as_str())
+                                        == self.onboarding.selected_model.as_deref()
+                                })
+                                .map(|model| model.reasoning_levels.len())
+                        })
+                        .unwrap_or(0);
+                    Some((16, 16 + count.min(9)))
```
wingleeio, line 7377: When the model is Automatic (`selected_model == None`) the rendered reasoning chips come from the harness's `reasoning_levels` (`render_defaults_step`), but this count only looks at the selected model, so it is 0 and the group is just `(16, 16)`; the same happens for Enter at the `(17..26)` branch below. Arrow keys can never reach those chips and Enter on them is a no-op. Suggest sharing one `fn reasoning_levels(&self) -> Vec<ReasoningLevel>` between render and key handling.

crates/ui/src/onboarding.rs
```diff
@@ -0,0 +1,2290 @@
...
+        )
+        .child(div().w(px(18.0)).flex_none().when(selected, |slot| {
+            slot.child(
+                icon(crate::icons::CHECK)
+                    .size(px(14.0))
+                    .text_color(theme.accent),
+            )
+        }))
+    });
+    let theme_menu = popover::popover_card(theme)
```
wingleeio: Nit: `theme_menu` (and the `Theme::for_selection` sample per variant) is built on every render even while the menu is closed; only `when_some(ui.theme_menu.get(), …)` uses it. Building it inside that closure keeps the Appearance step cheap.

crates/ui/src/onboarding.rs
```diff
@@ -0,0 +1,2290 @@
...
+            .min_h_0()
+            .overflow_y_scroll()
+            .track_scroll(handle)
+            .child(content),
+    )
+    .fade_overflow_y(handle)
+    .into_any_element()
+}
+
+fn harness_status(
```
wingleeio, line 725: Readiness for Claude/Codex/Cursor requires an active OAuth account. Users authenticating with `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` (no `oauthAccount` in `~/.claude.json`, no `auth.json`) never become Ready, Continue stays disabled, and the only way forward is Skip — which abandons Defaults, Titles and Project too. Consider treating installed + enabled as sufficient with the sign-in status shown as advisory, or make Skip step-local.

crates/ui/src/onboarding.rs
```diff
@@ -0,0 +1,2290 @@
...
+        .child(decision);
+    let decision_content = crate::motion::fade_quick(
+        SharedString::from(format!("onboarding-step-{}", step.index())),
+        decision_content,
+    );
+    let journey = div()
+        .w_full()
+        .h_full()
+        .max_w(px(560.0))
+        .max_h(px(720.0))
```
wingleeio: Nit: with `h_full()` + `max_h(720)` the card is always full height, so short steps (Workspace, Project) show a large empty band between the choices and Continue. Sizing to content with a `min_h` would look tighter on laptop viewports.

wingleeio (changes requested):
Thanks — this is a big, careful piece of work and the journey itself holds together well. I built the head (`d36d637`) and ran it on Linux (sway headless): all `ZERON_OPEN_ROUTE=onboarding/*` fixtures, plus two complete real first-run passes on empty profiles (Claude/Codex/Cursor/Devin/Grok/OpenCode CLIs installed), a relaunch on the finished profile, and **Show onboarding** from the user menu. `cargo test -p zeron-ui --lib` for the onboarding/settings/shell tests: 50 passed (incl. the 7 new onboarding tests). CI is green on the head.

Three things need fixing before this lands, all reproduced on a fresh profile (details inline):

1. **The default agent silently becomes the first harness that needs no sign-in (Devin here), not the first ready one.** Agent list arrives before the account probe, so Claude/Codex/Cursor read "Checking sign-in…" and the auto-pick falls through to Devin; when accounts land, Devin is "ready" so it is kept, and Harnesses→Continue persists it into `composer-defaults.json`. Reproduced on both runs.
2. **"Continue without a project" is lost on Finish.** `onboarding_remember_target` writes `noProject: true` to disk, but `Pickers` keeps its own in-memory `ComposerDefaults` and its next `save_defaults()` (catalog labels) rewrites the file with `noProject: false, device: null`. Relaunching the finished profile lands on "Add a project to get started" with no composer — exactly the state the step promised to avoid.
3. **Keyboard-only users cannot advance.** Nothing binds Tab: GPUI only moves focus when the app calls `focus_next`/`focus_prev`, and the only Tab handling here is inside the close dialog. Arrows move within a choice group and Enter/Space activate, but focus never reaches Continue.

Smaller items (inline): Escape inside the theme dropdown opens the "Continue later?" dialog instead of closing the menu; the dialog unmounts the whole journey and titlebar (blank screen behind the scrim, step re-animates on cancel, and "Keep setting up" hands focus to Skip); reasoning chips are unreachable by keyboard when the model is Automatic; the theme menu is rebuilt every frame while closed; OAuth-only readiness locks API-key users out of Continue.

Not exercised here: the Sync sign-in transition (needs the edge) and the macOS titlebar.

### Screenshots (Linux, 1280×800 @2x)

| Workspace (fresh profile) | Appearance after picking Dark |
|---|---|
| Workspace step | Appearance step, dark |

| Theme menu | Coding agents (real detection) |
|---|---|
| Theme menu open | Agents step |

| Session defaults — Devin auto-picked (bug 1) | Session titles |
|---|---|
| Defaults step with Devin selected | Titles step |

| Project — "Continue without a project" | Landing after Finish |
|---|---|
| Project step | Composer after finish |

| Relaunch of the finished profile (bug 2) | Escape inside theme menu (dialog replaces journey) |
|---|---|
| Relaunch lands on Add a project | Continue later dialog |

| `onboarding/agents-error` fixture | `onboarding/narrow` at 600×600 |
|---|---|
| Agents error fixture | Narrow fixture |

crates/ui/src/shell.rs
```diff
@@ -6158,6 +7053,19 @@ impl Shell {
...
+        if self.onboarding.active() {
+            if self.onboarding.theme_menu.is_open() {
+                self.onboarding_close_theme_menu(cx);
+                return true;
+            }
+            if self.onboarding.close_confirm {
+                self.onboarding_cancel_close(window, cx);
+            } else {
+                self.onboarding_request_close(window, cx);
+                self.onboarding_focus_control(26, window, cx);
```
wingleeio, line 7140: Focus set here does not survive to the next key press. Sequence observed on a fresh profile (Appearance step, Dark picked): Escape → dialog opens; Space → the "System" chip (control 0) is activated and the dialog stays open; a second Escape closes it. Reproduced twice, screenshot in the review body.

Two things combine: the deferred `restore_mounted_focus` in `render` runs after this handler and moves focus back to control 0 (the dialog's handles were not in the tree when the check ran / `preferred_focus` still points at control 0 while `close_confirm` is set), and GPUI synthesizes a click for Space/Enter on whatever is focused (`elements/div.rs`, `is_activation_key`), so the chip's `on_click` fires through the scrim. With Continue/Finish focused when Escape is pressed, Space would advance the step under the modal.

Suggestion: while `close_confirm` is true make `preferred_focus` the dialog's `control(26)` (and skip `restore_mounted_focus` for the onboarding root), and in `onboarding_key_down` return `true` + `stop_propagation` for any activation key while the dialog is open, regardless of which control is focused.

crates/ui/src/shell.rs
```diff
@@ -8708,9 +9979,26 @@ impl Render for Shell {
...
+        let preferred_focus =
+            if self.onboarding.active() && self.onboarding.step() != OnboardingStep::FirstSession {
+                self.onboarding.control(0).clone()
+            } else {
+                self.composer.focus_handle(cx)
+            };
         window.defer(cx, move |window, cx| {
-            restore_mounted_focus(&shortcut_focus, &preferred_focus, &unfocused, window, cx);
+            if onboarding_needs_initial_focus && window.focused(cx).is_none() {
+                window.focus(&preferred_focus, cx);
```
wingleeio: This branch is not reached in practice: right after launch, Right/Space are no-ops and the first Tab lands on control 0, i.e. nothing was focused (`window.focused()` is already `Some` on the first Ready frame — the gate/root handle — so the `is_none()` guard fails and `focus_initialized` is consumed). Either drop the guard when `onboarding_needs_initial_focus` is set, or check `!root.contains_focused(...)`/`focused == unfocused` instead of `is_none()`. Not blocking on its own since Tab recovers, but the comment above promises focus on the first frame.

wingleeio (changes requested):
Re-checked head `6d32948` the same way: rebuilt, ran the onboarding/settings/pickers/shell unit tests (53 passed, incl. the 3 new ones), and drove two more fresh-profile passes on Linux (sway headless) plus a relaunch of the finished profile. CI is green on the head.

**Fixed and verified:**
- Default agent: Defaults now opens on Claude Code and `composer-defaults.json` has `"harness": "claude-code"` (was Devin).
- "Continue without a project" survives Finish and relaunch: `noProject: true` stays set through the later catalog save (labels 7→437) and the relaunched app lands on the projectless composer.
- Tab/Shift+Tab move focus through the journey; Tab, Tab, Tab, Space from boot reaches Continue and advances, arrows move within a group, Space/Enter activate. No double activation from GPUI's synthesized keyboard click.
- Escape inside the theme dropdown now closes only the dropdown; the leave dialog overlays the journey instead of replacing it; the Workspace/Project cards are compact.

**Two keyboard items remain (inline):**
1. After Escape opens the leave dialog, focus does not end up on "Keep setting up". Pressing Space activates the control *behind* the scrim (here the "System" appearance chip) and the dialog stays open — GPUI synthesizes a click on the focused element for Space/Enter (`div.rs` `is_activation_key`). If Continue/Finish had focus when Escape was pressed, Space would advance the step under the modal. Reproduced twice.
2. Nothing is focused at boot: Right/Space right after launch are no-ops and the first Tab lands on control 0, so the new `focus_initialized` path is not taking effect (`window.focused()` is already `Some` on that frame, so it falls through to `restore_mounted_focus`). Functionally fine since Tab recovers, but the code does not do what it says.

### Screenshots (head 6d32948, Linux 1280×800 @2x)

| Workspace (compact card) | Escape closed only the theme menu |
|---|---|
| Workspace step | Appearance dark, menu closed |

| Leave dialog overlaying the journey | Space on the dialog picked "System" behind it (item 1) |
|---|---|
| Leave dialog | Dialog open, System selected |

| Coding agents | Session defaults — Claude Code (fixed) |
|---|---|
| Agents step | Defaults step |

| Project — no project | Relaunch: projectless composer restored (fixed) |
|---|---|
| Project step | Relaunch |

crates/ui/src/shell.rs
```diff
@@ -6170,6 +6990,19 @@ impl Shell {
...
+        if self.onboarding.active() {
+            if self.onboarding.theme_menu.is_open() {
+                self.onboarding_close_theme_menu(cx);
+                return true;
+            }
+            if self.onboarding.close_confirm {
+                self.onboarding_cancel_close(window, cx);
+            } else {
+                self.onboarding_request_close(window, cx);
+                self.onboarding_focus_control(26, window, cx);
```
wingleeio, line 7140: Still reproducible on `029e94e`: after this focuses control 26, the deferred `restore_mounted_focus` in `render` moves focus back to `preferred_focus` (control 0), and GPUI's synthesized keyboard click (`elements/div.rs`, `is_activation_key`) fires the focused chip's `on_click` through the scrim. Observed: Escape → dialog; Space → appearance flips dark → system, dialog stays open; Escape → dialog closes, focus lands on the theme trigger (the saved `close_return_focus`).

Suggested fix: while `close_confirm` is set, make `preferred_focus` the dialog's `control(26)` (and skip `restore_mounted_focus` for the journey), and in `onboarding_key_down` swallow any activation key while the dialog is open regardless of which control is focused.

crates/ui/src/shell.rs
```diff
@@ -8725,9 +9962,26 @@ impl Render for Shell {
...
+        }
+        let preferred_focus =
+            if self.onboarding.active() && self.onboarding.step() != OnboardingStep::FirstSession {
+                self.onboarding.control(0).clone()
+            } else {
+                self.composer.focus_handle(cx)
+            };
         window.defer(cx, move |window, cx| {
-            restore_mounted_focus(&shortcut_focus, &preferred_focus, &unfocused, window, cx);
+            if onboarding_needs_initial_focus && window.focused(cx).is_none() {
```
wingleeio: Unchanged since round 2 and still not reached: Right/Space immediately after launch do nothing and the first Tab lands on control 0, so `window.focused()` is already `Some` on this frame and the branch is skipped. Dropping the `is_none()` guard when `onboarding_needs_initial_focus` is set (or comparing against `unfocused`/root instead) would make it do what the comment says.

wingleeio (changes requested):
Round 3 on head `029e94e` (ten new commits incl. the upstream merge): rebuilt, ran the onboarding/settings/pickers/shell tests (58 passed, incl. the new discovery, model-catalog and keyboard-isolation tests), and drove another fresh-profile pass plus a relaunch on Linux (sway headless).

**Holding up on this head:**
- Defaults opens on Claude Code; picking "Fable 5.1" lands in `composer-defaults.json` under `modelByHarness`.
- "Continue without a project" survives Finish (labels 7→437) and relaunch; the relaunched app opens on the projectless composer with Fable 5.1 selected.
- Tab/Shift+Tab, arrows, Space/Enter all work; Escape closes only the theme dropdown; the leave dialog overlays the journey; Finish without a choice shows the inline error.
- The new cardless layout, footer navigation, titlebar Skip, compact Workspace/Project/Titles heights, and the scrollable full model catalog all render correctly.

**Still open from round 2 (neither site changed in these commits):**
1. Space on the freshly opened leave dialog still activates the control behind the scrim and leaves the dialog open — Escape → dialog; Space → "System" appearance chip selected (dark → system), dialog still up; Escape again closes it and focus lands on the theme trigger. Reproduced again, screenshot below.
2. Nothing is focused at boot: Right/Space right after launch are no-ops; the first Tab lands on control 0.

**Process:** the branch now conflicts with `main` in `crates/ui/src/shell.rs` (#216 and #320 landed after your upstream merge), so GitHub did not run CI on `029e94e`; the last CI-green head is `b550b57f`, which differs only by the one-line label commit. Please rebase onto current `main` and let CI run on the rewritten head — that plus item 1 is what I need before approving.

Small visual notes, no action required unless you agree: on Coding agents the list is capped at 6½ rows while the step keeps its 640px height, leaving an empty band above the actions; the "Choose a project or continue without one" error sits flush against the button row.

### Screenshots (head 029e94e, Linux 1280×800 @2x)

| Workspace (cardless, titlebar Skip) | Appearance, Escape closed only the menu |
|---|---|
| Workspace | Appearance dark |

| Leave dialog | Space on the dialog picked "System" behind it (item 1) |
|---|---|
| Leave dialog | Dialog open, System selected |

| Coding agents | Session defaults — full catalog, Fable 5.1 picked |
|---|---|
| Agents | Defaults |

| Session titles | Project — Finish without a choice |
|---|---|
| Titles | Project error |

| Project — "Continue without a project" | Relaunch: projectless composer, Fable 5.1 |
|---|---|
| Project none | Relaunch |

gaelcado:
Updated to `7d58d002420b36ffb87d70ecbbe9829690ec62be`, rebased on current `main` (`30a9a953`, fetched immediately before the lease-protected push). GitHub reports **mergeable**. `d63d5dd9` is preserved on a local follow-up branch; only its necessary obsolete-field compilation fix remains here.

The dialog now consumes activation during capture, claims control 26 after mounting, preserves control 27 through Tab, and restores the exact prior focus. Boot focus claims control 0 even when the root/gate owns focus and waits until the control mounts. Defaults Continue uses a separate handle so it cannot collide with dialog control 26.

Added frame-aware GPUI regressions with key-down and key-up dispatch:
- `onboarding_boot_focus_and_immediate_arrow_space`
- `onboarding_modal_mount_tab_trap_and_exact_cancel_focus`
- `onboarding_modal_consumes_activation_with_stale_background_focus`
- `onboarding_modal_continue_later_persists`

Local validation: **896 UI tests passed**; `cargo check --release --locked -p zeron` passed. Native macOS checks passed for the keyboard-only journey through Finish; dialogs from an appearance chip, Continue, Defaults Continue, and Finish; Tab trapping/cancel restoration; theme-menu Escape; persisted Continue later; relaunch with the projectless composer and Fable 5.1; Show onboarding; Skip/traffic-light spacing; and 600×600 / 1280×800 fixtures.

All five jobs in [fresh CI](https://github.com/zeronsh/zeron/actions/runs/34759907756) passed on this head. Linux UI CI passed **923 tests**, including all four new regressions.

Remaining validation gaps: repository-wide `cargo fmt --all -- --check` flags ten untouched files already in main, which I kept outside this patch; native Linux fixtures were not rerun because the available Linux environment lacks the required build dependencies. @wingleeio, requesting a fresh review now that CI is green, with these limits called out explicitly. GitHub denied my formal reviewer request because this account lacks repository permissions.

wingleeio (approved):
Security audit and review of head `7d58d00`.

**Scope checked:** 7 files, +4633/−44, all under `crates/ui/src/`. No `Cargo.toml`/`Cargo.lock`, CI workflow, build-script, or shell-script changes.

**Findings — nothing malicious, nothing that widens the attack surface:**
- No new network, process, or filesystem primitives in non-test code. The only `std::fs` writes are in `#[cfg(test)]` modules against `tempfile` dirs. The only `include_bytes!` is the existing `apps/landing/public/assets/zeron-app-icon.png`.
- The only env var read is the pre-existing `ZERON_OPEN_ROUTE` dev knob; the PR adds `onboarding/*` routes to it. Fixtures are display-only: every engine call in the onboarding path short-circuits on `fixture.is_some()`, so they cannot mutate a real profile.
- Engine RPCs used are all pre-existing methods (`LIST_HARNESSES`, `LIST_MODELS`, `LIST_AGENT_ACCOUNTS`, `SET_HARNESS_ENABLED`, `GET_TITLE_SETTINGS`, `SET_TITLE_SETTINGS`). No new RPC surface, no protocol changes.
- The Workspace step only routes through the existing `start_sign_in` / `request_sign_out` transitions. No token, credential, or keychain handling; no upload of local sessions.
- Persistence is limited to the new `UiSettings.onboarding` field (serde `default`, so existing settings files deserialize as `Completed`). `is_fresh_profile()` only tests for file existence and never deletes anything; the early `save()` on a fresh dir only writes the settings file.
- `set_target_device` visibility widened to `pub(crate)` and a `stop_propagation()` on Tab in the spaces menu are the only touches outside onboarding code.

**Built and run on Linux (Xvfb 1280×800, isolated temp `ZERON_DATA_DIR` per launch):**
- `cargo build -p zeron --locked` OK.
- `cargo test -p zeron-ui onboarding::tests --lib` — 13 passed.
- `cargo test -p zeron-ui --lib` — 923 passed.
- A genuinely empty data dir with no `ZERON_OPEN_ROUTE` auto-opens the Workspace step (pixel-identical to the `onboarding` fixture below).

### Screenshots (head 7d58d00, Linux 1280×800)

| `onboarding` (= fresh profile) | `onboarding/workspace-sync` |
|---|---|
| Workspace step | Workspace step, Sync devices |

| `onboarding/appearance` | `onboarding/agents` |
|---|---|
| Appearance step | Coding agents step |

| `onboarding/agents-error` | `onboarding/defaults` |
|---|---|
| Agents error fixture | Session defaults step |

| `onboarding/titles` | `onboarding/project` |
|---|---|
| Session titles step | Project step |

| `onboarding/projectless` | `onboarding/narrow` (900×780, window min-width clamp) |
|---|---|
| Project step, no project | Narrow fixture |

Safe to merge from a security standpoint.

PR #338, opened 2026-09-12
Remember composer model options for new chats

neo773:
## Summary
- Model options like context window and fast mode reset on every new chat. They only lived in the draft config, and the draft is cleared whenever the selected chat changes.
- They're now saved per harness in `composer-defaults.json`, next to the remembered model and reasoning.
- Before sending, saved picks the selected model doesn't offer are dropped. Harnesses apply some options without checking: Claude appends `[1m]` to any model id when `contextWindow` is `1m`.
- Existing chats behave as before. They use the options stored on the chat, and changes there don't update the new-chat defaults. This matches how reasoning works.

## Testing
- `cargo test -p zeron-ui --lib`: 805 passed. Added coverage for the new defaults field round-trip and the `offered_options` filter.
- Not checked by hand in the running app yet.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/338"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791843863&installation_model_id=425430&pr_number=338&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F338&signature=77f7bd9e981ff05a35ea2b5cb5274d114a333cf03fa0fff931cd8b3eb452a238"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


crates/ui/src/pickers.rs
```diff
@@ -761,16 +776,29 @@ impl Pickers {
...
+        let remembered = self
+            .defaults
+            .model_options_by_harness
+            .get(&harness)
+            .cloned()
+            .unwrap_or_default();
+        match self.selected_model(cx) {
+            Some(model) => offered_options(model, remembered),
+            // Catalog not loaded yet: nothing to filter against.
+            None => remembered,
```
wingleeio: [P2] Do not send unvalidated per-harness options when the catalog is unavailable

Reproduction: on the new-chat canvas, select a model offering 1M context and enable it, then select Haiku. `pick_model` persists Haiku but leaves `contextWindow: "1m"` in the per-harness memory; with the catalog loaded, the filter correctly hides that selection. Restart the app (or change projects, which clears the model cache), then send before the catalog finishes loading or while its request is failing. This branch now returns the stale 1M selection. `Composer::send_blocked` explicitly allows sends during loading/errors, `resolved()` includes these options in both the new chat config and run request, and the Claude harness blindly appends `[1m]`, producing `claude-haiku-4-5[1m]` even though Haiku does not offer that option. The invalid selection also survives later catalog loads because existing-chat options bypass this filter. Previously restarting cleared the draft options, so this is a new persistence-related failure.

Keep the sticky memory, but ensure the outgoing options are validated for the actual model even without a live catalog (for example, use model-scoped validated metadata, or safely omit unvalidated picks until they can be checked). Add coverage through `Pickers::resolved` for the model-switch followed by missing/loading/error catalog case; the pure `offered_options` test does not reach this fallback.

wingleeio (changes requested):
Reviewed the full diff and traced persistence, model switching, catalog invalidation, and the send path. One correctness issue needs addressing before merge: the catalog-unavailable branch bypasses compatibility filtering and can persist unsupported options into a new chat. This finding is based on the code paths; I have not exercised it interactively.

neo773:
Fixed in the latest commit:
- Remembered option picks are now saved per model, not per harness. A pick is only stored after checking it against the model that offered it.
- `resolved()` reads back the picks for the exact model id it sends. With no catalog (idle, loading, or error), Haiku gets Haiku's picks, which is none, so Opus's 1M no longer comes along.
- Switching models in an existing chat now also drops options the new model doesn't offer.
- New test `remembered_options_stay_valid_for_the_sent_model_without_a_catalog` goes through `Pickers::resolved`: pick 1M on Opus, switch to Haiku, clear the draft, then check the idle, loading, and error catalog states.

wingleeio (approved):
Re-reviewed the complete change and follow-up fix at 21ec2084. The previous blocker is addressed: remembered options are isolated by harness and exact model id, and resolved() reads the options for the same model it sends. The new regression exercises the Opus-to-Haiku switch with idle, loading, and failed catalogs while retaining the original Opus preference. Existing-chat model changes also filter incompatible options. No remaining blocking findings. Merge will wait for all CI jobs on this revision to pass.

PR #341, opened 2026-09-13
feat(ui): add new-thread backgrounds, composer handoff, and UI polish

gaelcado:
## Summary

Add device-local artwork to the new-thread screen and keep the same composer instance through navigation into an established thread, with related titlebar, selector, resize, and message-queue polish.

## Renderer dependency

Depends on https://github.com/zeronsh/zui/pull/9. This head pins its published commit `ce4a177684f1af929eefc6bfba7a44b208c9eec7` from the public fork so CI and reviewers can build it without local path overrides. The two Cargo patches keep `gpui-base` on the same GPUI types. Move the dependency back to upstream after the renderer change is accepted and the GPUI dependency pins are aligned; merge the renderer change first.

## Behavior

- Center the new-thread composer beneath floating titlebar controls; retain input focus and staged attachments through navigation.
- Add Appearance controls for choosing, replacing, and removing local artwork, with Dither, ASCII, Halftone, and Scanlines treatments.
- Validate the exact replacement bytes with the image decoder before changing settings or retiring the previous managed image. Unsupported/damaged files leave the existing background and saved settings untouched.
- Feather artwork around the measured rounded composer with a paint-time GPU alpha mask. Reuse the source/effect texture as geometry changes, removing asynchronous geometry-baked mask work. Read current composer geometry during paint, after prepaint, including the first visible frame.
- Preserve source color beneath ASCII and halftone textures. Prewarm image decoding/effects off-thread and gently fade in cold results without restarting warm artwork on navigation.
- Keep the artwork crop stable when the right panel changes. Fade through conversation-column changes, switch geometry while hidden, and retain the outgoing transcript width until hidden to prevent reflow flashes.
- Preserve the vertical glide when the column does not change, respect reduced motion, and keep transcript clearance and the scroll-to-bottom control stable.
- Restore floating destination selectors and align checkout, branch, pull-request, and usage controls.
- Expand pane resize targets, retain drag feedback, add rounded boundary feedback, update the macOS browser fixture, and restore the compact message-queue tray.

## Settings and compatibility

Artwork is copied into the device-local data directory. Existing settings default to no background and no effect. Effect selection survives shell geometry saves. No engine or protocol migration is required.

## Validation

- `cargo build --locked -p zeron` passes against the published renderer revision.
- `cargo test --locked -p zeron-ui --lib --no-fail-fast`: **913 passed** against that revision, including invalid replacement preservation, cold/warm loading, same-frame geometry, and pane-handoff regressions.
- Companion renderer: **16 Metal tests pass**, including actual offscreen GPU/reference comparisons and atlas reuse; WGSL validation and host/shader layout tests pass.
- Scoped formatting and `git diff --check` pass.

Earlier visual refinements were exercised by the user, but the latest GPU-mask transition has not received a fresh live-app recording or visual sign-off. Windows HLSL is updated in the companion PR but was not run on Windows locally. Remote CI results are reported by the checks on this head, not inferred from local tests.


avarayr:
imo the group segment lines from the left are animated nicely, but it adds a bit too much clutter to the minimal aesthetic.

gaelcado:
CI investigation for head `db052cc9`:

- [Linux browser CI](https://github.com/zeronsh/zeron/actions/runs/34788391038/job/103807981944) fails with `browser input did not return after menu dismissal`. macOS, iOS, UI and session-sync jobs all passed.
- Script order and the two preceding successful fixture recordings place the failure in the forced-OpenGL X11 execution; default X11 and Wayland passed the same binary.
- The Linux fixture, browser implementation, popup lifecycle and Linux runner script are unchanged by this PR. However, this PR does change shell/layout/resize behavior, so a contribution-related regression is **not ruled out**.
- One timing hypothesis: the fixture waits 150 ms after menu dismissal, while the existing popup teardown uses a 100 ms animation plus a 20 ms reap delay before requesting a redraw. That leaves a narrow margin for event processing/repaint. This has not been reproduced or proven as the cause.

The [previous head passed Linux](https://github.com/zeronsh/zeron/actions/runs/34766490488/job/103748276235), but it had different source. Main also has a [window-visibility failure](https://github.com/zeronsh/zeron/actions/runs/34784358803/job/103797016142); that different assertion is not evidence of the same root cause here.

Conclusion: intermittency is plausible, but this failure should not yet be dismissed as unrelated to the PR. A same-SHA rerun and/or inspection of captured input/overlay state would help. No reruns or code changes were made during this investigation. Is this input-restoration check known to be timing-sensitive?

For clarity, the earlier macOS resize-fixture failure was a genuine expectation mismatch after the hitbox change, fixed in `f8b3ceeb`; it was not classified as flakiness.

crates/ui/src/settings.rs
```diff
@@ -224,6 +275,110 @@ pub fn current(cx: &App) -> UiSettings {
...
+    // updates memory first and only logs an I/O failure; for a file-backed
+    // setting that order can leave disk pointing at an image we just deleted.
+    if next.save(&data_dir).is_err() {
+        let _ = std::fs::remove_file(&destination);
+        return Err(
+            "Unable to save the image. Check folder permissions and try again.".to_string(),
+        );
+    }
+    replace(next, SavePolicy::Immediate, cx);
+    remove_managed_new_thread_background(previous.as_ref(), &backgrounds_dir);
```
wingleeio, line 335: [P2] Validate that a replacement can render before retiring the old background

Choose a working PNG background, then replace it with a valid SVG. `attachments::stage_file` accepts SVG (and checks only extension/size, not decodability), so this function persists the new path, deletes the previous managed image, and returns success. However, `background_luminance` uses `image::ImageReader::decode`, which cannot decode SVG; the new-thread screen becomes blank without an error. A focused GPUI reproduction on this head reports `replacement accepted=true, background renderable=false, previous file retained=false`. Corrupt raster files have the same problem. Decode/validate using the actual background rendering path before committing the replacement; either support SVG rasterization or reject unsupported formats with an error while retaining the previous setting and file. Please add a replacement regression covering this case.

wingleeio (changes requested):
Found a reproducible background replacement bug that should be fixed before merge. The existing full UI suite passes locally (935 tests); a focused replacement regression fails. The Linux browser failure is being rerun separately and is not the basis of this finding.

wingleeio (approved):
Reviewed the replacement validation and GPU mask fixes, plus the official dependency alignment at df98d547. The invalid-image regression is fixed: decoding accepts the exact bytes before persistence or removal of the previous managed image. Same-frame composer measurements drive the GPU mask. Zui aa009411 has the same source tree as the reviewed renderer revision; gpui-base 94c1bbaf changes only the Zui pins from the previously used revision. No blocking findings. Validation: 940 UI tests passed on the updated PR and 962 passed integrated with current main. Fresh platform CI is running before merge.

PR #349, opened 2026-09-13
feat: add Antigravity ACP, improve context UI, and persist window geometry

laurentcodes:
## Summary

This PR adds Antigravity (AGY) through its managed ACP server, improves the context-window indicator, and adds persistence for the desktop window's size and position.

## Changes

- **Antigravity:** Replace the legacy `agy` stream-json integration with Google's pinned ACP server, including managed archive installation and integrity verification. Keep the harness opt-in, sign in before enabling it, and sign out when disabling it. Show setup progress, cancellation, and retry states, and preserve the server's configured authentication method. Resolve named-home and relative `GEMINI_HOME` paths before reading settings, pass the same absolute directory to every ACP child, and reject unresolved paths before authentication.
- **Models and commands:** Group model reasoning variants under their base models, resolve selections against the advertised catalog, and expose Antigravity skills as slash commands while hiding `/logout`. Add a diagnostic ACP probe that sends a live prompt only with `--prompt`.
- **Context-window UI:** Hide the indicator when no positive context capacity is reported, format token counts with thousands separators, and let the tooltip fit its full usage text.
- **Window persistence:** Save normal window bounds through the shared debounced settings store, flush on close and app quit, and restore saved bounds on launch or reopen using a stable display UUID. Recenter on the primary display when the saved monitor is disconnected. Validate saved geometry and constrain it to a display's available area. Preserve geometry when other shell settings are saved, and skip fullscreen and explicitly maximized bounds while preserving ordinary screen-sized macOS windows.

## Validation

- Passed `cargo check --workspace --all-targets`.
- Passed 7 window-geometry tests (including display identity, disconnected monitors, and legacy settings), 3 context-usage tests, and the shell settings-preservation regression test.
- Passed 17 Antigravity harness unit tests and all 39 ACP integration tests, including an isolated subprocess regression for relative auth-settings paths. After final hardening, reran the 17 unit tests and 11 Antigravity integration tests successfully.
- The full harness unit run passed 170 tests; the previously observed `shell_env::unix::tests::falls_back_when_interactive_attempt_hangs` test failed. Native Windows tests were not run.
- Passed `git diff --check`.
- Native multi-monitor restoration, live browser sign-in, and the iOS build were not exercised in this review. The full workspace test suite was not rerun.


wingleeio (changes requested):
Security and correctness review of dbb8681d95c2542cc04a6c8612e7480d21fd7a6b: requesting changes. I found no evidence of malicious contributor code, but the authentication and model-selection behavior is not ready to merge.

1. **[P1] Preserve the configured authentication boundary** — `crates/harness/src/acp/mod.rs:673-676, 933-943`. Enabling always calls `authenticate` with `oauth-personal`. The comment says an existing settings auth method is respected, but the actual pinned Google server does the opposite: its `authenticate` explicitly selects the requested method and persists it with `_remember_auth_type`. A previously configured `oauth-business`, `agent-platform`, or `gemini-api-key` user is forced into consumer OAuth, and after a successful login subsequent work uses that identity/backend. Respect an already configured usable method, or expose an explicit auth-method selection; do not silently replace enterprise/API-key configuration. The downloaded server's initialize response advertises all four methods.

2. **[P2] Fail closed when Antigravity rejects the selected model** — `crates/harness/src/acp/mod.rs:2628-2699`. Antigravity advertises `configOptions`, so `first_class_model_change` skips the strict legacy model path. A rejected model config RPC then falls through the Devin-only error guard and the prompt runs on the default model. I reproduced this with the PR's fixture modified to reject only the model-setting call: requesting Gemini 3.1 Pro produced `SessionStarted { model: "gemini-3.1-pro", ... }`, `TextDelta { text: "sets:mode=yolo;" }`, and `Done { status: Completed, ... }`. A regression asserting `Errored` failed. Google's real server rejects unavailable model IDs with -32602 before replacing the current agent, so this is a realistic stale-catalog/account-change case. Treat explicit Antigravity model-setting failures as fatal before any prompt is sent, and test that no prompt follows a rejection.

3. **[P2] Surface and retain retryability of sign-out failures** — `crates/engine/src/rpc.rs:1199-1214`. The enabled preference is persisted as false before logout is attempted. If the server cannot launch, times out, or rejects logout, the error is only logged and the RPC still returns success while credentials remain. Repeating disable cannot retry because `was_enabled` is now false. This contradicts the toggle's advertised sign-out behavior. Keep disabling possible, but return/display an explicit sign-out failure and retain a retry path; do not imply credentials were cleared. Add failure-path coverage.

Catalogue investigation:
- Downloaded official CLI 1.2.5 using its official installer manifest and verified its SHA-512. In an isolated profile with networking disabled and a noncredential placeholder for API-key mode, `agy models` advertised Gemini 3.8 Flash high/medium/low, 3.7 Flash high/medium/low, 3.6 Flash high/medium/low, and 3.1 Pro high/low. No inference request or real-account login was performed.
- Downloaded the exact ACP 1.1.1 archive pinned here, confirmed its URL/version against the ACP registry, inspected its bundled source, and ran initialize/session/new in the same kind of isolated offline profile. That server advertises 3.7 Flash as its default, with 3.6/3.5 Flash and 3.1 Pro also in its bundled API-key catalogue. Its personal-OAuth list is fetched dynamically and filtered to Gemini models.
- Recommendation: keep 3.7 Flash + 3.1 Pro as the conservative static fallback for this pinned ACP version. Prefer 3.8 Flash when the live ACP catalogue actually advertises it; do not copy the newer CLI-only fallback into ACP 1.1.1 without validating support. Do not add Claude/GPT-OSS simply because the broader Antigravity product offers them. Account-specific OAuth entitlements were not verified.
- Official references: https://github.com/agentclientprotocol/registry/blob/main/antigravity-acp/agent.json and https://antigravity.google/docs/cli/headless/ .

Additional installer hardening: the new managed archive path checks HTTP status and executable existence but has no pinned digest/signature verification, download size limit, or explicit archive-entry/link validation. The URLs are official and I did not find a malicious endpoint; adding integrity verification and bounded extraction would strengthen this new executable-download boundary.

Validation: 122 harness unit tests, 35 ACP integration tests, 13 registry tests, and 16 account tests passed (186 total); diff whitespace check passed. The additional adversarial model-selection test failed as described. Temporary test changes were restored. I did not approve the held CI runs or run the full cross-platform suite because the findings already block approval. GitHub also reports merge conflicts with main, which need resolving before final integration testing.

crates/harness/src/acp/mod.rs
```diff
@@ -428,14 +461,304 @@ fn pi_spec() -> AcpAgentSpec {
...
+                home.join("antigravity-cli").join("skills"),
+            ]
+        })
+        .unwrap_or_default()
+}
+
+fn antigravity_home() -> Option<PathBuf> {
+    std::env::var_os("GEMINI_HOME")
+        .filter(|home| !home.is_empty())
+        .map(PathBuf::from)
```
wingleeio: [P1] Expand GEMINI_HOME the same way as the server before reading auth settings

The pinned server resolves `GEMINI_HOME` with `os.path.expanduser`, but this helper keeps a value such as `~/.gemini` literal. Consequently Zeron looks for `./~/.gemini/antigravity-acp/settings.json`, treats the real configured auth as missing, and defaults to `oauth-personal`; the server then expands the same environment variable and persists that personal choice over the real enterprise/API-key settings. I reproduced this with a real temporary home-relative settings directory containing `oauth-business`: `configured_antigravity_auth_method()` returned `None` instead of `Some("oauth-business")`. I also verified the actual pinned server accepts a tilde-prefixed GEMINI_HOME. Normalize the home exactly as the server does and add a settings-file regression, not only a test of the method-selection helper.

crates/harness/src/acp/mod.rs
```diff
@@ -428,14 +461,304 @@ fn pi_spec() -> AcpAgentSpec {
...
+        Ok(settings) => settings,
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => return Ok(None),
+        Err(error) => {
+            return Err(HarnessError::Protocol(format!(
+                "could not read Antigravity auth settings at {}: {error}",
+                path.display()
+            )));
+        }
+    };
+    let settings: Value = serde_json::from_str(&settings).map_err(|error| {
```
wingleeio: [P2] Parse Antigravity settings as Hjson, matching the pinned server

Antigravity 1.1.1 reads this file with `hjson.load`, despite the `.json` suffix. Valid existing settings containing comments or unquoted keys now prevent enabling the harness because this new `serde_json::from_str` rejects them before the agent even starts. I reproduced the failure with `{ // comment followed by newline, then a quoted auth object }` and confirmed the same commented settings file is accepted by an actual offline initialize/session/new probe against the pinned server. Preserve the server-supported syntax (or obtain the resolved auth method through a server-supported mechanism) and cover valid Hjson in the regression tests.

crates/harness/src/acp/mod.rs
```diff
@@ -428,14 +461,304 @@ fn pi_spec() -> AcpAgentSpec {
...
+) -> Result<String, HarnessError> {
+    let available: Vec<&str> = initialized
+        .get("authMethods")
+        .and_then(Value::as_array)
+        .map(Vec::as_slice)
+        .unwrap_or_default()
+        .iter()
+        .filter_map(|method| method.get("id").and_then(Value::as_str))
+        .collect();
+    let selected = configured_method.unwrap_or(default_method);
```
wingleeio: [P2] Canonicalize the supported vertex-ai auth alias before validation

The pinned server explicitly accepts a saved `auth.type` of `vertex-ai` as an alias for `agent-platform`, but initialize advertises only the canonical ID. This exact string comparison rejects that previously valid configuration, so existing Vertex AI users cannot enable Antigravity. A targeted test with configured `vertex-ai` and advertised `agent-platform` fails with "configured auth method vertex-ai is not advertised". Apply the server-supported alias mapping before checking/sending the method, while continuing to reject genuinely unknown configured values.

wingleeio (changes requested):
Re-reviewed c7eb51ebeb23cc5b09b74e16192cc28fb59f4b36, including both follow-up commits and the merge resolution. Requesting changes because the auth-settings reader still disagrees with the real pinned server in three reproducible cases (inline findings).

Resolved from the prior review:
- Rejected Antigravity model selections now fail before prompting; the new integration regression passes.
- Logout failure is surfaced, keeps the harness enabled, and is retryable; the new engine regression passes.
- The installer now pins SHA-512, limits archive/download sizes, and rejects unsafe paths and nonregular archive entries. I independently verified the Linux x86_64 digest against the official archive downloaded in the previous review.
- The branch is now mergeable, and live Antigravity model discovery is refreshed rather than permanently cached.

Validation: 157 harness unit tests + 37 ACP integration tests + the engine sign-out retry test passed (195 total). Three additional targeted auth regressions failed: home-relative GEMINI_HOME loses the configured method, valid commented Hjson fails parsing, and the vertex-ai alias is rejected. An actual offline probe of the exact Google ACP 1.1.1 server accepted both commented Hjson and tilde-prefixed GEMINI_HOME. No real credentials or inference were used. Temporary source changes were restored; the checkout is clean. Full cross-platform CI was not run because the remaining auth issues block approval.

The fallback recommendation is unchanged: keep 3.7 Flash / 3.1 Pro for the pinned ACP 1.1.1 static catalogue, and use 3.8 Flash when the live ACP catalogue advertises it. The newer CLI catalogue alone does not establish support in the pinned ACP server.

crates/harness/src/acp/mod.rs
laurentcodes: Fixed in e92448c1.

Added `expand_user()` (`crates/harness/src/acp/mod.rs:543`), a port of Python's
`os.path.expanduser` for the leading-`~` case, applied in `antigravity_home()`
(:562) so the value is normalized before the settings path is built. Since
`antigravity_skill_dirs()` shares `antigravity_home()`, the skill folders get the
same normalization.

Home resolution is now `user_home()` (:531 — HOME, then USERPROFILE) so the
tilde path and the `.gemini` fallback agree on where home is.

`~user` is deliberately left literal: that form needs the passwd lookup Python
does, and `expanduser` itself returns the path unchanged when that lookup fails.
Noted in the doc comment so it doesn't read as an oversight.

Regressions, both file-backed as asked — reading was split into
`configured_auth_method_in(&Path)` so tests hit real files without touching
process env:
- `antigravity_home_expands_a_home_relative_gemini_home` — `~/x`, bare `~`,
  absolute, `~user`, and the no-home case
- `antigravity_home_relative_settings_still_resolve_the_configured_method` —
  writes `oauth-business` under a home-relative dir and asserts it resolves
  (your exact repro)

crates/harness/src/acp/mod.rs
laurentcodes: Fixed in e92448c1.

Swapped `serde_json::from_str` for `deser_hjson::from_str` into a typed
`AntigravitySettings` struct (`crates/harness/src/acp/mod.rs:628`). New workspace
dep `deser-hjson = "2.2"` (MIT; `serde` is its only transitive dep).

Regressions:
- `antigravity_settings_reader_accepts_hjson` — a file with a `//` comment,
  unquoted keys and an unquoted value, which `serde_json` rejects
- `antigravity_settings_reader_accepts_plain_json` — the plain-JSON form the
  server writes
- `antigravity_settings_reader_reports_no_method_when_unset` — missing file,
  `{}`, and a blank `auth.type`

crates/harness/src/acp/mod.rs
laurentcodes: Fixed in e92448c1.

Added `ANTIGRAVITY_AUTH_ALIASES` (`crates/harness/src/acp/mod.rs:573`) mapping
`vertex-ai` to `agent-platform`. The reader now returns a `ConfiguredAuthMethod`
holding both the literal from the settings file and the canonical id;
validation and `authenticate` use the canonical one, while the error message
still names what the user actually wrote, so a genuinely bad value stays
diagnosable.

Unknown values are still rejected, per your note.

Regressions:
- `antigravity_settings_reader_canonicalizes_vertex_ai` — `vertex-ai` on disk
  gives `configured: "vertex-ai"`, `canonical: "agent-platform"`
- `antigravity_sign_in_resolves_the_vertex_ai_alias` — configured `vertex-ai`
  against an `initialize` advertising only `agent-platform` resolves
- `antigravity_sign_in_still_refuses_an_unknown_configured_method`

crates/harness/src/acp/mod.rs
```diff
@@ -428,14 +461,377 @@ fn pi_spec() -> AcpAgentSpec {
...
+        home.to_path_buf()
+    } else {
+        home.join(rest)
+    }
+}
+
+fn antigravity_home() -> Option<PathBuf> {
+    let home = user_home();
+    match std::env::var_os("GEMINI_HOME").filter(|home| !home.is_empty()) {
+        Some(gemini_home) => Some(expand_user(Path::new(&gemini_home), home.as_deref())),
```
wingleeio: [P2] Resolve the same settings directory as the server before choosing auth

The new expansion fixes ~/..., but valid paths still make the reader miss the configured account and fall back to oauth-personal. I reproduced both cases with an oauth-business settings file: (1) GEMINI_HOME=~ubuntu/<temporary-directory>, which expand_user leaves literal although Python successfully resolves the existing user through the passwd database; and (2) GEMINI_HOME=<relative-directory>, where this reader resolves against Zeron’s cwd but sign_in starts the server in HOME (line 1019). In both cases sign_in_auth_method selected oauth-personal. The actual pinned server also accepted the named-home path in an isolated offline probe. After successful personal sign-in, the server persists that method to the real settings file, replacing the configured enterprise method.

Resolve the directory with the same platform/home/current-directory semantics used by the child, or fail explicitly when a supported path cannot be resolved instead of treating it as missing settings. Keep the reader and child on the same directory. Add end-to-end settings-selection regressions for named-home and relative paths; the current test asserting that all ~user paths remain literal codifies the mismatch. Python’s documented behavior is at https://docs.python.org/3/library/os.path.html#os.path.expanduser .

wingleeio (changes requested):
Re-reviewed fe1ebecb5a8fb25200b007b1dfb9a34ed0878119, including all four commits since c7eb51e and the added window-geometry changes. Requesting changes for one remaining authentication-settings path mismatch; it is reproduced below.

The prior Hjson and vertex-ai alias findings are fixed. The ordinary ~/... case is also fixed. Rejected-model handling, retryable logout, SHA-512 verification before extraction, and archive path/type/size checks remain in place. I inspected the new deser-hjson dependency and the settings/UI persistence changes; I found no additional actionable security issue in those changes. This is a code review, not a guarantee that the implementation has no vulnerabilities.

Validation on this revision:
- 167 harness unit tests and 37 ACP integration tests passed.
- The engine sign-out retry regression passed.
- Seven window-geometry tests and the shell settings-preservation regression passed: 213 existing tests total.
- Additional tests using a real temporary oauth-business settings file passed with absolute and ~/... GEMINI_HOME, but failed with ~ubuntu/... and a relative GEMINI_HOME: the selected method was oauth-personal in both failures.
- The actual pinned Google ACP 1.1.1 binary successfully initialized and opened an offline session with GEMINI_HOME=~ubuntu/.gemini, confirming named-home expansion is supported by the server. This used an isolated filesystem/network namespace and a placeholder API key, with no real credentials or inference.
- Temporary regression source changes were restored and the checkout is clean. Native macOS/Windows tests and full cross-platform CI were not run; approval is already blocked by the reproduced auth issue.

The fallback recommendation remains Gemini 3.7 Flash / 3.1 Pro for the pinned ACP server, with 3.8 Flash used when advertised by its live catalogue. The new catalogue regressions cover that distinction.

crates/harness/src/acp/mod.rs
laurentcodes: fixed

wingleeio (approved):
Re-reviewed and security-audited db623bbda036285eb59ca6442902a05553b7dc74. The remaining authentication-path blocker is fixed: the settings reader and all Antigravity launches now share an absolute GEMINI_HOME, named Unix homes use reentrant passwd lookups, Windows home selection follows its native environment variables, and unresolved paths fail explicitly before authentication. I found no remaining blocking security or correctness issue in the reviewed changes.

Validation on this exact revision: 171 harness unit tests, 39 ACP integration tests, 183 engine unit tests, and 1,054 UI tests passed (1,447 existing tests). I separately exercised the sign-in subprocess with absolute, ~/..., ~username/..., and relative GEMINI_HOME values pointing at a temporary oauth-business settings file; all four preserved business auth and passed the same absolute directory to the child. The new Unix FFI handles lookup errors, checks returned pointers, bounds buffer growth, and copies strings before dropping the backing buffer.

The prior fixes for Hjson settings, the vertex-ai alias, rejected model selection, retryable logout, and verified archive installation remain intact. The earlier independently downloaded official Linux ACP archive matched the pinned SHA-512. Static fallback models remain appropriate for pinned ACP 1.1.1; newer models are accepted when its live catalogue advertises them.

No real-account OAuth or live inference was performed. Native cross-platform CI is still running at approval time; I am holding the merge for its results. The review checkout is clean.

PR #352, opened 2026-09-13
Fix token refresh storms and preserve sessions during transient DNS failures

jsgrrchg:
## Summary

Prevent a temporary DNS or HTTP failure from triggering a token-refresh storm or being reported as a signed-out session. Concurrent consumers now share refresh work and its failure result, transient failures use a bounded account-wide cooldown, and HTTP diagnostics retain the underlying transport cause.

For example, when an expired token is requested by 20 chat/HTTP consumers while DNS is unavailable, the engine performs one refresh attempt during the cooldown and returns the same temporary failure to the callers. It preserves the signed-in session and continues bounded automatic recovery. An actual sign-out still clears credentials and disconnects authenticated transports.

The implementation is split into three independently reviewable commits. **The branch passed a Codex Security diff audit: no introduced vulnerabilities were confirmed across all 16 changed files.** The audit scope and runtime limitations are documented below.

## Problem and incident context

A reported macOS incident affected native hostname resolution outside Zeron: both `curl` and `getaddrinfo()` failed to resolve the edge, while direct DNS queries returned addresses and HTTPS worked when an IP was supplied explicitly. Flushing the system DNS state restored resolution, and the already-running Zeron process recovered without an application restart.

That evidence is consistent with defective native resolver state, but it does not conclusively distinguish a stuck negative cache entry from resolver selection/routing state. It also does not establish that Zeron or a particular VPN caused or prolonged the incident.

The incident exposed three separate application weaknesses:

1. The refresh mutex serialized requests but did not share a failed result. Waiting consumers could each retry the same failing operation immediately after acquiring the mutex.
2. `access_token()` reduced temporary network failures to `None`, which downstream consumers interpreted as signed out even though a persisted session remained.
3. Converting `reqwest::Error` directly to a string discarded its source chain, obscuring DNS, TLS and socket causes.

The reported incident included 304 failed refreshes and a peak of 42 in one second. Those figures motivate the change; they are incident observations, not a benchmark produced by this PR's tests.

### 1. Share refresh attempts and bound retry cooldown

Commit: `e7d6c99a` — `fix(auth): share refresh attempts and bound retry cooldown`

- Share the ordinary in-flight refresh future between token consumers, including the background refresh loop.
- Keep the actual refresh operation serialized, including organization-scoped exchanges.
- Run ordinary refresh work independently of an individual consumer's lifetime. A canceled room connection or caller does not cancel an in-progress exchange that may rotate the refresh token.
- Remember transient refresh failures globally within the shared `Auth` instance. Consumers arriving during the cooldown receive the remembered failure instead of sending another request.
- Use exponential backoff starting at one second, with 0–255 ms jitter and a five-second maximum delay. Successful refreshes reset the failure state.
- Continue automatic retry after cooldown expiry, including when DNS recovers without a network-path event.
- Allow wake/connectivity notifications to interrupt the failure wait. Expose an explicit reset through the optional `retry` parameter on `PROBE_SYNC`.
- Invalidate obsolete cooldown results when a retry or session transition occurs. An explicit organization switch does not reuse another scope's cached failure.
- Fence refresh response application with the session generation and previous refresh-token identity. A late success or rejection cannot restore a signed-out session or overwrite/revoke a replacement session.

### 2. Distinguish sign-out from temporary token unavailability

Commit: `14645f72` — `fix(auth): keep transient token failures distinct from sign-out`

- Change the token-provider contract to return a typed result with `SignedOut` and `TemporarilyUnavailable` errors.
- Propagate the distinction through engine errors, room URL providers, synchronization, device RPC, workspace/chat supervisors and preview signaling.
- Recheck the current stored session and usable cached token after refresh completion, including when an obsolete refresh response was discarded during a session transition.
- Preserve credentials on network failures and retryable HTTP responses such as 408, 429 and server errors. Preserve definitive revocation handling.
- Keep retry supervisors and already-authenticated device connections alive during temporary token unavailability.
- Continue requiring a successfully acquired token before a new authenticated HTTP request or WebSocket dial. Temporary unavailability does not authorize an anonymous fallback.
- Retain sign-out notifications and transport cleanup so a real revocation still closes host and cached peer connections.

The transient distinction belongs to token acquisition; a DNS failure does not rewrite the persisted account session into a signed-out state.

### 3. Preserve DNS and HTTP transport diagnostics

Commit: `2b630ebc` — `fix(engine): preserve DNS and HTTP transport error causes`

- Add a shared HTTP error formatter that retains the destination origin and the underlying error-source chain.
- Remove the top-level request URL before formatting, excluding its user information, path, query and fragment from diagnostics.
- Apply the formatter to authentication, chat/registry HTTP sync, checkpoint and relevant sidecar operations.
- Preserve the last transport error when checkpoint retries are exhausted.
- Add injected DNS-failure tests that verify useful causes survive the RPC/sync error seams and synthetic credentials do not appear in the resulting messages.

This is targeted handling of the affected `reqwest` paths, not a universal sanitizer for every possible future dependency error string.

## Regression coverage

Added or updated tests cover:

- Twenty concurrent consumers sharing one failed refresh during cooldown, including interaction with the background loop and HTTP/room consumers.
- Automatic recovery after cooldown and explicit retry bypassing the failure wait.
- Cancellation of an individual consumer without canceling the shared refresh exchange.
- Sign-out during refresh and stale success/rejection/transient responses after replacement sign-in.
- HTTP 408, 429 and 503 preserving the signed-in session.
- Typed temporary failures propagating across HTTP and room token consumers.
- Workspace synchronization recovering after temporary refresh failure.
- Existing host/cached peer links surviving temporary unavailability, while sign-out still closes them.
- DNS cause retention and exclusion of synthetic URL/token secrets from diagnostics.

### Executed during the security review

```bash
cargo test --locked -p zeron-engine --test auth -p zeron-rpc --test device_room
cargo test --locked -p zeron-engine --lib dns
```

**Results: 32 passed, 0 failed, 1 ignored.** This consists of 16 auth integration tests, 12 device-room integration tests and 4 DNS/diagnostic tests. The ignored device-room test requires a live edge.

The test binaries ran with a minimal runner that changes their working directory to `/tmp` before execution, isolating their loopback listeners from preview discovery by running Zeron instances. The tests use synthetic credentials and local/injected failure fixtures. The list of regression cases above describes the added coverage; the commands here are the tests specifically rerun during the audit.

## Codex Security audit

**Result: passed the branch diff audit with no confirmed introduced vulnerabilities.**

| Item | Audited scope/result |
| --- | --- |
| Baseline | `bdeba13bae740cc45003d0554b8eb4c66176618d` |
| Reviewed head | `2b630ebcf6d2535ae4e6c45564a2febc9e8cbf91` |
| Change coverage | All 16 changed files: 12 production sources and 4 test files |
| Review method | Source review, independent bounded auth/token-consumer reviews, cross-file reconciliation and local regression execution |
| Confirmed introduced vulnerabilities | 0 |
| Deferred branch candidates | 0 |

The review examined refresh concurrency and session-generation fences, stale response handling, sign-out and live-connection cleanup, token requirements before dialing, the local IPC Origin guard, verified user/organization routing at the edge, and credential exposure through expanded diagnostics.

This conclusion applies to the pinned branch diff. It is not a whole-repository security certification. Production deployment settings, real log permissions/audiences, external WorkOS revocation timing and macOS resolver behavior were not independently exercised. The live-edge test remained ignored.

## Behavior and operational limits

- The changes improve request coordination, session classification and recovery diagnostics. They do not repair a defective macOS resolver or guarantee that a system DNS flush will never be necessary.
- A DNS-only recovery may not produce a network-path notification. Bounded periodic retries remain necessary to discover that resolution works again; the five-second cap bounds the cooldown delay, not total end-to-end recovery time.
- Healthy, already-authenticated connections survive temporary token failures. Actual sign-out retains the revocation/cleanup path.
- No automatic DNS flush, system resolver replacement, VPN configuration change, edge authentication change or persistence migration is introduced.
- No production service was contacted as part of the local security regression tests.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/352"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791923453&installation_model_id=425430&pr_number=352&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F352&signature=89ee1c0375af1c919d399ac1e8cb49d43322fcf28a557efe605782ed49858685"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (approved):
Independent safety and network-stability review of 2b630ebcf6d2535ae4e6c45564a2febc9e8cbf91. No blocking introduced security vulnerability or network regression found in the 16-file diff; approving.

Reviewed refresh concurrency and cancellation, single-use refresh-token rotation, retry/cooldown behavior, session-generation and token-identity fences, sign-out and replacement-session races, credential requirements before new HTTP/WebSocket connections, existing authenticated connection retention, and expanded transport diagnostics. The implementation preserves the distinction between temporary token unavailability and actual sign-out. It does not add an anonymous request fallback or change DNS resolution, TLS verification, edge authorization, dependencies, or workflows. Diagnostics remove the top-level request URL before including transport causes; the targeted synthetic-secret/DNS tests pass. This formatter is not a universal sanitizer for arbitrary future nested error messages.

Validation:
- On the PR head, 16 auth + 11 local-first/recovery + 12 device-room integration tests passed; one live-edge test was skipped.
- I independently added and ran a temporary multithreaded sustained-outage regression: 1,980 token requests from batches of 20 concurrent consumers over 10 seconds produced only four failed refresh exchanges. The persisted session remained signed in. Restoring the stub transport, without a Retry action or connectivity notification, recovered automatically in 2.45 seconds with one successful exchange. Subsequent sign-out removed the session file and all 20 callers received SignedOut. The temporary source change was restored.
- I also tested GitHub's merge tree b2033f4bcf304389a3faa09768d84ed6bc98f844, whose parents are current main 787f583117ea772e7484ade808e36a2c4d8e6ba0 and this PR head. 301 tests passed across engine, RPC, sync, and preview; zero failed. Coverage includes injected DNS failures, HTTP 408/429/503, stale refresh results, caller cancellation, workspace recovery, live-link preservation, actual sign-out cleanup, sync convergence, preview proxy/transport, peer reconnection, and teardown/churn.
- The merged-tree commands were `cargo test --locked -p zeron-engine -p zeron-rpc --lib --test auth --test local_first --test device_room` and `cargo test --locked -p zeron-sync -p zeron-preview --features zeron-sync/mock-server`. Test executables ran from /tmp via a runner to isolate their loopback fixtures from repository preview discovery.
- Existing GitHub checks for this head are green, including Linux/macOS networking, session-sync, desktop UI/browser, and iOS. Both review worktrees are clean.

Limits: four merged-tree tests requiring a live edge or local Worker were skipped. No production credentials, real WorkOS exchanges, VPN transitions, or macOS resolver fault injection were exercised. The measured recovery time is from a local fault fixture; the five-second retry cap is a cooldown bound, not an end-to-end network recovery guarantee. These changes improve application recovery but cannot guarantee that an OS resolver or external network remains healthy.

PR #374, opened 2026-09-15
Choose the terminal and code/diff fonts independently

ShayanAbbas1:
The interface font was the only family you could pick. Code, diffs, and the terminal were pinned to Geist Mono at fixed sizes.

Each surface now has its own family **and** its own size. The interface and code/diff pickers offer every installed family; the **terminal picker is restricted to fixed-width families**, because its renderer paints into a cell grid sized by the `m` advance and a proportional face makes click and drag select different cells than the ones under the pointer. Nerd Fonts are fixed-width and qualify.

## Settings

Three rows, one control pair each. The family picker is generalized behind `FontKind { Ui, Terminal, Code }` and the size picker alongside it, so each widget is written once and called three times rather than duplicated.

Every dropdown is searchable.

Appearance settings with Interface, Terminal, and Code and diff font rows

## Terminal and code, rendering independently

MesloLGL Nerd Font in the terminal — powerline glyphs and all — with Dank Mono for code and diffs.

Terminal pane rendering MesloLGL Nerd Font with powerline prompt glyphs

Diff view rendering in Dank Mono

## Implementation

Families and sizes live on `Theme`, so every render site that already had a theme in scope needed no new plumbing.

Each surface keeps its historical baseline size and scales proportionally from the shared setting, so at the default the app renders pixel-identically to `main` (markdown 12.5, diff 12.0, editor 13.0, plain preview 11.5) and nothing shifts until you pick something. `filesEditorFontSize` already existed as a user-settable code size but only reached the workspace file editor; it migrates into `codeFontSize` and now also drives diffs and markdown fences, rather than becoming a second knob for the same thing.

The terminal keeps its ligature-disabling and derives cell metrics from the resolved font via `em_advance`. Families qualify as fixed-width by comparing the real `i`/`m`/`W`/`0` advances from every face in the family, rather than trusting a PANOSE bit or fontdb's `monospaced` flag. A persisted proportional terminal family falls back to Geist Mono. Font-size changes reach the PTY through the same path as a window resize.

## Bugs found along the way

Making the sizes variable exposed two latent defects:

- `TerminalElement::paint` recomputed row height from the constant while `prepaint` used the live value. At any non-default size the cursor and selection quads drifted off the glyph rows, worsening down the screen.
- Pixel-delta (trackpad) scroll divided by the hardcoded row height, moving the wrong number of rows at any non-default size.

And one in the new picker: Enter committed the font correctly, but the event kept bubbling to the trigger, whose closed-state guard read the now-closed menu and reopened it on the same keystroke. The handler now stops propagation only for keys it actually consumes — blanket-swallowing would have broken typing in the search box.

## Testing

`cargo check --workspace --bins` clean; 966 lib tests passing. New coverage for the mono size ladder, size-dropdown labelling, search filtering with navigation confined to matches, the `filesEditorFontSize` → `codeFontSize` migration, downgrade safety against unknown keys, and mono-preferring fallback when a device-local family disappears.

Verified in the app, as shown above.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792045204&installation_model_id=425430&pr_number=374&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F374&signature=e64efa8e0689040c303eafd71fada5de3ab1011e3284067b9352d73ab7bc0c81"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


crates/ui/src/terminal/view.rs
```diff
@@ -430,21 +436,21 @@ impl gpui::Element for TerminalElement {
         // to the pty (the command runs), only the painted run lost a cell.
         // The landing page disables the same three features on its ASCII art
         // for the same reason.
-        let mut mono = font(theme.font_mono.clone());
+        let mut mono = font(theme.font_terminal.clone());
```
wingleeio, line 439: [P2] Preserve terminal cell positioning before accepting proportional fonts

This makes proportional families such as Geist/System UI available to the terminal, but shape_row still batches consecutive ASCII into naturally shaped runs (its comment explicitly assumes monospace). paint anchors only the start of each run, whereas cursor, selection and cell_at use m-width × column. Measuring em_advance once does not make the other glyph advances equal.

Metric reproduction with installed DejaVu Sans at 13 px: m is 12.664 px, while "iiiiiiii" shapes to 28.895 px; the grid reserves 101.309 px. The cursor/selection boundary is therefore 72.414 px away from the painted text. Clicking or dragging over terminal output can target/copy different cells than those visibly under the pointer, and runs can be displaced around column-pinned non-ASCII segments. This is a display/selection integrity problem, not evidence of arbitrary command execution.

Either constrain terminal choices to fonts satisfying the renderer's fixed-width assumptions, or place glyphs/clusters into their actual terminal cells, including bold/italic and fallback faces. Add a renderer/selection regression using a proportional font and mixed ASCII/non-ASCII output.

crates/ui/src/typography.rs
```diff
@@ -379,27 +477,122 @@ pub fn set_family(family: UiFontFamily, cx: &mut App) -> bool {
...
+        Category::Terminal => {
+            state.terminal_requested = family.clone();
+            state.terminal_effective = effective;
+        }
+        Category::Code => {
+            state.code_requested = family.clone();
+            state.code_effective = effective;
+        }
+    }
+    reinstall_theme(cx);
```
wingleeio, line 632: [P2] Preserve all four new font settings during Shell saves

The new setters write terminal_font_family/code_font_family and their sizes to SettingsStore, but Shell keeps an older UiSettings snapshot. Shell::schedule_save (shell.rs:3474) refreshes only the existing UI font fields and three independent background/link fields, then settings::replace replaces the entire central settings object with that stale snapshot. The new CodeFontSizeChanged handler updates open editors but does not refresh Shell's settings snapshot either.

Choose a new terminal or code font/size, then resize a panel or perform another action invoking schedule_save: the four persisted values revert to the Shell's old values. TypographyState can keep displaying the chosen values until restart, masking the loss. Sync all four fields from the current store/typography state before replacement, or update only Shell-owned fields. Extend panel_saves_preserve_settings_selected_outside_the_shell to cover both new families and sizes and verify a disk reload.

crates/ui/src/files/preview.rs
```diff
@@ -3179,7 +3179,7 @@ impl FilesSurface {
                     .pr(px(18.0))
                     .when(!word_wrap, |element| element.whitespace_nowrap())
                     .font_family(theme.font_mono.clone())
-                    .text_size(px(11.5))
+                    .text_size(px(theme.code_font_size))
```
wingleeio: [P2] Scale plain file-preview row geometry with the new font size

The text now uses code_font_size (supported up to 32 px), but render_preview_line still sets non-wrapped rows to PREVIEW_LINE_HEIGHT = 20 px, and the list's uniform-height hint remains 20 px. With a larger supported size, consecutive baselines remain 20 px apart even when glyphs/line boxes exceed that height, causing overlap or clipping and making source text unreliable to read.

Derive the preview row/line height from the selected code size and invalidate the list measurements on typography changes. Verify a multi-line non-wrapped preview at the maximum supported size as well as the default. The editable-buffer path already accounts for its font size; this plain-preview path needs equivalent handling.

wingleeio (changes requested):
Reviewed the complete PR at 4afe72351722d39ffbbcbb19dc5688de3fb4bd04, including the follow-up fixture fix and the prospective merge against main 6d39f8c6a8f3c62a332f2abbafe989177da289e9. Requesting changes for the three concrete regressions below.

No evidence of a backdoor, exfiltration, obfuscated payload, malicious dependency change, or intentional security bypass was found. This is not a guarantee of author intent or absence of vulnerabilities. The terminal issue has a security-relevant display/selection integrity consequence; the other findings are functional regressions.

Audit coverage: every diff hunk across all 16 changed files; font-family deserialization, availability and fallback; system-font enumeration and parsing; numeric settings normalization and migration; central settings persistence and Shell saves; picker filtering, indexing and keyboard dispatch; theme installation and cache invalidation; terminal shaping, cursor/selection geometry, mouse mapping, scroll and PTY resize; file-editor/preview, markdown and diff sizing; example initializer callers; CI logs and merge contents. No dependency, lockfile, workflow, build-script, font asset, binary, symlink or permission change is part of the prospective merge diff. Added text was checked for hidden control/format characters, and git diff --check passed.

Security boundaries: families are selected from the existing bundled/device-local catalog, not downloaded or interpreted as executable commands or file paths. System-font enumeration/parsing already existed on main. Settings load clamps finite sizes to 8–32 px and replaces non-finite values with defaults. Picker sizes come from fixed ladders. Terminal rows/columns remain bounded to 1–500/2–500 and resize uses the existing structured RPC, device target and debounce. Terminal keyboard encoding, paste sanitization and session routing are unchanged. The new clamp_font_size helper itself does not normalize NaN, but current production callers use normalized settings or fixed finite ladder values; I did not find an external NaN path and am not reporting that as an exploitable vulnerability.

Validation: inspected the initial revision's CI logs: 997 UI library tests passed, including the added settings/typography/picker tests. Its Linux browser fixture failed with E0061 because typography::init changed from four to eight arguments. The follow-up commit fixes all four stale example callers; that initial compile failure is therefore not an outstanding finding. The replacement CI run is still running at review time. Those earlier passing tests do not establish that the updated head is fully green.

A local HarfBuzz metric probe using the installed DejaVu Sans face reproduced the proportional-font/grid mismatch described below. This is a metric-level reproduction combined with tracing the actual renderer, not an interactive application screenshot. No full application build/test suite or interactive multi-platform visual review was run locally. The settings overwrite and fixed preview-row defects were established by tracing their production code paths; they need regression tests covering those paths.

Additional non-blocking accuracy note: the claim that defaults preserve every previous size is not exact. The shared code default is 12.5 px, while the old diff, plain file preview, and file editor defaults were 12, 11.5, and 13 px respectively. Please document or deliberately reconcile that behavior.

ShayanAbbas1:
Thanks — all three findings are fixed in 94f2cfd3, along with the size-default inaccuracy you flagged as non-blocking. Taking them in your order.

## [P2] Terminal cell positioning

Took your first remedy: constrained terminal choices rather than rewriting glyph placement. Per-cell placement is the more correct answer and would keep proportional families usable, but doing it properly means bold/italic and fallback faces, wide and combining clusters, and the selection geometry — that belongs in its own change, not bolted onto this one.

Families qualify by comparing the real `i`/`m`/`W`/`0` advances from each face's `hmtx` table within a 1% tolerance, not a PANOSE bit or fontdb's `monospaced` flag — the renderer's assumption is about actual advances, so that is what is measured. The check ANDs across every face in a family, since selecting a family also selects its italics and one proportional italic would break the grid just as badly.

Only the terminal picker narrows. Code and diffs keep the full catalog, as they lay text out naturally with no cell grid.

A persisted proportional terminal family now resolves to Geist Mono rather than rendering a broken grid — covered at both `typography::init` and the setter path, since a settings file written by an earlier revision of this branch can name one.

## [P2] Shell saves

Confirmed exactly as described. `schedule_save` refreshed only the two older UI font fields before `settings::replace` published the stale snapshot wholesale. `sync_independent_settings` already existed for this class of bug, and its doc comment says so; the four new fields belong in it. I also folded `ui_font_family`/`ui_font_size` into that same sync — they were refreshed separately two lines below, and that split is what let the new fields get missed in the first place.

`panel_saves_preserve_settings_selected_outside_the_shell` now covers both families and both sizes across repeated saves, and asserts they survive a real `UiSettings::load` round-trip. Verified it fails without the fix.

## [P2] Plain file-preview row geometry

Row height now derives from the preview's own text size, with `PREVIEW_LINE_HEIGHT` as both the exact default result and the floor. The uniform-height hint and the painted row read one helper, so they cannot disagree. Typography-generation changes call `remeasure()` — deliberately not `reset_with_uniform_height`, which nulls `logical_scroll_top` and would jump an open file to line 1 on every size change.

Tested at the default and the 32px maximum. This is a unit test on the height derivation, not a full render test: standing up a preview row needs a `Window` plus a loaded document and highlight cache that the files-surface harness does not provide. Stating that plainly rather than implying render coverage.

## Size defaults — your non-blocking note

You were right and this was worse than an inaccuracy in the description; the code was wrong. One shared 12.5 default had replaced four real defaults (markdown 12.5, diff 12.0, editor 13.0, plain preview 11.5), so a fresh install rendered three of the four surfaces at sizes the app had never used.

Rather than document the drift, each surface now keeps its historical baseline and scales proportionally from the shared setting, so the default is pixel-identical to `main` and the surfaces keep their relative proportions at every other size. Regression tests assert the exact historical value per surface at the default; they fail on the previous revision.

The trade-off is that the chosen number is exact for code blocks and proportional elsewhere — 14px renders diffs at 13.4. Preserving default fidelity was the explicit requirement, so I took that over the literal reading. Happy to invert it if you would rather the number be exact everywhere.

## On `clamp_font_size` and NaN

Left as-is, matching your assessment. The scaled per-surface sizes feed from already-clamped, already-normalized settings values, so no NaN path was introduced. Worth hardening on its own terms, but not as part of this change.

## Verification

`cargo check --workspace --all-targets` clean — `--all-targets` specifically, since the fixture break you saw in CI slipped through because neither `-p zeron-ui` nor `--workspace --bins` compiles examples. 977 lib tests pass.

Separately, I verified against installed fonts on my macOS machine that the fixed-width filter keeps Nerd Font families (MesloLG* variants, including the `Propo` ones, whose ASCII stays monospace even though their icon glyphs are not) and Dank Mono, while excluding proportional families — 28 of 390 installed families qualify. The committed tests parse the bundled Geist/Geist Mono bytes instead, so nothing depends on device-installed fonts in CI.

Not re-verified interactively at every size on every platform; the terminal change is the one I would most want a second pair of eyes on.

crates/ui/src/changes.rs
```diff
@@ -289,8 +307,8 @@ impl DiffHorizontalGeometry {
         let font_id = window.text_system().resolve_font(&mono);
         let column_width = window
             .text_system()
-            .ch_advance(font_id, px(DIFF_TEXT_SIZE))
-            .unwrap_or(px(DIFF_TEXT_SIZE * 0.6))
+            .ch_advance(font_id, px(diff_text_size(theme)))
```
wingleeio: [P2] Measure shaped diff widths before allowing proportional code fonts

FontKind::Code still accepts the entire font catalog, including bundled Geist, but DiffHorizontalGeometry::resolve computes max_text_width as visual-column count × ch_advance (the advance of 0). That is only a reliable extent for fixed-width text. code_text_viewport subsequently fixes the content to that estimate and calls overflow_hidden, so a wider shaped line is permanently clipped; horizontal scrolling cannot reveal its suffix.

Reproduced with this repository's bundled Geist at the default 12px diff size: 0 advances 7.956px, so a 100-column line gets a 795.6px text extent. HarfBuzz shapes 100 W characters to 1122.12px. Even the additional CODE_PADDING_RIGHT=24px leaves approximately 302.52px clipped (the left padding is consumed before the text). The same estimate feeds unified and split views. This can conceal changed code during review; it is a display-integrity defect, not evidence of code execution or malicious intent.

Compute/cache the maximum actual shaped line width for the selected font and size, including tabs and fallback glyphs, and invalidate it when typography changes. Alternatively constrain the shared code/diff catalog until proportional layout is supported. Add a renderer regression showing that a long wide-glyph line and its suffix remain reachable at maximum horizontal scroll in both modes.

wingleeio (changes requested):
Re-audited 94f2cfd3fe62120eac9c98236aee4f77cb9e473d, including the follow-up review fixes and integration with current main e74f275e69b8a274d76d7a4db8086f3916845160. Requesting changes for one newly identified diff-display integrity defect below.

The previously reported issues are addressed in the updated source: terminal family selection and persisted requests are checked against a fixed-width catalog; Shell synchronizes all new settings before publishing its snapshot, with disk-reload coverage; plain preview row sizing and list remeasurement now track typography. Historical per-surface default sizes are restored for fresh default settings. No evidence of malicious code, dependency/asset payloads, credential harvesting, exfiltration or deliberate authorization bypass was found. This is not proof of contributor intent or a guarantee of zero vulnerabilities.

I resolved the Appearance-settings merge conflict locally while retaining main's page scrollbar and adding independent scroll states to the three searchable family pickers. No resolution was pushed: maintainer authorization made that conditional on the remaining code being good, and the new blocker still needs fixing. The prospective diff against main remains confined to the expected 16 UI/example files, with no dependency, lockfile, workflow, binary or executable-mode changes.

Validation: 1026 desktop library tests passed locally under the repository lockfile on the integration checkout. Existing CI on the contributor head is green. The additional independent HarfBuzz probe used the repository's own bundled Geist.ttf at the default 12px diff text size: the 0 advance is 7.956px; 100 W characters shape to 1122.12px, while the production column-count estimate is 795.6px. That gap exceeds the fixed 24px shaping padding by about 302.52px. This is an actual-font metric reproduction combined with tracing the production clipping/scroll-extent code, not an interactive screenshot or a new automated renderer test. The existing tests therefore do not cover this defect.

No manual native-platform interaction matrix or arbitrary installed-font fuzzing was performed. The PR description also still describes proportional terminal fonts as supported; it should be updated to match the fixed-width restriction already implemented.

ShayanAbbas1:
Fixed in e519cdce, with main merged in c2fe9b71. PR description updated for the fixed-width terminal restriction and the per-surface default sizes.

## [P2] Diff shaped widths

Took the first remedy — measuring — rather than constraining the code catalog, since your wording framed constraining as a stopgap "until proportional layout is supported."

`max_text_width` is now the max of two things: the old `visual_columns × ch_advance` estimate, kept only as a **floor**, and the real `window.text_system().shape_line(...).width()` of each row's exact string. The value can therefore only grow relative to today, and a fallback face's true advances are measured rather than assumed. Tabs are probed rather than assumed to fit: rows paint raw `\t` and let the shaper decide, so a single `shaped_width("\t")` probe per measurement decides whether the column estimate still bounds a tab, and tab-bearing lines take the shaping path when it does not.

One correction to the framing, since it affects how far the fix reaches: **this was not confined to proportional fonts.** `columns × ch_advance` under-measures any line whose glyphs resolve through a fallback face, so CJK or emoji in a diff could already clip on `main` under Geist Mono. Making the code family configurable exposed the defect rather than introduced it. The fix covers both.

Cached per file against `(typography::generation, resolved size)` — generation covers family and size changes through the typography setters, the size bits additionally cover a theme-driven change that leaves the generation alone. Only files with a rendered row measure at all, so collapsed and off-screen files in a virtualized list cost nothing.

Performance tradeoff, stated plainly: every line of a measured file is shaped, with no sampling and no widest-by-column heuristic, because shaped width is not monotone in column count and any cheap filter risks under-measuring — which is the bug. A fixed-width family skips shaping for ASCII lines, where `columns × advance` is exact rather than approximate. Worst case is an O(lines) shaping pass on first paint of a large diff in a proportional family. That ceiling and its upgrade path (background executor) are named in a comment rather than pre-built.

### What the test covers, and what it does not

`wide_glyph_lines_stay_reachable_at_max_horizontal_scroll` is a `#[gpui::test]` in a real window with the bundled Geist registered, so it shapes through real HarfBuzz and needs no device fonts. It asserts the fixture is genuinely proportional (`shaped > 100 × ch_advance`, which is also the regression proof — the old formula produced exactly `100 × ch_advance`), then asserts the content box clears `padding_left + shaped` in **both** unified and split, plus that the cache returns a stable value and re-measures on size change.

**It does not instantiate the `Changes` view or drive a scroll handle to its maximum offset.** It tests the geometry the renderer consumes, against the same padding constants `code_text_viewport` applies. Calling that out explicitly rather than letting it read as rendered-pixel coverage.

## Merge resolution

Resolved independently and pushed, matching the shape you described: main's page scrollbar and scroll-host treatment are intact, and each of the three family pickers owns its own `PageScroll` reached through the same match accessor the branch already uses for `Popup`/`FocusHandle`. Worth confirming your instinct there — a single shared state would have meant the second picker opened at the first one's offset with shared rail fade timers, and neither the compiler nor the tests would have caught it.

Main's `on_font_list_hover` was hardcoded to the single list, so it is now a kind-capturing listener on the host.

The three size dropdowns intentionally get no scroll state: `render_size_picker` has no `max_h` or `overflow_y_scroll`, and with 7 and 10 rungs the menus size to content and never scroll, so a rail would have nothing to track. Adding one would mean inventing a max height the design does not ask for. Say the word if you would rather they match the family pickers structurally.

## Verification

`cargo check --workspace --all-targets` clean, 997 lib tests pass post-merge, no conflict markers.

Unchanged from last time in what I have not done: no interactive multi-platform pass, and the terminal and diff renderer changes remain the two I would most want a second pair of eyes on.

crates/ui/src/changes.rs
```diff
@@ -5623,6 +5723,74 @@ rename to new_name.rs
...
+                let column_width = window
+                    .text_system()
+                    .ch_advance(window.text_system().resolve_font(&mono), size)
+                    .unwrap()
+                    .as_f32();
+
+                // Guards the fixture itself: if Geist had not registered, the
+                // fallback could be fixed-width and the regression unprovable.
+                assert!(
+                    shaped > 100.0 * column_width,
```
wingleeio: [P2] Exercise a real font backend in the clipping regression

This test does not load or measure Geist. `#[gpui::test]` constructs TestPlatform with NoopTextSystem; that implementation ignores add_fonts and the requested font descriptor and gives every BMP character a 0.6-em advance. The fixture guard still passes: summing 100 simulated 7.2 px advances yields 720.000671 px, while `100.0 * column_width` is 720.0 px. That floating-point difference is mistaken for proof of a proportional font. This leaves real shaping/fallback behavior, including syntax-run boundaries on macOS, untested. Please use a native text backend (or a deterministic injected backend that genuinely models these different advances, plus native integration coverage), require a material proportional-width difference, and check the actually highlighted rendered extent rather than only the unstyled helper.

wingleeio (changes requested):
Re-reviewed c2fe9b717133d4ac2c77e5f1ba945119adee7417, including the shaped-width fix and merge-conflict resolution, and integrated locally with current main e4702fa68cacdc378c31a1b61bb9ff79403d66bc. No evidence of malicious additions, credential access, exfiltration, execution hooks, dependency/workflow changes, or authorization weakening. Earlier terminal grid, persistence, baseline sizing, and preview geometry fixes remain intact. The merge resolution correctly preserves separate scroll state for all three family menus.

Validation: 1,027 desktop library tests passed; all substantive PR CI checks are successful. Raw diff/mode and added-line hidden-control checks are clean. Also built and ran a temporary native Linux GPUI fixture with bundled Geist and actual TypeScript syntax spans. The plain/proportional extent fix works for the tested Linux cases.

Holding approval because the new regression is not actually a native-font test (inline finding), and there remains a platform-specific measurement mismatch that it cannot rule out. shaped_width measures one all-black TextRun, while painting uses syntax-colored runs. In the pinned GPUI revision aa009411, WindowTextSystem::layout_line splits FontRuns at decoration changes, and gpui_macos/src/text_system.rs:532 onward deliberately alternates adjacent runs between font_size and next_up(font_size) to break cross-run ligatures. The cached width also has no highlight-state key, although highlighting arrives asynchronously.

Concrete diagnostic: TypeScript source consisting of `WWW(WWW);` repeated 200 times, using bundled Geist at 12 px. Actual syntax runs are `WWW`, `(`, `WWW`, `);`. Independent HarfBuzz measurement gives 15468.36 px for the whole line versus 15540.00 px when shaped at those boundaries: a 71.64 px difference, greater than the 24 px slack. The actual Linux renderer preserves shaping across these same-font spans and passes. I cannot execute CoreText here, so this is a source-supported macOS concern, not a claimed native macOS reproduction.

Please validate this with real native macOS text shaping and make the extent bound the same runs that are painted (including any change when highlights become ready), or use another demonstrably conservative bound. Add a regression that checks the highlighted suffix is reachable in unified and split views, with real fonts. This closes the remaining gap in the previous clipping finding; a green simulated-metrics test does not establish it.

wingleeio (approved):
Security re-review and maintainer fix complete for 88ba5fc780b759e09e500ce90ed6d206a0260e1a.

Fixed the remaining diff-width concern: measurement now uses exactly the syntax TextRuns that painting uses, including native shaping at color boundaries. The per-file cache retains the highlight snapshot and invalidates on highlight replacement, typography, size, family, and theme/style changes. Measurement and rendering share the same highlight snapshot. Removed the sampled-monospace shortcut, which could bypass actual shaping.

Replaced the simulated-font regression with the real platform text backend and a material proportional-width assertion. Coverage includes plain-to-highlighted transitions, replacement highlights, wide proportional text, repeated TypeScript syntax boundaries, tabs/Unicode, minimum/default/maximum sizes, cached measurements, and suffix reachability in unified and split views. Added this regression to the existing macOS CI job. Also corrected the stale terminal-family documentation.

Validation: 1,027 local desktop library tests passed. All CI jobs passed on this exact commit (run 35007958509), including the native macOS diff font geometry test, Linux desktop/native font tests, session-sync regressions, iOS, Linux browser, macOS browser/preview, and frame recovery. Integrated with current main e4702fa6; diff checks clean.

Reviewed the full feature and the maintainer changes. No evidence of malicious code, credential access, exfiltration, execution hooks, or authorization weakening. No outstanding merge-blocking findings. Earlier terminal-cell, settings-persistence, default-size, preview-geometry, and merge-conflict findings are addressed. Approval is scoped to this reviewed commit; it is not a guarantee that the entire application has no vulnerabilities.

PR #376, opened 2026-09-15
fix(opencode): harden 1.18 session create + speak the 2.x wire

rabden:
## Summary
- opencode 2.x moved its entire HTTP surface under `/api/*`, changed the SSE event vocabulary, and answers **405** to `POST /session` — every opencode run through zeron fails at session create against a 2.x binary. This PR keeps 1.18 support and adds 2.x (verified end-to-end against 2.0.3).
- Commit 1 (standalone fix): `POST /session` could 500 once per directory on a shared db whose schema a newer 2.x install rewrote (lazy `Project.migrateProjectId`); the failing request itself commits the project row, so one immediate retry heals it.

## Root cause (the 405)
- 2.x serves its web UI as a catch-all, so `/global/health` still answers 200 there and the old harness "connected" fine — then died on the first real call.
- Everything the turn engine needs changed shape: session create takes `location.directory` in the body, model+variant are set on the session (`POST /api/session/{id}/model`) instead of riding the prompt, prompts are `{text, files}`, abort is `/interrupt`, and the bus moved to `/api/event` (serving nothing without an `Accept: text/event-stream` header).

## Changes
- Boot-time generation detection via the health endpoints — the `version` field is the only unambiguous discriminator (1.x also serves `/api/health`, but without one; both observed live on 1.18.31 + 2.0.3).
- A per-server `Protocol` branches every wire call: session create/resume, prompt/command posting, model catalog (`GET /api/model`, a flat enabled list with `{id, settings}` variants, folded into the provider-catalog shape + a short poll for the models.dev sync), commands, busy-map, abort, and permission replies.
- 2.x SSE frames (`session.execution.*`, `session.step.*`, `session.text.delta`, `session.tool.*`, `session.usage.updated`, `permission.asked`) are rewritten into the 1.x payload shapes at the bus boundary (`normalize_v2_frame`), so feeds, settle, retries, and subagent handling stay wire-agnostic.
- `permission.asked` keeps its name on 2.x and auto-approves on the session-scoped reply route — without it, a tool reading outside the workspace waits forever (caught live).
- Session-create 5xx retry (commit 1), with a pointer to `~/.local/share/opencode/log/opencode.log` where the body's `ref` keys the cause.

## Test plan
- New fake-server tests for both wires (text stream + settle, tool open/resolve chips, execution failure, interrupt) and table tests for the frame normalizer, the `/api/model` fold, and the 2.x prompt body.
- Full workspace suite green (all crates).
- Live, against real binaries with the new `opencode_turn_probe` example: 2.0.3 — discovery (64 models), plain turn, tool turn with out-of-workspace read (permission auto-approved, correct `ToolResult`); 1.18.31 — discovery + tool turn unchanged.


rabden:
@wingleeio . Please merge this without that I'm unable to use opencode which my daily driver

wingleeio (approved):
Thorough security review completed for 4663ead800c099d189e8b0ef8ab9ab56dc2d7648. Reviewed all feature changes, tests, probe, and the maintainer fixes; compared protocol assumptions against the published OpenCode v2.0.3 source. No evidence of malicious code. The original PR required security and correctness fixes, now committed on this branch.

Fixed security findings:
- OpenCode 2.x `reply: always` persists project-wide saved permissions and can release other pending requests. The adapter now approves only once; it never persists a blanket grant. Source: https://github.com/anomalyco/opencode/blob/v2.0.3/packages/core/src/permission.ts
- The inherited global-bus handler answered permissions without checking ownership. Permissions and questions now require an explicit session belonging to this run or a known child; foreign and ownerless requests are ignored.
- `auto_approve: false` now requires an explicit answer through the existing input bridge. Declines or dropped answers reject the request.
- Tool identities are scoped to session/message/call, so repeated provider call ids cannot mix up tool names or chips. Pending entries retire on completion/failure/interruption, with a hard limit that fails the run visibly instead of retaining unbounded state.
- The optional live probe now uses a private temporary workspace rather than a predictable shared /tmp directory.

Fixed protocol/reporting gaps:
- Handle the actual `session.tool.failed` and `session.reasoning.delta` event names from v2.0.3.
- Preserve external interruption as Interrupted instead of reporting success.
- Surface recoverable step failures without poisoning a later successful execution; terminal execution failures still fail the run.
- Correct the fake model catalog's variant shape and verify actual model/prompt POST bodies. Restore the removed v1 pre-busy error/interrupt regression and verify repeated 5xx session creation stops after exactly one retry.

Validation: 222 local harness tests passed, 10 live-provider tests intentionally ignored; the probe example compiles. Added OpenCode unit and HTTP/SSE regression tests to the required CI job. All CI jobs passed for this exact commit (run 35013492495), including the new permission/protocol checks, session sync, desktop, iOS, Linux browser, and macOS checks. Current main advanced to d8ce7b6e during review; inspected those changes, merged locally without conflict, and reran harness library/OpenCode integration tests successfully. Diff/hidden-control checks are clean.

No new dependencies, credential destinations, shell interpolation, or authentication weakening were found. Automatic approval remains an explicit unattended-run capability; this does not claim OpenCode is an OS sandbox. Live model-provider turns were not executed during this review. No remaining merge-blocking findings in the reviewed change; this is not a guarantee of zero vulnerabilities in the entire application.

PR #383, opened 2026-09-15
Show the account's plan usage in the composer footer, with switching

ghostpyy:
What is left of the plan the open chat spends against was only visible in Settings → Accounts, the wrong place to notice you are about to run out mid-session.

A chip in the composer footer shows the brand mark and what is left of the headline window: Claude's 5-hour "Session", or whatever a weekly-only plan reports. Its colour keys off the window shown, so the number and its tint tell one story. Clicking opens a card with every window and its reset, the other logins one click from being live (the same `ActivateAgentAccount` the settings page uses), and a way out to Settings, which keeps ownership of adding and forgetting.

Percentages say what is left, not what is spent, and the meters drain to match; Settings → Accounts moves with them.

The accounts snapshot moves onto `AppState` so the chip and the settings page read one copy behind the engine's 60s usage cache. The chip refreshes once per TTL, only while a chat is on screen in a focused window; a failed probe counts toward the TTL so it cannot retry every frame.

Validation: `cargo test -p zeron-ui --lib -- --test-threads=1` — 1005 passed. No new warnings (134 before and after).

Eren Saglam

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/383"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792096600&installation_model_id=425430&pr_number=383&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F383&signature=5731563d18d26a543b5480aa0ec1256bbec12cf2844c1fc2ca882af1868daead"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


crates/ui/src/pickers.rs
```diff
@@ -2756,6 +2801,382 @@ impl Pickers {
...
+        let Some(engine) = self.state.read(cx).engine().cloned() else {
+            return;
+        };
+        self.switching_account = Some(account.id.clone());
+        // Tolerant param shape, matching settings/accounts.rs.
+        let params = serde_json::json!({
+            "id": account.id,
+            "accountId": account.id,
+            "harness": account.harness,
+        });
```
wingleeio: [P1] Target account switching at the selected chat's device

For a chat hosted on another device, this request still omits `targetDeviceId`, so `EngineRpc` executes the credential swap locally. The display is also sourced locally: `usage_account()` only reads the chat's harness, and `drive_account_usage()` always loads `None`. I reproduced a chat labeled “Remote build host” showing `local-personal@example.test`; selecting `local-work@example.test` sent `ActivateAgentAccount` without a target and changed the mock local login, leaving the remote account untouched. In production this writes the local CLI credential store and can change the identity used by subsequent local sessions, while failing to switch the chat the user is viewing.

Derive one target from the selected chat's `device_id` and use it consistently for the snapshot, activation and post-action refresh; alternatively hide/disable this control for remote chats until supported.

Native UI before and after activation (synthetic account fixture):

Remote chat incorrectly displays the local personal account

Remote chat switch changes the local account to local-work

crates/ui/src/settings/accounts.rs
```diff
@@ -1221,9 +1199,11 @@ impl Render for AccountsPage {
         let theme = Theme::of(cx).clone();
         let now = Utc::now();
         let dialog = self.render_login_dialog(window.viewport_size(), cx);
-        let refreshing = matches!(self.snapshot, Loadable::Loading);
-        let account_count = self
-            .snapshot
+        // Owned for the frame: the rows below need `&mut Context` for their
+        // listeners, which rules out holding a borrow of the shared state.
+        let snapshot = self.state.read(cx).agent_accounts.clone();
```
wingleeio, line 1208: [P2] Match the shared snapshot to the settings device before rendering it

This now renders the global snapshot without checking `agent_accounts_target` against the page's `target_device`. Reproduction: open Accounts, select a remote device, return to a chat, then reopen Accounts. The chat resets the shared snapshot to local accounts, but `Shell` reuses the existing AccountsPage, which retains its remote target and does not load again. The header still says “Remote build host” while the rows, active flags and usage belong to the local laptop. Row actions still use the retained remote target, so the displayed state and the action destination disagree. Selecting the already-selected remote device also returns early without repairing it.

Keep snapshots keyed by device, or reload/validate the page's target when it becomes visible before displaying actionable rows.

Before visiting the chat:

Remote settings correctly shows remote-team

After reopening the same settings page:

Remote device header now accompanies local account rows

crates/ui/src/pickers.rs
```diff
@@ -2756,6 +2801,382 @@ impl Pickers {
...
+        self.switch_account_task = Some(cx.spawn(async move |this, cx| {
+            let result = engine
+                .client()
+                .call(zeron_rpc::methods::ACTIVATE_AGENT_ACCOUNT, params)
+                .await;
+            this.update(cx, |pickers, cx| {
+                pickers.switching_account = None;
+                if result.is_ok() {
```
wingleeio: [P2] Surface activation failures in the account card

The completion handler clears “Switching…” and discards every `Err`. If the credential store is locked, a saved login was removed, or the RPC fails, the control returns to its idle appearance with no explanation and the old account remains active. I injected `RpcError::Failed("AUDIT: credential store is locked")`; after the attempted switch back to the personal account, the card was pixel-for-pixel identical to its pre-click state. The existing Settings account action displays these errors, but this new entry point gives users no way to distinguish failure from an ignored click.

Persist and render the failure (or show a toast), and clear it on the next attempt. Only present successful completion after activation succeeds.

Native card after the injected activation failure; no error is displayed:

Failed account switch silently leaves local-work active

wingleeio (changes requested):
Request changes on `c6e20a7bde5fbdeb72e9f4183d1ef75232db4d26`: three reproduced findings, with five screenshots uploaded as GitHub user attachments in the inline comments.

- **P1:** A remote chat's account switch targets local CLI credentials.
- **P2:** Returning to Accounts can show local accounts under the remote device label.
- **P2:** The new account switcher silently discards activation failures.

Audited all six changed files and traced account activation through device routing and the credential-writing backend. The first finding is a security-sensitive credential-targeting error; the other two concern account-state correctness and failure feedback. This review does not establish an authentication bypass or token exfiltration.

Validation: **1,035 UI tests passed**, and **16 engine account unit tests passed**. Native GPUI/X11 captures use the PR's actual UI with fixture-only accessors and a deterministic in-memory RPC service; all identities are synthetic. RPC assertions confirmed the missing device target. Live provider services and OS credential stores were not exercised. A broader engine integration-test build exceeded temporary-storage quota and was not counted as passing.

crates/ui/src/pickers.rs
```diff
@@ -2756,6 +2801,382 @@ impl Pickers {
...
+                )
+                .on_click(cx.listener(|this, _, window, cx| {
+                    this.toggle(PickerKind::Account, window, cx);
+                }))
+                .child(usage::brand_mark(account.harness, 12.0, theme))
+                .child(
+                    div()
+                        .text_size(crate::typography::ui_rems(11.0))
+                        .text_color(tint)
+                        .child(usage::remaining_label(headline.used_fraction)),
```
wingleeio, line 2880: Design: please keep the percentage text neutral here and in the popover/Settings. The purple and yellow numbers compete with the account identity and provider mark. Reserve restrained warning color for the meter near exhaustion, with readable neutral percentages throughout.

crates/ui/src/pickers.rs
```diff
@@ -2756,6 +2801,382 @@ impl Pickers {
...
+                                    .child(SharedString::from("Usage unavailable")),
+                            )
+                        } else {
+                            el.child(div().flex().flex_col().gap(px(5.0)).children(meters))
+                        }
+                    }),
+            )
+            .when(has_others, |el| {
+                el.child(popover::menu_separator()).child(
+                    div()
```
wingleeio, line 3034: Design: keep a “Switch account” button in this popover, and refine its styling and placement. Use a compact secondary button with a subtle neutral surface or border, consistent padding, and clear hover/focus states below the usage section. Consider revealing the account choices on click so the initial card stays compact, with consistently aligned account rows and a clear active-account indicator. Keep “Manage accounts” quieter. The requested change is to the visual hierarchy and interaction design.

wingleeio (commented):
The design needs refinement before this is ready to merge. Keep a “Switch account” button in the popover and improve its presentation alongside the usage information.

Requested direction:

- Keep percentage text neutral and consistent across the composer chip, popover and Settings. Use restrained warning color on the meter only when remaining usage is low.
- Make “Switch account” a compact secondary button with a subtle neutral surface or border, clear hover/focus states and padding consistent with the rest of the app. Place it beneath the usage section with enough separation to read as an action.
- Consider revealing the account choices when that button is clicked, keeping the initial card compact. Give account rows consistent alignment and a clear active-account indicator.
- Establish a clear hierarchy: current account identity first, session/week usage next, actions last. Right-align percentages and put reset times beneath the meters in quieter text. Keep “Manage accounts” visually secondary to switching.

Current design for reference:

Current percentage colors and account-switching section

ghostpyy:
Thanks — all three findings were real. Pushed 83a3203.

**P1, remote chats targeting local credentials.** The root cause was that nothing owned the question "whose logins does this chat spend": `active_account_for` assumed local, the chip loaded `None`, and the switch sent no `targetDeviceId`. There is now one answer — `AppState::selected_chat_account_target()`: `None` for this device, else the chat's host. `active_account_for` compares the snapshot against it instead of assuming, the chip's refresh loads it, and `ActivateAgentAccount` carries it as `targetDeviceId`, the same passthrough `SwitchRef` already uses. A remote chat now reads and swaps its own device's credentials, or shows nothing. Pinned by `a_remote_chats_accounts_never_resolve_against_this_device`, which also covers a third device's snapshot.

**P2, stale device on the Accounts page.** The page takes the shared snapshot back before rendering a row whenever the chip has retargeted it, so the header, the rows and the row actions can no longer name different machines. That also repairs the reselect-the-same-device case you noted, without touching the early return.

**P2, silent failures.** The activation error is kept and rendered in the card until the next attempt.

**Design.** Percentages are neutral across chip, card and Settings; colour now lives only on the meter and only once a window is genuinely low, so one warm bar means something. Switching moved behind a secondary button below the usage block — the card opens on the account you are spending and reveals the list on click. Resets sit under their own meters in quieter text, and Manage accounts reads below switching.

Validation: `cargo test -p zeron-ui --lib -- --test-threads=1` — 1006 passed. No new warnings (134 before and after). Not exercised: a real second device, and a live provider.

Eren Saglam

PR #389, opened 2026-09-15
Add agent update monitoring and controls on desktop and iOS

gaelcado:
## Summary

Show installed coding-agent CLI updates across supported devices in the desktop Home composer, with a device label on every row and independently targeted controls. Settings and local desktop notifications remain available while in a chat; the Home card itself stays on the new-chat page. iOS manages updates on a selected host from Devices. Updates wait for active runs to finish and continue if the initiating view closes or its relay disconnects.

- Watch each supported online host independently. Keep last-known notices visible but disable remote actions while disconnected; resume controls only after fresh status arrives. Removed or unsupported hosts leave the list.
- Add device-local checks, notify/automatic/off policies, version dismissal, progress, and cancellation before installation. Gate remote controls with `harness-updates-v1` and recover watches after reconnecting.
- Keep execution leases through subprocess cleanup. On Unix, update commands own a process group so timeout or cancellation terminates installer descendants; timed-out commands reap the direct child before releasing the lease.
- Verify Codex standalone downloads and activate releases atomically; show manual guidance for unsupported layouts.
- Resolve Pi's CLI independently of `PI_ACP_EXECUTABLE` and custom ACP adapter paths, while preserving native ACP overrides. Check `@earendil-works/pi-coding-agent` releases and handle Hermes commit updates without inventing a version.
- Preserve upstream's Antigravity sign-in/sign-out behavior after rebasing. Its pinned ACP server remains manually managed.

## Screenshots

Selected local captures from **September 16, 2026**, before the September 18 rebase. These are visual references, not runtime validation of the current head. Desktop progress styling and multi-device labels were added afterward; the iPhone Simulator views use demo fixture data.

**Desktop — expanded update list in light appearance (01:14:58).**

Desktop Home showing available Claude Code and Codex updates in light appearance

**iOS — host update controls and offline device selection (01:46).** The update pane demonstrates an available update, a queued update with cancellation, a recoverable failure, and manual CLI guidance.

iOS agent updates pane showing available, waiting, failed, and manual update states
iOS Devices pane showing offline MacBook Pro and server hosts

<details>
<summary>Desktop dark appearance and compact single-update indicator</summary>

**Expanded update list — dark appearance (01:10:02).**

Dark desktop Home with expanded Claude Code and Codex update list

**Compact state — one available update (01:15:15).**

Compact desktop indicator for one available Claude Code update

</details>

## Review fixes

Addressed the [September 19 review](https://github.com/zeronsh/zeron/pull/389#pullrequestreview-5256916395):

- Acquire execution leases inside the spawned run task. An update waiting or installing on one harness no longer blocks serial queue dispatch for other chats. Waiting runs remain interruptible, and a cancelled wait never launches the CLI. Turn-start listeners fire when execution starts or steering is accepted.
- Treat the host's explicit apply-cancelled reply as benign on iOS while retaining real failures. A demo Cancel during installation leaves the non-cancellable operation running.
- Preserve and display Settings watch errors with a retry control. Keep the aggregate notification timer alive through frequent AppState changes and clear it after firing.
- Prefer the first CLI version token over a later runtime version, wake shutdown after the update leaves the active set, normalize the local Settings target, and remove hardcoded adapter versions that could misrepresent a remote host.

The independent follow-up review found no remaining issues in the four lifecycle fixes:

- Preserve an active turn when an update defers steering, retaining the prompt in the queue.
- Gate model and command discovery through subprocess cleanup, including caller cancellation and title-discovery timeouts.
- Cancel pre-install automatic work when its agent is disabled and recheck enablement before installation.
- Serialize preference snapshots through the temporary-file write and atomic replacement.

Also explicitly release Unix instance locks when their owner drops, even if a subprocess still holds an inherited descriptor. A deterministic duplicate-descriptor regression covers the failure observed in session-sync CI. The shell fallback fixture now includes `/usr/bin`, where macOS provides `env`.

## Validation

Current head: `16754f6d3aa15cb831b8b979ab73cb5656e467c4`. Based on `main` at `657c6842061b39c3eae514bb2148e7c82384f42a`.

- **240 engine unit tests passed** (1 ignored), **183 harness unit tests passed**, and **17 account/title/turn-quiescence integration tests passed** on the current head.
- **23 end-to-end tests passed** (2 opt-in tests ignored) on the current head, including active-turn steering deferral. Tests use a canonical macOS TMPDIR to avoid the existing `/var` versus `/private/var` path assertion mismatch.
- The prior rebased head passed 9 iOS agent-update tests, 126 shell/UI tests, 7 device-routing tests and 1 project-action command test; these were not rerun for this engine/harness follow-up.
- Changed production Rust formatting and `git diff --check` passed. Existing unrelated formatting differences remain outside this change.
- Fresh CI is pending for `16754f6d`; the previous head's session-sync failure was reproduced locally and fixed by the explicit lock release.

Fresh visual/runtime and animation verification of the multi-device UI remains unverified. The screenshots above predate this change.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/389"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792108519&installation_model_id=425430&pr_number=389&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F389&signature=57841876159f3ce650fa7e46589b10d4ef1396f5444097c100e0add96c518b46"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (changes requested):
Reviewed the full diff at 583de5f0 and ran the engine, harness, RPC and proto suites locally on Linux (engine lib 221 passed, harness lib 182 + codex 22, rpc 15, proto 31). CI is green on the head. I also pulled the real `codex-package-x86_64-unknown-linux-musl.tar.gz` from `rust-v0.155.1` and its layout and `codex-package.json` match `validate_codex_package` exactly, so the standalone path should work against the live release.

This is careful work: the lease/marker design, SHA-256 + size verification, nonce-scoped staging, atomic `current` swap, and process-group cleanup all hold up. One engine regression is blocking for me; the rest are smaller.

### Blocking

**Queue flushing for every chat stalls while any one harness update waits or installs.** `SessionsEngine::dispatch` now awaits `registry.execution_lease(harness_id)` (`crates/engine/src/sessions.rs:372`), which spins on `update_generation.changed()` for as long as `update_pending` is set, i.e. from `begin_update` through post-install verification (minutes, up to the 15-minute updater timeout). Every caller of `dispatch` awaits it inline: `DocHost::drain_queue` holds the chat's `drain_lock` across it, and `spawn_queue_flush_watcher` (`doc_host.rs:843-853`) is a single task that awaits `drain_queue` for **all** chats serially on every status change. So:

1. Claude update accepted while chat A is mid-turn on Claude → phase `WaitingForIdle`.
2. Chat B (Claude) has a queued message; the watcher hits B's `drain_queue` → `dispatch` → blocks on the lease.
3. Chat C (Codex) finishes its turn; its held message never releases until the Claude update fully completes, because the watcher is stuck in step 2. Same for `chat_task` on chat B (doc-change handling is behind the blocked drain).

Before this PR `dispatch` only awaited short things (interrupt grace). Suggestion: don't block in `dispatch`. Either acquire the read lease inside the spawned run task before `harness.run(...)` (the writer is already queued so ordering is preserved), or have `dispatch` `try_read`/check `update_pending` and return a "deferred" outcome so `drain_queue` leaves the row queued and re-drains on `update_generation` change. `note_turn_start` also fires before the wait, so a dropped dispatch has already reported a turn start.

### Should fix

- **iOS: Cancel produces a spurious error banner.** `HarnessUpdates.swift:427-441`: `ApplyHarnessUpdate` is awaited for the whole update (20-min deadline). When the user taps Cancel, the engine's `apply()` returns `Err("update cancelled")`, so the still-pending apply call throws and `action()` writes "update cancelled. Reconnect to check the device's status before retrying." into the pane, while the watch correctly returns the row to Available. Demo mode masks this because `DemoHarnessUpdates.apply` returns normally after cancel. Treat a cancel-originated failure of the apply key as benign.
- **Settings → Agents swallows the watch error.** `crates/ui/src/settings/harnesses.rs:325-352`: `updates` is set to `Loadable::Error(...)` but only `Loadable::Ready` is ever rendered (lines 258, 1160), and the real error at :339 is immediately overwritten by "Reconnecting to device…" at :347 before a paint. A remote engine that hard-fails `WatchHarnessUpdates` leaves the page retrying forever with "Check now" greyed and no explanation.
- **Desktop banner debounce can be starved.** `crates/ui/src/shell.rs:2294-2296`: every `on_state_changed` with an unseen Available row re-assigns `harness_update_banner_task`, dropping the previous 1 s timer, and nothing enters `harness_update_seen` until a timer fires. During an active run AppState notifies far more than once a second, so the banner never posts until the app goes quiet. Only spawn when the task is `None` (or clear it after firing).

### Nits

- `extract_version` picks the numerically largest dotted token in the output (`harness_updates.rs:1648-1655`). Any vendor that prints a runtime/Node version alongside its own would be misread as a newer install. Preferring the first token, or the token following the CLI name, would be safer.
- `harnesses.rs:1229,1235` hardcode "Cursor SDK 1.0.28" / "pi-acp bridge 0.0.33" with no shared constant; the next adapter bump silently makes Settings lie.
- `shutdown()` can wait an extra ~4 s in the success path because the last mutate before `cancellations.remove` is the `Updated` publish, and the next one is the 4 s `Updated → Current` timer.
- `open_harness_update_steps` passes the local device id to `set_target_device` (the page's own switcher uses `None` for local), which discards the page's initial `load()` and re-subscribes twice per open. Works, just wasteful.
- Demo only: `DemoHarnessUpdates.cancel()` cancels the `applyRun` task before checking `cancellable`, so a programmatic cancel during `installing` leaves the fixture row spinning forever.

gaelcado:
Fresh visual review for **319a7210** after rebasing onto main.

The desktop update island now has the tighter compact layout, Settings-style material and separators, a 3.5-row viewport with fades and the small scrollbar, clipped rounded corners, and glyph-only activity (no loading bar).

Desktop captures use the production GPUI renderer with synthetic update data, at 960 × 720 points. Mobile captures are from a fresh native build on iPhone 17 Pro / iOS 26.3 Simulator using the offline demo dataset. The iOS app currently forces dark mode. Versions and errors shown are fixtures.

| Compact summary | Active update — glyph only |
| --- | --- |
|  |  |

| Expanded — dark | Expanded — light |
| --- | --- |
|  |  |

| Scrolled — dark | Scrolled — light |
| --- | --- |
|  |  |

| iPhone — device picker | iPhone — agent updates |
| --- | --- |
|  |  |

[Capture provenance and reproducible fixture](https://github.com/gaelcado/comet/blob/3015e6f358e8d449cceb0859c06b6e9a66a3a66c/evidence.md). These stills verify the captured layouts; they do not establish animation smoothness or live update success. Capture-only changes were removed from the contribution branch.

**Validation:** local Rust application/test/example compilation and the SDK contract check pass; the fresh iOS app build passes. CI now passes Windows, desktop UI, macOS frame recovery, Linux browser, networking, session sync, and Cursor compatibility. The remaining red job is [iOS unit tests](https://github.com/zeronsh/zeron/actions/runs/36071281749/job/107872615840): `TranscriptLayoutTests.testInterruptedKeyboardMotionKeepsTranscriptAttached` failed its keyboard reversal assertions. CI is not fully green yet.

PR #407, opened 2026-09-16
Polish command palette and add quick theme action

gaelcado:
Add the shared glass-modal scrim behind Cmd+K to dim the workspace while preserving the palette’s frosted appearance.

Tighten the Cmd+K palette with a smaller shell, denser action rows, no category headings, and one shared keyboard-hint style. Actions and chats are separated by the same hairline helper used in the sidebar. A clearer magnifier and a 16px icon slot align the search header with action rows. Replace fixed viewport gutters with the existing overflow-aware edge fade; first and last results keep an 8px content inset that scrolls away naturally.

Use Solar Linear sun/moon icons for light/dark throughout the command palette and Appearance settings, with the monitor glyph for System mode.

Add a searchable theme action that switches light/dark immediately, persists the preference, and updates its label while the palette stays open. Enter activation is latched until key release, including Linux/X11 repeats whose `is_held` flag is false. Window deactivation clears the latch to recover from releases outside the app. Action rows expose accessible names, allow text growth, and use concise search and selection hints.

Validation: all 5 command-palette unit tests pass; fixture build, rustfmt, and git diff checks pass. Visually verified light/dark spacing, inline theme switching, keyboard navigation to the final result, and a single filtered result in the isolated fixture. Independent static review of the main change and spacing follow-up found no actionable issues.

Review follow-up: added regression coverage for 35 unflagged X11-style repeats, physical release/re-press, and flagged repeats. Final static review found no remaining actionable issues. Live Linux/X11 fixture validation remains to be repeated; local tests ran on macOS.


crates/ui/src/shell/command_palette.rs
```diff
@@ -327,6 +363,10 @@ impl Shell {
                             }
                         }
                         "enter" => {
+                            if event.is_held {
```
wingleeio: [P2] Suppress held Enter on Linux/X11 as well

This guard relies on a repeat flag that the pinned GPUI X11 backend never sets: `crates/gpui_linux/src/linux/x11/client.rs:1121` in zui revision `c2d273dc` constructs every keydown with `is_held: false`. Since the new theme action keeps the palette open, X11 auto-repeat repeatedly activates it. Reproduced at this PR head in the command-palette fixture: open Ctrl+K, search `theme`, and hold Enter for 2 seconds. Polling the isolated `ui-settings.json` showed 35 alternating appearance states, with rapid whole-window light/dark flashing and repeated immediate settings writes. Please fix repeat detection in the pinned backend (or otherwise suppress repeated activation for one physical press) and add a regression covering the supported X11 path; a test that only injects `is_held: true` will miss this.

wingleeio (changes requested):
Reviewed commit `c5101f33fab7505879ac4d07399a293e33a1eeb7`. Requesting changes for the reproducible Linux/X11 held-Enter regression described inline; not approving or merging this revision.

Security audit: no actionable security vulnerability found in the changed code. Inspected all 10 changed files, the existing settings persistence path, modal input handling, embedded SVG contents, and shared rendering helpers. No new dependencies, external asset references, network calls, command execution, or authentication changes.

Validation:
- All 1,044 UI library tests passed locally (0 failures); all 3 command-palette tests also passed in a focused run.
- CI UI job independently passed all 1,044 tests; all 7 executed PR checks passed.
- Changed Rust files pass rustfmt; PR diff passes whitespace checks. Workspace-wide rustfmt reports unrelated pre-existing differences.
- Built and ran the production command-palette fixture at the reviewed commit, with synthetic data and isolated temporary settings, under Linux/X11 (Xvfb). Checked light/dark switching and persistence, filtered/empty results, keyboard reveal of the last result, Escape, settings activation, and Appearance rendering. These captures are Linux fixture evidence, not native macOS/iOS visual validation.
- Holding Enter for 2 seconds produced 35 observed alternating appearance states in the settings file instead of one toggle. The pinned X11 backend always emits `is_held: false`, so the new guard does not suppress repeat events.

Screenshots captured during this review and uploaded as GitHub user attachments:

Dark palette

Dark palette

Light theme after activation; palette remains open

Light theme after activation; palette remains open

Dark filtered palette after held Enter

Dark filtered palette after held Enter

Last result revealed by keyboard

Last result revealed by keyboard

Empty search results

Empty search results

wingleeio (approved):
Re-reviewed and security-audited `d23c8e8e4f06df1d869924f6e9462a01e4c1a84d`. The previously reported X11 held-Enter issue is fixed; no remaining actionable security or correctness findings.

The follow-up is confined to palette key state and the existing window-deactivation callback. The latch survives repaint/theme changes, ignores both flagged and unflagged repeats, resets on release, and recovers when the release happens outside the app. No dependency, network, command execution, authentication, or settings-path changes were introduced. The original review of the remaining PR changes still applies.

Validation on this exact commit:
- `cargo test --locked -p zeron-ui --lib -- --test-threads=1`: **1,046 passed, 0 failed**, including all 5 command-palette tests.
- Built the production `command-palette-fixture` and reran the original reproduction on Linux/X11 under Xvfb with synthetic data and isolated temporary settings.
- Held Enter for **3 seconds**: exactly **one appearance change and one settings-file modification**, with no further changes during repeat (previous revision alternated 35 times in 2 seconds).
- Physical release/re-press switches again. Switching to another window while Enter is down, releasing there, and returning to the palette also permits the next activation.
- Visually inspected the light/dark filtered action, open palette, and keyboard-revealed last result. Changed-file formatting and diff whitespace checks pass; the checkout remains clean.

Fresh GitHub user attachments below are Linux fixture captures from this revision. Cross-platform CI is still running at review submission; local validation above is complete.

Light theme remains stable while Enter is held

Light theme remains stable while Enter is held

Dark theme after release and a fresh press

Dark theme after release and a fresh press

Unfiltered palette after focus-loss recovery

Unfiltered palette after focus-loss recovery

Final result revealed by keyboard

Final result revealed by keyboard

Issue #406, opened 2026-09-16
Sent message appears in transcript but agent never responds — must resend (Devin via ACP)


tony1u:
Sometimes when I type a message and press Enter, the message appears in the transcript as a normal user bubble but the agent never responds — no error, no working indicator, nothing. Resending the same message works.

### Screenshot

Two identical user messages in a row: the first got no response, the second (manual resend) worked. No error shown between them.

<img width="854" height="273" alt="Image" src="https://github.com/user-attachments/assets/cfc67f26-cb1c-405d-af2f-dcfca5b6ac2b" />

### Setup

- Zeron 0.2.67 on macOS
- Agent: **Devin** (connected through Zeron's ACP client). So far I've only noticed this while using Devin as the harness.
- Has happened multiple times in normal conversation — not tied to a specific tool call that I can tell

### Notes

This looks related to #296 (fixed by #298 in v0.2.59), but the symptom differs: in #296 the follow-up prompt surfaced a `harness protocol error: session/prompt: Invalid request`. In my case nothing is surfaced at all — the sent message just sits there. So the prompt seems to be silently dropped or rejected without any error reaching the UI.

Logs live at `~/.zeron/logs/zeron-headed.log` — happy to attach a relevant excerpt or run a debug build if you tell me what timestamps/markers to look for.

tony1u:
the issue is occurred very frequently.

syntony666:
Found a root cause: with an empty composer during a live run, pressing Enter again fires the interrupt (the composer's Stop mode).
A stray double-Enter right after sending can cancel the just-dispatched prompt, so the message vanishes silently. Fixed in #412.

PR #412, opened 2026-09-16
Stop Enter from interrupting a live run

syntony666:
## Summary

Pressing Enter with an empty composer during a live run dispatched an interrupt via the `Stop` arm of `Composer::on_submit`. A habitual double-Enter right after sending could cancel the prompt that was just dispatched — the message vanished silently (one of the triggers behind #406).

Enter now only submits; it never stops a run. Stop remains available on the button, and on Esc when `escape_stops_active_agent` is enabled.

Refs #406.

## Test plan

- `cargo test -p zeron-ui composer::` — 76 passed, including the new `enter_on_empty_composer_during_a_live_run_never_interrupts` regression test (verified to fail without the fix).
- Manual: send a message, press Enter again on the empty composer — the run continues; the Stop button still interrupts.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/412"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792173124&installation_model_id=425430&pr_number=412&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F412&signature=e8c859657306a84fcc06c1814574ab59d2153d7f5334fb56d442f2fa8809124e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (approved):
Reviewed the complete diff and surrounding submit, queue, wizard, Stop-button, and Escape handling. No security or correctness blockers found. The only production behavior change makes empty Enter a no-op during a live run; explicit Stop and the configured Escape path still invoke the interrupt handler directly. The new engine setter is test-only. No dependencies, build scripts, workflows, permissions, or external endpoints changed.

Validation: all 79 composer-related tests passed locally with `cargo test --locked -p zeron-ui --lib composer::`; formatting and diff whitespace checks passed. Independently restored the old interrupt arm and confirmed the new regression fails at the synchronous interrupt-state assertion, then restored the PR code and confirmed it passes.

Approved the held contributor CI runs after inspecting the diff and workflows. Broader CI is still running; merge should wait for its results.

wingleeio (approved):
Re-reviewed the follow-up commit d1f91f3. It only replaces the regression test's timed recording service with a direct channel assertion; production behavior and the test-only setter are unchanged. No security or correctness blockers found.

All 79 composer-related tests pass on this exact head. Independently verified the revised regression fails at the interrupt-state assertion with the old production behavior and passes again after restoring the fix. Formatting and diff checks also pass. The earlier head's complete Linux, macOS, iOS, and Windows workflows passed; fresh CI for this test-only update is running.

PR #421, opened 2026-09-16
Import Claude Code sessions as chats

UMCEKO:
> **Stacked on #418** (the Nix flake). In git this branch sits directly on `nix-flake`; GitHub's base here is `main` only because stacked PRs aren't enabled on this repo and a fork branch can't be used as a base.
>
> So until #418 merges, this diff also shows its 8 files. **The import work is everything under `crates/`** — 11 files. Once #418 lands this gets rebased onto `main` and the diff becomes exactly those 11.

## Summary

Sessions run in the Claude Code CLI could not be opened here. Nothing read the transcripts under the Claude config dir, and there was no way to give a chat an existing session id, so past CLI work stayed outside Zeron. This adds discovery, import, and a picker.

Opening the command palette and choosing **Import Claude Code session** lists every transcript on the device, newest first. Picking one writes it as a chat and opens it; picking the bulk row imports every session not already here.

### How it works

- **Reading the transcript** (`crates/harness/src/claude/transcript.rs`). It sits beside the stream-json driver on purpose: a transcript record carries the same message envelope the live wire does, so tool calls decode through the existing `decode_tool_use` and imported tool chips render exactly like live ones. `discover` reads only each file's head and tail, so listing stays cheap no matter how large the store is.
- **Which records are conversation.** Only what a person actually sent becomes a user turn. `promptSource` separates `typed`/`queued`/`suggestion_accepted`/`sdk` from the CLI's own injections; transcripts written before that field existed fall back to a string body that is not a `<system-reminder>` or an interruption marker. Attachments, mode records, file-history and the rest are skipped, and sidechain records are left to their own transcript.
- **Writing the chat** (`crates/engine/src/claude_import.rs`). Events fold into doc entries the same shape the live path produces: one entry per user turn, one per reply, with reasoning, text and tool parts inline and each call paired to its result by id. Chats are *claimed* rather than created, so an imported chat resolves (or creates) its project exactly as a live run's first command does.
- **The chat id is the Claude session id.** Re-importing is a no-op, and the row carries `harnessSessionId`, so the first message sent to an imported chat **resumes that session in Claude** instead of starting a new one. Title and branch come over too.
- **Dates are the transcript's own**, not the moment of import: `created_at` is its first record, `last_message_at` its last, so the sidebar sorts imported chats by when the work actually happened. The activity write has to come last because `note_message` stamps the preview with the wall clock.
- **RPC**: `ListClaudeSessions` (unary) and `ImportClaudeSessions` (stream), modelled on the existing profile importer, including its unbounded-channel + `spawn_blocking` bridge.

Tool *output* is deliberately not imported. The live claude adapter never populates it either, so imported chats match live ones and docs stay small.

## Test plan

Unit and integration tests are included: 7 for the transcript reader, 5 for the doc fold, 2 end-to-end through an assembled engine, 1 for the picker filter.

- [x] `cargo test -p zeron-harness -p zeron-engine` — no failures from this change.
- [x] End-to-end (`crates/engine/tests/claude_import.rs`): a transcript becomes a chat whose row carries the title, cwd, branch, project and resume id, whose doc holds the right parts with the tool call resolved, and where an injected turn never becomes a message. Re-import skips rather than duplicating.
- [x] `cargo clippy` clean on every file this touches.
- [x] Verified against a real store on this machine (79 sessions, 2.4 GB): listing took ~200 ms, and importing a 2.6 MB transcript took ~170 ms and produced 9 entries with 5 user turns and 194 tool parts, 1 of them correctly left unresolved because that call was still in flight. Injected task notifications were excluded, and the row came out with the right title, branch, project and resume id.

Known gaps, all deliberate for a first cut: `image` content blocks are dropped (they need a materialized path), subagent transcripts are not imported as nested docs, and there is no keybinding — the palette entry and the `ImportClaudeSessions` action are the entry points.

One pre-existing test, `shell_env::unix::tests::falls_back_when_interactive_attempt_hangs`, fails in my nix-shell environment. It fails identically on a clean `origin/main` tree with none of this applied, and passed under a dev shell with a different shell setup, so it is environment-dependent and unrelated.

Opened as a draft for review.


UMCEKO:
Superseded by UMCEKO/zeron#1, which is based on the `nix-flake` branch (PR #418) so the diff shows only the import work. Will reopen against `main` here once #418 lands.

PR #434, opened 2026-09-17
Fix transcript catch-up animations and add conversation width setting

jsgrrchg:
## Problem and behavior

Returning to a thread while an agent is working can deliver accumulated transcript activity in several batches. Previously, those historical batches could restart activity entrance animations and make the viewport chase the growing transcript before reaching the bottom.

This change distinguishes recovered content from live arrivals. Historical activity appears without replaying its entrance sequence, historical text prefixes are seeded as already visible, and genuinely live content retains its animations. Catch-up snaps to the tail when the reader is following it and preserves the reading anchor when they have scrolled away.

The branch also adds an Appearance setting for conversation width, with live reflow during streaming and no change to composer width.

## Implementation

- Carry a compact replay baseline through sync, the engine's transcript watch, and the UI: stable entry/part IDs and UTF-8 text lengths, without duplicating transcript text. Backfill/checkpoint delivery and ordinary live delivery remain distinguishable even when watch updates coalesce or interleave.
- Derive replay provenance from the Loro containers changed by an import, avoiding an extra full-transcript reconstruction for each historical row. Metadata-only imports inspect no transcript parts in the provenance tracker; snapshots publish the presentation baseline alongside their content.
- Preserve tool animation epochs by stable part ID so interleaved history does not suppress or restart existing live animations. Seed only historical markdown prefixes, leaving new suffixes animated.
- Keep the opening historical cutoff per subscription and combine it with later replay baselines. This protects cached text that has not been painted yet, while allowing two viewers opening at different times to retain different live/historical boundaries. Normal live updates do not perform this merge.
- Add a persisted **Conversation width** slider: **560–1200 logical px**, **16 px steps**, **736 px default**. User bubbles follow the conversation width; the composer stays unchanged. Drag updates coalesce once per frame, saves are debounced, and virtual row heights are remeasured while preserving animation and scroll state. Value/range/reset details appear on hover or interaction, without the unwanted selection outline.

## Measured CPU and memory

**Benchmark revision: `75ba2764`, before integrating `upstream/main` into this PR.** References were pinned to `b4dd24d7` (`main`) and `af59afea` (the branch's original common base). The merge integration has not been benchmarked again; these figures must not be read as measurements of the merged PR head.

The latest pass contains **36 measured executions**: three repetitions of four workloads on each of the three versions. Values below are changes in this branch's mean resource use; negative means lower consumption. CPU changes are relative percentages, not percentage points.

| Workload | CPU vs. main | CPU vs. common base | PSS vs. main | PSS vs. common base |
|---|---:|---:|---:|---:|
| Short foreground stream | −6.64% | −1.66% | −1.84 MiB | −0.24 MiB |
| Foreground stream with long history | −6.13% | −0.01% | +2.93 MiB | +2.30 MiB |
| Stream after reopening long history | −8.41% | +0.12% | −3.24 MiB | −0.37 MiB |
| Simulated historical activity catch-up | +1.39% | +2.41% | −17.76 MiB | −18.79 MiB |

The repeatable resource improvement is **about 18–19 MiB lower PSS during simulated activity catch-up**. Streaming and reopening were approximately equivalent to the common base within the observed variation, with no clear CPU penalty in these workloads. Main included seven unrelated commits, so the larger CPU differences against main cannot be attributed entirely to this branch.

Catch-up did **not** show a consistent CPU improvement: the earlier pass measured −1.23% against the common base, whereas this pass measured +2.41%, with overlapping run ranges. These results do not establish a universal resource reduction or a faster streaming rate.

<details>
<summary>Benchmark method and limits</summary>

- Release builds with thin LTO; immutable executable hashes checked for every run. Fresh processes and data directories, four excluded warmups, and three rounds with alternating version order. All versions produced identical normalized transcript content hashes.
- Linux, Ryzen 9 5950X, affinity to CPUs 8–15, dedicated Xvfb display, 1280×800 window, conversation width 736, and Mesa lavapipe with four workers. Rendering was **software-based**, so absolute CPU readings do not extrapolate directly to a GPU-backed desktop. Other applications remained running.
- A local replay CLI supplied the same text, reasoning, code, and simulated tool events, with no model calls or execution of the simulated tools. Short: 52,696 response bytes and 37 activities. Long: a tenfold preload followed by the same measured turn, finishing with 579,676 bytes and 407 activities. Reopening restarts only the UI against the same populated engine/chat before the measured turn.
- The catch-up workload injects an existing activity followed by four batches of 100 historical activities into the UI subscription, 200 ms apart. It measures **UI catch-up, not actual remote synchronization or engine-side replay-baseline merging**. The reopening workload exercises retention of a nonempty initial snapshot, but measures subsequent live streaming, excluding window startup.
- `/proc` sampling every 500 ms. CPU totals include UI + engine, excluding the driver/replay CLI; 100% means one occupied core. Memory is the sum of each process's peak PSS during the measured phase, which need not peak simultaneously.
- Three repetitions characterize these workloads, not all usage patterns. No FPS, time-to-final-paint, animation completion latency, or long-running leak benchmark was measured. Comparing the previous pass with this one does not isolate the cost of the last fix because they ran at different times.

</details>

## Validation

- Added a regression test that failed before the opening-cutoff fix: a later replay lost the cached text cutoff. It now passes with two subscribers, different opening times, live text between them, and multiple subsequent replay batches.
- Added a UI test that delays painting historical text until after another delivery, verifying that history stays visible without fading while new text still animates.
- Focused engine tests cover coalescing, interleaved replay/live updates, reopening, metadata-only imports, and changed-part work scaling. UI tests cover tool entrances, historical prefixes, tail following versus reading anchors, and width changes during streaming.
- After integrating upstream main and preserving both sides of the two settings/test conflicts: `cargo test -p zeron-ui --lib` (**1,068 passed**), `cargo test -p zeron-engine --lib context_usage_tests` (**5 passed**), and `cargo test -p zeron-engine --lib transcript_history::tests` (**3 passed**). `git diff --check` is clean.

## Remaining limitation

A replay that extends a part already marked live can still animate its recovered suffix until a new viewing session. The per-part live protection deliberately preserves ongoing animation epochs; this PR does not fully distinguish later recovered and live ranges within that same part.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/434"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792274901&installation_model_id=425430&pr_number=434&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F434&signature=9cf6db13fcad324d9e5929b128e4d535c76ab8db4349061ceeca056dc001d51f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio (approved):
Reviewed d699f8e96b2d8e02453233d990b76c1ba57ad965. No blocking security or correctness findings in the 13-file diff.

Audited replay provenance across sync/import/publication/RPC/UI, locking and subscription lifecycle, cursor persistence and causal repair, backward-compatible transcript decoding, UTF-8 cutoff handling, and width normalization/persistence/reflow. No new dependencies or build scripts are introduced.

Independent local validation on this exact head, with --locked:
- Document, sync, and engine library suites: 107 + 35 + 194 passed.
- Complete UI library suite, single test thread: 1,068 passed.
- Engine integration suites (session_publication, restart_resume, codex_subagents, local_profiles): 19 passed; 3 existing tests ignored because they require authenticated real services or private incident fixtures.
- git diff --check: clean.

Total: 1,423 local tests passed, zero failures. Also verified the successful GitHub Actions run against this same head, including the available Linux browser, macOS, and iOS checks: https://github.com/zeronsh/zeron/actions/runs/35280491558

The documented presentation limitation remains: a replay extending a part already marked live can animate its recovered suffix until a new viewing session. This is non-blocking for this review. Native platform coverage above comes from CI; local UI coverage uses the automated/headless suite. These checks provide regression confidence, not a guarantee that every runtime scenario is defect-free.

PR #452, opened 2026-09-18
fix: stabilize Antigravity discovery and suppress unsolicited sign-in

laurentcodes:
Antigravity's slow startup could prevent model and slash-command discovery, while expired credentials could open unsolicited Google sign-in tabs during background probes or chat startup. This change gives discovery enough time and suppresses automatic browser launches, aborting promptly when the server requests sign-in and directing chat users to Settings → Agents.

- Allow 90 seconds for Antigravity model and command discovery, with 100-second remote forwarding deadlines. Retain the last live model catalog after a failed refresh.
- Prefetch slash commands per harness and keep discovery running when the popup closes.
- Strip echoed background-task system messages across streamed chunks, preserving filter state across metadata and tool updates.
- Suppress automatic sign-in browsers with a no-op executable on Unix and Zeron's `--noop-browser` mode on Windows. Fail before launching the agent if suppression cannot be configured. Explicit Settings sign-in keeps its browser flow.

Validation:

- Echo-filter unit tests, slow-start command discovery, echoed-wakeup integration coverage, and forwarding-deadline tests passed.
- Windows command construction, the actual no-op app invocation, stale-login chat/model discovery, and explicit Settings sign-in tests passed locally.
- Python's browser launcher successfully invoked the actual no-op executable; Windows path parsing was checked with literal URL arguments.
- Native Windows execution remains unverified from the macOS development environment.
- The earlier full ACP suite had one failure in the existing `devin_discovery_errors_and_timeouts_retry_without_stale_success` timeout test; all other tests passed.


crates/ui/src/composer.rs
```diff
@@ -5393,51 +5401,76 @@ impl Composer {
...
+        if !self.slash_cache.contains_key(&harness) && !self.slash_failed.contains(&harness) {
+            self.load_slash_commands(harness, cx);
+        }
+    }
+
+    /// one `ListCommands` per harness at a time, targeted like file search (the
+    /// chat/space host device owns the agent binary). `false` when no engine
+    /// is connected to ask.
+    fn load_slash_commands(&mut self, harness: HarnessId, cx: &mut Context<Self>) -> bool {
+        if self.slash_loads.contains_key(&harness) {
```
wingleeio: **[P2] Scope command prefetching by device**

The new automatic prefetch can start a request on device A before the user ever opens `/`. If they then switch to a chat or space on device B using the same harness, `slash_loads` treats A's request as B's pending load, and its result is stored in `slash_cache` under the harness alone. B then displays A's commands/skills and does not fetch its own catalog. This also happens when A's prefetch finishes before the switch, because the cached entry survives navigation. Please key pending requests, cached results, and failed-load state by the effective target device plus harness, and ensure a completion only updates the popup for its matching target.

wingleeio (changes requested):
Security audit and code review: no security blockers found in the changed code. One correctness issue in command prefetching should be addressed before merging.

Validation: 65 ACP unit tests, 44 ACP integration tests, and the model-refresh integration test passed (110 total). Unix browser suppression also passed. Native Windows execution and UI behavior were not tested.

crates/ui/src/composer.rs
laurentcodes: fixed in a4723b04. pending requests, cached commands, and failed loads are now keyed by device and harness. switching targets refreshes the popup, and late responses only update it if the target still matches.

added regression coverage for cached results, late responses, and separate RPC requests across devices.

wingleeio:
Landed via #492 (merged as f1e3edc8), which ports this PR's forwarding-deadline and echoed system-message hunks with co-author credit and covers the rest differently: Antigravity discovery gets a 90 s budget, runs and probes never open a browser, and sign-in is now an explicit Settings action with no automatic downloads. Thank you for the groundwork.

PR #475, opened 2026-09-20
Fix leaked temp unpack directories from archive ACP adapters

LoikStyle:
## Problem

Archive-installed ACP adapters leak their PyInstaller unpack directory on every launch.

The Antigravity adapter is a PyInstaller one-file bundle (~410 MB). Each launch unpacks about 1 GB into `%TEMP%\_MEI<random>`. Observed on Windows: 26 leftover directories totalling 20 GB, accumulated over 4.85 hours of ordinary use — roughly 4.1 GB/hour.

## Root cause

A PyInstaller one-file bundle extracts to `_MEI<random>`, re-executes itself as a child that runs the real application, and **the bootloader parent is the process that removes that directory once the child exits**. Kill the parent and the cleanup never runs.

`crates/harness/src/lib.rs:254` reaps children on Windows through `TerminateJobObject` (via `windows_process.rs`), and the job is created with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`. That unconditionally kills the bootloader parent, so its cleanup is skipped and the unpack directory outlives the run.

Reproduced directly: spawn the adapter, let it unpack, kill the parent — the child survives as an orphan still holding the `_MEI` directory.

## Fix

`Archive`-launched adapters now get a private temp root that this process owns. `ScratchDir` creates it, `apply()` points the child's `TEMP`/`TMP`/`TMPDIR` at it, and `Drop` removes the whole tree once the job is gone. Removal retries briefly, because an adapter that was just terminated may still be closing handles.

Only archive launches take a scratch root: those are the managed, downloaded adapters, and the ones whose payload is worth containing. `spawn_agent` returns the guard alongside the child, and `Session` holds it so it drops after the job is terminated.

Two drop-order details matter and are commented in place — the `Session` field is declared *after* `child` (fields drop in declaration order), and `run_session` binds the guard *first* (locals drop in reverse declaration order). Both keep the job termination ahead of the directory removal.

## Verification

```
cargo test -p zeron-harness --lib                                       183 passed
cargo test -p zeron-harness --features native-fixture --test windows_native   16 passed
```

New unit tests cover the guard removing a nested unpack tree, concurrent roots not colliding, and `apply` pointing every temp variable at the scratch root.

End-to-end on Windows against the real adapter, driving `AcpHarness::antigravity().models()`:

```
SCRATCH LIVE : zeron-agy-acp-server-40900-0
   UNPACK INSIDE SCRATCH ["_MEI000009602"] bytes=24127424
scratch after : []      _MEI after : []
```

The unpack directory lands inside the scratch root instead of the shared temp root, and both are gone once the session ends. Confirmed again on a GUI session: a 1192 MB scratch root was removed when the app exited, leaving no `_MEI` or scratch residue behind.

---
<a href="https://app.blacksmith.sh/zeronsh/codesmith/zeron/pr/475?ref=codesmith_pr_footer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792469227&installation_model_id=425430&pr_number=475&ref=codesmith_pr_footer&repository=zeronsh%2Fzeron&return_to=https%3A%2F%2Fgithub.com%2Fzeronsh%2Fzeron%2Fpull%2F475&signature=ebf16a16c6e72c15216963f4c1a82982d8bc3dd7d3477303177a0db3fed1e21c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


wingleeio:
Ported into #492 (merged as f1e3edc8) with co-author credit: every archive-launched adapter now gets a private temp root that is removed once the process is reaped, so the PyInstaller unpack no longer accumulates. Thank you for the diagnosis and the fix.

